PR #20, opened 2020-10-06
Pull historical events data for 2016-2020

mathcolo:
Work in progress, but it works.


friendchris:
Nice work on rebuilding the APIs. My one suggestion: you should make "arrival" and "departure" into lists & global variables, and set arrival to "ARR" and "PRA" and departure to "DEP" and "PRD". PRA and PRD (predicted arrival and departure) are particularly helpful for data around terminals.

mathcolo:
@friendchris I hear you on `PRA` and `PRD`, but what would the global variables be for?

friendchris:
Just that they would be global constants. This is good to go from my perspective.

friendchris (approved):

PR #21, opened 2020-10-07
Update lint script and pipfile stuff

mathcolo:
Title says it all


package.json
```diff
@@ -25,7 +25,9 @@
     "start-python": "cd server && pipenv run chalice local --port=5000",
     "start-react": "react-scripts start",
     "build": "react-scripts build",
-    "lint": "flake8 --ignore=E501,E731 server && npx eslint src",
+    "lint": "npm run lint-js && npm run lint-python",
+    "lint-js": "eslint src",
```
nathan-weinberg, line 29: Does this not need to be `npx eslint src`?

package.json
```diff
@@ -25,7 +25,9 @@
...
+    "lint-python": "cd server && pipenv run flake8 --ignore=E501,E73 --exclude .chalice",
```
nathan-weinberg, line 30: My fault because I sent this to you :)

```suggestion
    "lint-python": "cd server && pipenv run flake8 --ignore=E501,E731 --exclude .chalice",
```

nathan-weinberg (changes requested):
Not sure we can/need to/should do some kind of `.gitignore`-fu for the Pipfiles but other than that and two nits LGTM!

package.json
mathcolo: Interestingly no; npm is smart enough to know this is the `eslint` provided by the eslint package!

PR #22, opened 2020-10-07
Changed CI to mimic manual linting

nathan-weinberg:
CI now uses npm to lint both frontend and backend instead of just the former


mathcolo (approved):
Awww yeah

PR #27, opened 2020-11-19
Add color to station names in chart titles

austinjpaul:
I've had a go at implementing one of the improved designs for the chart titles. I think the splitting it into smaller chunks helps readability, and I'm a fan of the added color for orientation. (My preferred option with a highlighted background is substantially more difficult to do in HTML canvas, since you have to draw the background color rectangles separately...)

I'm in no way a front-end dev, so let me know your thoughts, esp. if I've done something totally funky.  There are some issues with words overlapping on very small windows- resolving this might make things complicated at best, and infeasible at worst.

Anyway, I'm curious to hear others' thoughts on this approach...

Blue Line - 2020-10-07 - TransitMatters Data Dashboard - Mozilla Firefox 2020-11-19 5_56_19 PM

red_demo

overlap_issue

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -177,15 +177,18 @@ class App extends React.Component {
     }
   }
 
-  graphTitle(prefix, showDirection, showTo) {
+  locationDescription(bothStops) {
     const { from, to, line } = this.state.configuration;
+    var result = {};
+
     if (from && to) {
-      const direction = showDirection ? ` ${station_direction(from, to, line)}` : ""
-      const preposition = showTo ? "from" : "at";
-      const suffix = showTo ? `to ${to.stop_name}` : "";
-      return `${prefix} ${preposition} ${from.stop_name}${direction} ${suffix}`;
+      result['bothStops'] = bothStops;
+      result['to'] = to.stop_name;
+      result['from'] = from.stop_name;
+      result['direction'] = station_direction(from, to, line);
+      result['line'] = line;
     }
-    return prefix;
+    return result;
   }
 
   chartTimeframe() {
@@ -233,7 +236,8 @@ class App extends React.Component {
   renderCharts() {
     return <div className='charts main-column'>
       <Line
-        title={this.graphTitle('Travel times', false, true)}
+        title={"Travel times"}
+        location={this.locationDescription(true)}
         tooltipUnit={"travel time"}
         seriesName={'traveltimes'}
         isLoading={this.getIsLoadingDataset('traveltimes')}
@@ -246,7 +250,8 @@ class App extends React.Component {
         legend={true}
       />
       <Line
-        title={this.graphTitle('Time between trains (headways)', true, false)}
+        title={'Time between trains (headways)'}
+        location={this.locationDescription(false)}
         tooltipUnit={"headway"}
         seriesName={'headways'}
         isLoading={this.getIsLoadingDataset('headways')}
@@ -259,7 +264,8 @@ class App extends React.Component {
         legend={true}
       />
       <Line
-        title={this.graphTitle('Time spent at station (dwells)', true, false)}
+        title={'Time spent at station (dwells)'}
+        location={this.locationDescription(false)}
         tooltipUnit={"dwell time"}
         seriesName={'dwells'}
         isLoading={this.getIsLoadingDataset('dwells')}
--- a/src/App.js
+++ b/src/App.js
@@ -179,16 +179,17 @@ class App extends React.Component {
 
   locationDescription(bothStops) {
     const { from, to, line } = this.state.configuration;
-    var result = {};
 
     if (from && to) {
-      result['bothStops'] = bothStops;
-      result['to'] = to.stop_name;
-      result['from'] = from.stop_name;
-      result['direction'] = station_direction(from, to, line);
-      result['line'] = line;
+      return {
+	bothStops: bothStops,
+	to: to.stop_name,
+	from: from.stop_name,
+	direction: station_direction(from, to, line),
+	line: line,
+      };
     }
-    return result;
+    return {};
   }
 
   chartTimeframe() {
--- a/src/App.js
+++ b/src/App.js
@@ -182,11 +182,11 @@ class App extends React.Component {
 
     if (from && to) {
       return {
-	bothStops: bothStops,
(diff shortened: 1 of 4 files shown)


androotubelli:
Austin, this looks good. One thing that you could do to prevent the overlap on smaller screens, if possible, is to use a media query to move the rightmost text below the leftmost. For example:

Time between trains (headways)
Downtown Crossing southbound
Graph

Though you say this is Canvas, which I have less experience with than SVG, so I don't know if it is as easy as I imagine.

austinjpaul:
Thanks @androotubelli!
I'll have a play around with what you suggested. It's easy enough to calculate whether or not there will be overlap and do something different in that scenario.
The hard part I think will be leaving a larger vertical margin above the chart in case we need two lines for the title, since that happens before drawing any text. I'll keep fiddling with chartjs to see if I can get something working that's not too hacky.

androotubelli:
Yeah a space between the title and the chart would not be great. If a media query is possible for moving the text to the second line, I would think creating an additional space would also be possible as a media query, but I can imagine that may be unnecessarily complicated for Canvas. Perhaps the title could be extracted and applied as an HTML/CSS element rather than as a canvas element?

austinjpaul:
According to @mathcolo, there were previous attempts at adding the title in CSS. I like having the title as part of the image itself if possible, but of course there are tradeoffs, and I won't deny that this approach feels a little hacky.

I haven't figured out if the vertical padding above the chart can be modified dynamically in chartjs, and I'm starting to suspect it can't, so I've added a larger vertical margin, which leaves the unused whitespace _above_ the title unless we need to bump into it on small screens: see screenshots. 

Screen Shot 2020-11-30 at 14 02 20
Screen Shot 2020-11-30 at 14 02 32
Screen Shot 2020-11-30 at 14 02 40
Screen Shot 2020-11-30 at 14 02 47
Screen Shot 2020-11-30 at 14 03 11

src/Title.js
```diff
@@ -0,0 +1,87 @@
...
+      line_color = '#159765';
+      break;
+    default:
+      break;
+    }
+
+  return line_color;
+}
+
+const parse_location_description = (location) => {
```
mathcolo, line 24: It's slightly confusing that `locationDescription` is assembling an object (pretty much for the purpose of data transfer?) and then it's being immediately parsed here. Maybe this can be cleared up or simplified somehow?

src/Title.js
```diff
@@ -0,0 +1,87 @@
...
+
+  if ((title_width + location_width + left_buffer) > chart.chart.width) {
+    // small screen: centered title stacks vertically
+    ctx.textAlign = 'center';
+    ctx.fillStyle = 'grey';
+    ctx.fillText(title, chart.chart.width/2, vpos_row1);
+
+    ctx.textAlign = 'left';
+    var position = chart.chart.width/2 - location_width/2;
+    parse_location_description(location).forEach(
```
mathcolo, line 59: Would this read better as a regular for-of loop, w/ de-structuring?

src/Title.js
```diff
@@ -0,0 +1,87 @@
+const get_line_color = (line) => {
+  var line_color = 'grey';
```
mathcolo, line 2: Since we're in an es6-compatible environment here, `let` may be best over `var` use to avoid weird scope things

src/Title.js
```diff
@@ -0,0 +1,87 @@
...
+
+const drawTitle = (title, location, chart) => {
+  var ctx = chart.chart.ctx;
+  ctx.save();
+
+  const left_buffer = 50;
+  const vpos_row1 = 25;
+  const vpos_row2 = 50;
+
+  var title_width = ctx.measureText(title).width
```
mathcolo, line 48: Just a heads up, some missing semicolons here!

src/Title.js
austinjpaul: Hmm... I see your point. I was trying to avoid having code very specific to the title drawing in the main App file though, so transferring all the info was my solution. But I can move the logic to just the one function if you think that would be cleaner?

src/Title.js
austinjpaul: Much more readable indeed!

src/Title.js
austinjpaul: Should I try to use let in most situations? or just here because of the same variable name?

src/App.js
```diff
@@ -177,15 +177,18 @@ class App extends React.Component {
     }
   }
 
-  graphTitle(prefix, showDirection, showTo) {
+  locationDescription(bothStops) {
     const { from, to, line } = this.state.configuration;
+    var result = {};
```
idreyn, line 182: IMO it is more idiomatic to write:

```js
if (from && to) {
    return {
        bothStops: bothStops,
        to: to.stop_name,
        ...
    };
}
return {};
```

src/Title.js
```diff
@@ -0,0 +1,90 @@
+const get_line_color = (line) => {
```
idreyn, line 1: Potentially more idiomatic to say:

```js
const colorsForLine = {
    Red: '#d13434',
    Orange: '#e66f00',
    Blue: '#0e3d8c',
    Green: '#159765',
};

const getLineColor = (lineName) => colorsForLine[lineName] || 'grey';
```

It might even be helpful to move the `colorsForLine` objects to `constants.js` since it's likely we'd want to reuse it.

idreyn (approved):
Functionally and visually this looks great, just have a couple of minor JS suggestions!

src/Title.js
austinjpaul: great suggestion, thanks for the tips!

src/Title.js
austinjpaul: One further thought: I'm treating the result returned by parse_location_description as disposable, because the call to `reverse`, e.g. it will mutate it. There are a plethora of ways around this, but hopefully that makes the current approach make some more sense?

src/Title.js
mathcolo: I think so! The behavior of "var" is really anachronistic enough that I would recommend you always avoid it in favor of the newer keywords. And because the dashboard was scaffolded with create-react-app, there's no risk of the newer language features breaking somebody (it passes everything through babel)

mathcolo (approved):

PR #32, opened 2021-01-18
Aggregation

mathcolo:


server/chalicelib/MbtaPerformanceAPI.py
```diff
@@ -76,8 +88,8 @@ def get_single_api_data(url):
     return data
 
 
-def get_api_data(day, module, params):
-    url_list = get_many_urls(day, module, params)
+def get_api_data(start_day, module, params, end_day=None):
+    url_list = get_many_urls(start_day, module, params, end_day)
```
austinjpaul, line 92: I know I made these changes in a rush, but i want to note here that we should make start-day and end-day params adjacent. There's also the possibility of multithreading here as well, so a further refactor may be in order.

server/chalicelib/data_funcs.py
```diff
@@ -1,17 +1,70 @@
...
+        s3_dates = (start_date, CUTOFF - datetime.timedelta(days=1))
+        api_dates = (CUTOFF, end_date)
+
+    return (s3_dates, api_dates)
+
+
+def headways(sdate, stops, edate=None):
+    if edate is None:
+        if use_S3(sdate):
+            return s3_historical.headways(stops, sdate, sdate + datetime.timedelta(days=1))
```
austinjpaul, line 38: the mbta api will return the range [start_date, end_date] whereas s3_historical right now will return [start_date, end_date). While the latter is pythonic, I think the former is easier to conceptualize here. In any case, we should standardize.

server/chalicelib/data_funcs.py
```diff
@@ -65,9 +146,39 @@ def travel_times(day, params):
     return travel
 
 
-def dwells(day, params):
+def dwells(sdate, stops, edate=None):
+    if edate is None:
+        if use_S3(sdate):
+            return s3_historical.dwells(stops, sdate, sdate + datetime.timedelta(days=1))
```
austinjpaul, line 152: I was originally thinking we don't bother with threads for the single-day case? I might address this when I experiment with different threading options.

server/chalicelib/s3.py
```diff
@@ -1,20 +1,60 @@
+from concurrent.futures import ThreadPoolExecutor, as_completed
 import boto3
+import datetime
 import csv
 import zlib
 
 BUCKET = "tm-mbta-performance"
-s3 = boto3.resource("s3")
+THREAD_COUNT = 3
```
austinjpaul, line 8: I think we can bump this to 5- i hope to do some profiling soon!

server/chalicelib/s3.py
```diff
@@ -1,20 +1,60 @@
...
+    return (l[i:i+n] for i in range(0, len(l), n))
+
+def download_event_range(stop_id, sdate, edate):
+    delta = edate - sdate
+    jobs = []
+    for i in range(0, delta.days):
+        date = sdate + datetime.timedelta(days=i)
+        jobs.append((stop_id, date))
+
+    thread_sets = chunks(jobs, len(jobs) // THREAD_COUNT)
```
austinjpaul, line 21: Is the chunking approach to cut down on thread-startup overhead? My first thought was just throw all the individual jobs in there and let the pool handle it.

austinjpaul (commented):
Leaving these comments here before I forget they exist.
I'm hoping to do some more playing around with back-end stuff this weekend
Awesome work getting this together, i'm super excited for it to go out! (once we address the things from the mtg notes)

server/chalicelib/aggregation.py
```diff
@@ -0,0 +1,121 @@
...
+    all_data = []
+    delta = edate - sdate       # as timedelta
+
+    # get a range of dates
+    for i in range(delta.days + 1):
+        day = sdate + datetime.timedelta(days=i)
+        data = data_funcs.travel_times(day, [from_stop], [to_stop])
+        for data_dict in data:
+            data_dict['service_date'] = day
+        all_data.extend(data)
```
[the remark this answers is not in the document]
austinjpaul: resolved!

src/stations.js
```diff
@@ -5,7 +5,7 @@ const all_lines = () => {
 };
 
 const lookup_station_by_id = (line, id) => {
-  return stations[line].find(x => [...x.stops.northbound, ...x.stops.southbound].includes(id));
+  return stations[line]?.find(x => [...x.stops.northbound, ...x.stops.southbound].includes(id));
```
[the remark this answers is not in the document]
mathcolo: Optional chaining operator... I sadly don't remember exactly what issue I was having such that I needed to prevent an unhandled error here, so maybe we can do away for now!

server/chalicelib/s3.py
mathcolo: That's right... I made the assumption that it would be best to use boto3 in as serial of a methodology as possible to minimize killing and recreating HTTP connections (although I didn't spend time testing that). I don't actually know if boto3 is doing any keepalive anyway. It probably uses urllib though, so perhaps it's automatic? IDK

server/chalicelib/s3_historical.py
```diff
@@ -9,15 +10,43 @@
...
+    a, b = itertools.tee(iterable)
+    next(b, None)
+    return zip(a, b)
+
+
+def unique_everseen(iterable, key=None):
+    # an itertools recipe from the docs
+    # "List unique elements, preserving order. Remember all elements ever seen."
+    # unique_everseen('AAAABBBCCDAABBB') --> A B C D
+    # unique_everseen('ABBCcAD', str.lower) --> A B C D
```
austinjpaul, line 25: we could import these from `more-itertools` if we choose to install it, but pipenv gave me version conflicts when i tried to add it.

arosgood (approved):

nathan-weinberg (approved):

src/stations.js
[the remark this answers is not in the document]
austinjpaul: nope, let's leave it to be safe then. I just hadn't seen the syntax before.

server/chalicelib/MbtaPerformanceAPI.py
austinjpaul: may punt on this to a future refactor

Issue #29, opened 2020-11-30
Tooltip points to unclear location


austinjpaul:
the hover-over black info box points to the midpoint between a data point and its corresponding benchmark value, and the benchmark is highlighted with a grey dot while the data point is not highlighted at all.

We should consider adjusting how this is displayed: perhaps pointing straight at the data point instead of the midpoint?

PR #36, opened 2021-01-20
Add custom tooltip positioner to center on data value

austinjpaul:
Addresses #29 
I think the new behavior is more intuitive, but someone should verify. 

The new nit to pick is that there's a bias towards points with smaller x values, but I don't think fixing that is as high a priority.


mathcolo (approved):
I checked out the branch and gave it a shot. I like it! Curious what @idreyn thinks, but I'm for getting this in even if not 100% in all cases.

PR #38, opened 2021-01-24
Show beta tag only on dashboard-beta or localhost

mathcolo:
image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -30,6 +30,12 @@ const documentTitle = (config) => {
   return `${config.line} Line - ${config.date} - TransitMatters Data Dashboard`;
 };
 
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -30,6 +30,12 @@ const documentTitle = (config) => {
   return `${config.line} Line - ${config.date} - TransitMatters Data Dashboard`;
 };
 
+const showBetaTag = () => {
+  const beta_tag = document.getElementById("beta-tag");
```
idreyn, line 34: Or `document.querySelector(".beta-tag")`, which would let us continue to refer to it by a classname.

idreyn (approved):
LGTM!

Issue #28, opened 2020-11-30
Adjust data point colors


austinjpaul:
Currently the green and yellow points are not sufficiently distinguishable for colorblind users.

androotubelli:
![colorblindtrains](https://user-images.githubusercontent.com/45049337/101563830-006eb680-3998-11eb-9936-55c82011fb99.png)

This is a potential fix to the data colors. It is a conservative change keeping the stoplight colors but adjusting the hue/saturation. In general, stoplight colors are a terrible color scheme, though it is understandable why it is used rather than choosing any 3 arbitrary colors. I did my best to simulate the change in two types of colorblindness, but I am not colorblind, so only someone colorblind can say for sure if this works better.

austinjpaul:
Here's how the app would show it with the new colors:
![new_colors](https://user-images.githubusercontent.com/8498946/101570121-c4415300-39a3-11eb-8d61-13eb887c1203.PNG)

I think these are great, and much improved! I was a bit worried about the green-on-gray, but the point outlines help with that.

I also had a look at the combination (new green, new yellow, old red) since the old red looks more Red to my protanopia, whereas the new red just looks dark (brownish). The difference to me is quite subtle in the whole plot, so I'd defer to someone with full color vision to decide which set looks better as a whole.
![old_red](https://user-images.githubusercontent.com/8498946/101570192-e33fe500-39a3-11eb-854c-517fc2b06335.PNG)

Warning to whoever ends up implementing: the point colors in the plot and the point colors in the legend are defined in 2 separate places.

austinjpaul:
As discussed in 2021-02-09 meeting, we also would like to _test out_ adding a 4th color (purple?) to highlight really obscenely bad delays (>100% of benchmark)

PR #40, opened 2021-02-16
Make points colorblind-friendly and add purple for >100%

austinjpaul:
Addresses #28 

image

I'm opening this against main for public release, I think. lmk if I should be trialing this in the beta server first. (also, which branch is the beta version using?)


mathcolo (approved):
Nice! Let's try it

PR #41, opened 2021-02-19
Download and store alerts every morning for all lines

mathcolo:
- Download MBTA alerts across all routes every morning
- Use our storage instead of MBTA-performance for quick retrieval

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1,3 +1,4 @@
+import json
 import os
 from chalice import Chalice, Cron, CORSConfig
 from datetime import date, timedelta
@@ -85,7 +86,4 @@ def traveltime_route(user_date):
 @app.route("/alerts/{user_date}", cors=cors_config)
 def alerts_route(user_date):
     date = parse_user_date(user_date)
-    if use_S3(date):
-        return []
-    else:
-        return data_funcs.alerts(date, mutlidict_to_dict(app.current_request.query_params))
+    return json.dumps(data_funcs.alerts(date, mutlidict_to_dict(app.current_request.query_params)))
--- a/server/app.py
+++ b/server/app.py
@@ -1,7 +1,7 @@
 import os
-from chalice import Chalice, CORSConfig
-from datetime import date
-from chalicelib import data_funcs, s3_historical
+from chalice import Chalice, Cron, CORSConfig
+from datetime import date, timedelta
+from chalicelib import data_funcs, s3_historical, s3_alerts
 
 app = Chalice(app_name="data-dashboard")
 
@@ -10,6 +10,15 @@ TM_FRONTEND_HOST = os.environ["TM_FRONTEND_HOST"]
 cors_config = CORSConfig(
     allow_origin=f"https://{TM_FRONTEND_HOST}", max_age=3600
 )
+
+
+# Every day at 10am UTC: store alerts from the previous day
+@app.schedule(Cron(0, 10, '*', '*', '?', '*'))
+def store_yesterday_alerts():
+    # Only do this on the main site
+    if TM_FRONTEND_HOST == "dashboard.transitmatters.org":
+        yesterday = date.today() - timedelta(days=1)
+        s3_alerts.store_alerts(yesterday)
 
 
 def parse_user_date(user_date):
--- a/server/chalicelib/data_funcs.py
+++ b/server/chalicelib/data_funcs.py
@@ -1,8 +1,9 @@
 import datetime
 import pytz
-from chalicelib import MbtaPerformanceAPI
+from chalicelib import MbtaPerformanceAPI, s3_alerts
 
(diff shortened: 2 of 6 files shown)


server/chalicelib/s3_alerts.py
```diff
@@ -0,0 +1,28 @@
...
+def get_alerts(day, route):
+    alerts_str = s3.download(key(day), "utf8")
+    alerts = json.loads(alerts_str)[0]["past_alerts"]
+    return list(filter(lambda alert: route[0] in routes_for_alert(alert), alerts))
+
+
+def store_alerts(date):
+    api_data = MbtaPerformanceAPI.get_api_data(date, "pastalerts", {})
+    alerts = json.dumps(api_data).encode("utf8")
+    s3.upload(key(date), alerts, True)
```
austinjpaul, line 28: This will create an empty-ish json file for days that don't have any alerts (one can dream...), right? I like this, since it provides a nice sanity check for data completeness.

server/app.py
```diff
@@ -12,6 +12,15 @@
...
 
+# Every day at 10am UTC: store alerts from the previous day
+@app.schedule(Cron(0, 10, '*', '*', '?', '*'))
+def store_yesterday_alerts():
+    # Only do this on the main site
+    if TM_FRONTEND_HOST == "dashboard.transitmatters.org":
+        yesterday = date.today() - timedelta(days=1)
+        s3_alerts.store_alerts(yesterday)
+
+
```
austinjpaul, line 23: I'm a bit worried if this cron job fails once (site is offline or mbta api request times out for taking too long), that our data will just have a hole in it. Not sure the best way to address-- notify someone to re-load that day manually?
(speaking of, should we have a command to upload a given day manually?)

server/chalicelib/s3_alerts.py
```diff
@@ -0,0 +1,28 @@
...
+    return f"Alerts/{str(day)}.json.gz"
+
+
+def get_alerts(day, route):
+    alerts_str = s3.download(key(day), "utf8")
+    alerts = json.loads(alerts_str)[0]["past_alerts"]
+    return list(filter(lambda alert: route[0] in routes_for_alert(alert), alerts))
+
+
+def store_alerts(date):
```
austinjpaul, line 25: date vs day- we should probably be consistent at least within a file

server/chalicelib/data_funcs.py
```diff
@@ -91,20 +92,28 @@ def dwells(day, params):
...
-                "text": alert_version["header_text"]
-            })
-    return flat_alerts
+    try:
+        yesterday = datetime.date.today() - datetime.timedelta(days=1)
+        if day <= yesterday and day >= WE_STARTED_COLLECTING_ALERTS:
+            # This is stupid because we're emulating MBTA-performance ick
+            api_data = [{"past_alerts": s3_alerts.get_alerts(day, params["route"])}]
+        else:
+            api_data = MbtaPerformanceAPI.get_api_data(day, "pastalerts", params)
```
austinjpaul, line 101: We should only make this api request on "today". I would also suggest, for the case of day < WE_STARTED_COLLECTING_ALERTS, we say "we don't have data" or something distinct from "no alerts on this day" or whatever the default message is now.

austinjpaul: Maybe return None instead of [] in that scenario?

server/chalicelib/s3_alerts.py
```diff
@@ -0,0 +1,28 @@
+import json
+from chalicelib import MbtaPerformanceAPI, s3
+
+BUCKET = "tm-mbta-performance"
```
austinjpaul, line 4: I don't think we need this here?

austinjpaul (approved):
I love it! Nice and simple approach.

server/chalicelib/s3_alerts.py
mathcolo: Ooh good catch, thanks!

server/app.py
mathcolo: I agree we should notify! I might do that in a new patch

server/chalicelib/data_funcs.py
mathcolo: `json.dumps(None)` turns into string "null", so that sounds good! I've adjusted this so that dates < 2016-01-15 (which is the current lower bound of the frontend date picker) will do that. Given the date picker is like that, it shouldn't be possible to hit, but good to make sure it functions right anyway

server/app.py
nathan-weinberg: Could we spin off a process that attempts some number of retries if we can't reach the MBTA API?

nathan-weinberg (approved):

server/app.py
mathcolo: Absolutely yes! Let's see if it's an issue first?

PR #42, opened 2021-02-20
[Proposal] Support asdf?

mathcolo:
I have 75 versions of python on my laptop, and it's making me a little crazy. [asdf](https://asdf-vm.com/#/) is like pyenv or nvm, but for everything! Adding a `.tool-versions` to our project offers an opportunity to codify the versions we expect, and lets us get going with asdf easily.

For the moment, it works in conjunction with `pipenv`. asdf installs python and node; pipenv installs packages. Perhaps if we went with this option, we could rip pipenv out.

Thoughts?


nathan-weinberg (approved):
Open to this for sure, especially if the team generally wants it. Maybe require some minor changes to CI and npm linting scripts - currently these are using pipenv.

arosgood (approved):

mathcolo:
nah

PR #44, opened 2021-02-22
To add download csv

arosgood:
Just adding something to give people the option of downloading their data to CSV. Let me know what you guys think!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,6 +9,7 @@ import AlertBar from './AlertBar';
 import './App.css';
 import Select from './Select';
 import { configPresets } from './constants';
+import { CSVLink } from "react-csv";
 
 const FRONTEND_TO_BACKEND_MAP = new Map([
   ["localhost", ""], // this becomes a relative path that is proxied through CRA:3000 to python on :5000
@@ -286,11 +287,15 @@ class App extends React.Component {
         isLoading={this.getIsLoadingDataset('dwells')}
         data={this.state.dwells}
         xField={'arr_dt'}
-        xFieldLabel={'Time of day'}
+        xFieldLabel={'Time of day'} 
         yField={'dwell_time_sec'}
         yFieldLabel={'Minutes'}
         benchmarkField={null}
       />
+      <CSVLink style={{verticalAlign: 'text-bottom'}}
+      data={this.state.traveltimes.concat(this.state.headways,this.state.dwells)}
+      filename="data.csv">Download data
+      </CSVLink>
     </div>
   }
 
--- a/src/App.js
+++ b/src/App.js
@@ -294,7 +294,7 @@ class App extends React.Component {
       />
       <CSVLink style={{verticalAlign: 'text-bottom'}}
       data={this.state.traveltimes.concat(this.state.headways,this.state.dwells)}
-      filename="data.csv">Download data
+      filename="dashboard_data.csv">Download data
       </CSVLink>
     </div>
   }
--- a/src/App.js
+++ b/src/App.js
@@ -9,7 +9,6 @@ import AlertBar from './AlertBar';
 import './App.css';
 import Select from './Select';
 import { configPresets } from './constants';
-import { CSVLink } from "react-csv";
 
 const FRONTEND_TO_BACKEND_MAP = new Map([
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -286,11 +287,15 @@ class App extends React.Component {
         isLoading={this.getIsLoadingDataset('dwells')}
         data={this.state.dwells}
         xField={'arr_dt'}
-        xFieldLabel={'Time of day'}
+        xFieldLabel={'Time of day'}
```
nathan-weinberg, line 290: nit: can we remove the whitespace here?

nathan-weinberg (approved):
Code LGTM, can we get a visual of what it looks like @arosgood?

src/App.js
arosgood: Yes haha my bad, removed it.

arosgood:
> Code LGTM, can we get a visual of what it looks like @arosgood?

Here's the visual:
Screen Shot 2021-02-22 at 10 51 23 PM

mathcolo:
Hey @arosgood in letting this percolate more, I am wondering if folks might be confused that all three metrics are in the same CSV. Not sure though?

friendchris:
[dashboard_data.xlsx](https://github.com/transitmatters/t-performance-dash/files/6046196/dashboard_data.xlsx)

I think it would be good to have each chart download separately, if possible. I've attached an Excel with how I think each one should look. Let me know if you all think this makes sense.

nathan-weinberg:
I like the look of the link but I'm wondering if we can make it part of the footer?

arosgood:
Can a csv have multiple sheets like that? I could also make it a .xlsx instead but maybe that's less accessible for people? I was also thinking about adding it to the footer, but Andrew recommended having it above the footer. Maybe it's better to have each metric separate?

arosgood:
New visual:
image

src/line.js
```diff
@@ -38,13 +42,16 @@ const point_colors = (data, metric_field, benchmark_field) => {
...
     else if (ratio <= 1.5) {
-      return '#e5a70b'; //yellow
+      return '#f5ed00'; //yellow
+    }
+    else if (ratio <= 2.0) {
+      return '#c33149'; //red
     }
-    else if (ratio > 1.5) {
-      return '#e53a0b'; //red
+    else if (ratio > 2.0) {
```
mathcolo, line 53: Is this branch based on an old revision possibly? My guess is that you aren't intending to change these colors?

src/line.js
```diff
@@ -3,6 +3,8 @@ import classNames from 'classnames';
 import { Line, Chart } from 'react-chartjs-2';
 import Legend from './Legend';
 import drawTitle from './Title';
+import { CSVLink } from "react-csv";
```
mathcolo, line 6: Should be fine to import these libraries, but we probably need to add attribution to the open source page at https://github.com/transitmatters/t-performance-dash/blob/main/src/OpenSource.js !

src/line.js
```diff
@@ -83,12 +90,21 @@ class LineClass extends React.Component {
     //   }
     // });
     const { isLoading } = this.props;
+
+    const graphData=   
+    <span className="graphData">
```
mathcolo, line 95: Might be cleaner if indented in?

mathcolo (commented):
I think this is close! added a few more comments

src/line.js
arosgood: ah yea mb

src/line.js
arosgood: ok will do ty

src/line.js
arosgood: yes, will change

nathan-weinberg:
@mathcolo @arosgood Looks like this one fell through the cracks a bit, we still interested in getting it in?

ianhi:
Hi,

What more needs to happen to get this in? This would be great.

nathan-weinberg:
Hi @ianhi,

So we discussed this internally and due to the age of this PR we're going to need to close it and refactor the code - if you're interested in helping out please shoot us an email at labs@transitmatters.org and we would be happy to get you in the loop!

\- Nathan

PR #45, opened 2021-02-23
Use 3:30-3:30 days when determining whether to query T API for alerts, and cron two days ago instead

### LINKED ISSUE #3: Graph legends

mathcolo:
I think @friendchris is right about this!

1. We need to do the 3:30am math for determining which alert day to use at request time
2. The job itself was already using transit days, since it was using @friendchris MbtaPerformanceAPI.py
3. That _still_ presents a problem for the job, because the job runs at 5am (or 6am on DST), not 3:31. So between 3:30 and 5am/6am, there will be a gap where it's the new transit day, but we haven't downloaded yet.
4. To fix #3, just use the API for today AND yesterday's "transit days".

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -13,13 +13,14 @@ cors_config = CORSConfig(
 )
 
 
-# Every day at 10am UTC: store alerts from the previous day
+# Every day at 10am UTC: store alerts from the past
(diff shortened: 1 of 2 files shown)


nathan-weinberg (approved):
LGTM

PR #46, opened 2021-02-23
Speed up agg

austinjpaul:
A bit of an update to s3_historical functions to accommodate them handling a date-range and having maybe thousands of entities from multiple service days.
Might still be more to do to improve performance, but this should help at least somewhat.

### DIFF
--- a/server/chalicelib/s3_historical.py
+++ b/server/chalicelib/s3_historical.py
@@ -62,30 +62,21 @@ def headways(stop_id, sdate, edate):
         "benchmark_headway_time_sec": None
     }, only_departures))
 
-# For a given trip ID and a list of events, find when that trip ARR'ed
-
-
-def find_trip_id_arrival(trip_id, event_list):
-    arrival = list(filter(
-        lambda event: event["trip_id"] == trip_id and event["event_type"] in EVENT_ARRIVAL, event_list))
-    if len(arrival) == 1:
-        return arrival[0]["event_time"]
-    else:
-        return None
-
 
 def travel_times(stop_a, stop_b, sdate, edate):
     rows_by_time_a = s3.download_event_range(stop_a, sdate, edate)
     rows_by_time_b = s3.download_event_range(stop_b, sdate, edate)
 
-    only_departures = list(
-        filter(lambda event: event["event_type"] in EVENT_DEPARTURE, rows_by_time_a))
+    departures = filter(lambda event: event["event_type"] in EVENT_DEPARTURE, rows_by_time_a)
+    arrivals = {(event["service_date"], event["trip_id"]): event for event in rows_by_time_b if event["event_type"] in EVENT_ARRIVAL}
+
     travel_times = []
-    for departure in only_departures:
-        dep = departure["event_time"]
-        arr = find_trip_id_arrival(departure["trip_id"], rows_by_time_b)
-        if arr is None:
+    for departure in departures:
+        arrival = arrivals.get((departure["service_date"], departure["trip_id"]))
+        if arrival is None:
             continue
+        dep = departure["event_time"]
+        arr = arrival["event_time"]
 
         dep_dt = datetime.strptime(dep, DATE_FORMAT_MASSDOT)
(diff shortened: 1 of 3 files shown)


server/chalicelib/s3_historical.py
```diff
@@ -9,15 +10,23 @@
...
     dwells = []
-    for i in range(0, len(rows_by_time) - 1):
-        maybe_an_arrival = rows_by_time[i]
-        maybe_a_departure = rows_by_time[i + 1]
+    for maybe_an_arrival, maybe_a_departure in pairwise(rows_by_time):
         # Look for all ARR/DEP pairs for same trip id
-        if maybe_an_arrival["event_type"] in EVENT_ARRIVAL and maybe_a_departure["event_type"] in EVENT_DEPARTURE and maybe_an_arrival["trip_id"] == maybe_a_departure["trip_id"]:
+        if maybe_an_arrival["event_type"] in EVENT_ARRIVAL and \
+           maybe_a_departure["event_type"] in EVENT_DEPARTURE and \
+           maybe_an_arrival["trip_id"] == maybe_a_departure["trip_id"]:
```
austinjpaul, line 28: Are we sure these events will always be arrive/depart? Green Line @ park street comes to mind...

server/chalicelib/s3_historical.py
```diff
@@ -37,55 +46,41 @@ def dwells(stop_id, sdate, edate):
...
+            "current_dep_dt": this["event_time"],
+            "headway_time_sec": headway_time_sec,
+            "benchmark_headway_time_sec": None
+        })
 
-        only_departures[i]["headway_time_sec"] = delta.total_seconds()
-
-    # The first departure of the day has no headway..
-    if len(only_departures) >= 1:
-        only_departures[0]["headway_time_sec"] = 0
```
austinjpaul, line 54: I think we can just not report the first 0 here.

server/chalicelib/s3_historical.py
```diff
@@ -9,15 +10,23 @@
 EVENT_DEPARTURE = ["DEP", "PRD"]
 
 
+def pairwise(iterable):
+    "s -> (s0,s1), (s1,s2), (s2, s3), ..."
```
mathcolo, line 14: Is this supposed to be a comment?

server/chalicelib/s3_historical.py
```diff
@@ -9,15 +10,23 @@
...
+
+
 def dwells(stop_id, sdate, edate):
     rows_by_time = s3.download_event_range(stop_id[0], sdate, edate)
 
     dwells = []
-    for i in range(0, len(rows_by_time) - 1):
-        maybe_an_arrival = rows_by_time[i]
-        maybe_a_departure = rows_by_time[i + 1]
+    for maybe_an_arrival, maybe_a_departure in pairwise(rows_by_time):
```
mathcolo, line 24: It has been a while since I was thinking about this, but does this patch make the assumption that there is exactly one departure per arrival? i.e. that they can be paired up 1:1?

If that is the case, this might not be right?

server/chalicelib/s3_historical.py
mathcolo: Pretty sure they are always either ARR, PRA, DEP, or PRD, yeah

austinjpaul: Ah, I more meant do they always strictly alternate

server/chalicelib/s3_historical.py
austinjpaul: Lol yeah, I guess so! I always copy this function from the itertools docs page without thinking about it much. I really wish it were built-in, since I use it all the time

server/chalicelib/s3_historical.py
austinjpaul: This logic should be identical to the original code in behavior. I do worry a bit about that assumption though.

server/chalicelib/s3_historical.py
austinjpaul: To revisit later (#50)

PR #47, opened 2021-02-26
Fix breakage when alerts are malformed

mathcolo:
Some alerts didn't have proper route info... if that happens, the alert parsing was bailing out early. Example: https://dashboard.transitmatters.org/rapidtransit?config=Orange,70034,70024,2021-02-22

Might as well throw (ha) a try/except at the problem

### DIFF
--- a/server/chalicelib/s3_alerts.py
+++ b/server/chalicelib/s3_alerts.py
@@ -3,11 +3,14 @@ from chalicelib import MbtaPerformanceAPI, s3
 
 
 def routes_for_alert(alert):
-    routes = set()
-    for alert_version in alert["alert_versions"]:
-        for informed_entity in alert_version["informed_entity"]:
-            routes.add(informed_entity["route_id"])
-    return routes
+    try:
+        routes = set()
+        for alert_version in alert["alert_versions"]:
+            for informed_entity in alert_version["informed_entity"]:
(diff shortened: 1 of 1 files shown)


server/chalicelib/s3_alerts.py
```diff
@@ -4,9 +4,13 @@
...
     routes = set()
-    for alert_version in alert["alert_versions"]:
-        for informed_entity in alert_version["informed_entity"]:
-            routes.add(informed_entity["route_id"])
+    try:
+        for alert_version in alert["alert_versions"]:
+            for informed_entity in alert_version["informed_entity"]:
+                routes.add(informed_entity["route_id"])
+    except Exception:
+        pass
```
nathan-weinberg, line 12: Do we want to do some logging here maybe? Something like:
```suggestion
    except Exception as e:
        print("Error fetching alert for route: {e}")
        pass
```
Not sure if this would be useful or not

nathan-weinberg (approved):

server/chalicelib/s3_alerts.py
mathcolo: Oh, yeah! :D

austinjpaul:
Oops, I'm too late. Can we add a date/timestamp to the log message?

PR #48, opened 2021-03-01
We have alerts cached back to November 6, 2017 now

mathcolo:
Over this weekend I downloaded alerts from the T as far back as they would go. That date is November 6, 2017! So I'm updating this logic to only use the live API for today and yesterday.


austinjpaul (approved):
Lgtm!

PR #49, opened 2021-03-06
Add feedback link in footer

mathcolo:
I threw in the envelope 📩 instead of the rocket 🚀. Mailto goes to `labs@` with a simple, but generic subject. At the moment, displays the same on main and beta.

I like @androotubelli suggestion of trying this and seeing whether users actually use it.

image


austinjpaul (approved):
I REALLY liked the rocketship- it was more eye-catching and less... dry, for lack of a better word.
Also, maybe we want to tweak the subject line to be "[Dashboard Feedback] -" In hopes they'll add to it with specifics?

looks good otherwise

nathan-weinberg (approved):
LGTM

mathcolo:
@austinjpaul I like your ideas! Went with it. Can always change if we need to...

PR #52, opened 2021-03-12
Updated docs and moved to new folder

nathan-weinberg:
Also made slight tweak to how app.py fetched TM_FRONTEND_HOST


server/Pipfile
```diff
@@ -17,6 +16,7 @@ boto3 = "*"
 flake8 = "*"
 pandas = "*"
 numpy = "*"
+importlib-resources = "*"
```
mathcolo, line 19: I've seen this before...but it was a dependency of a dependency I thought?

nathan-weinberg: It was probably a flask dependency - flask is no longer used but flake8 stills needs this b/c we are running CI with Python 3.7

mathcolo (approved):

PR #53, opened 2021-03-15
use population std to prevent NaN json error

austinjpaul:
This will prevent only the NaN error when there's only 1 train reported. Perhaps we should handle NaNs more generally?
Also, maybe it makes sense to throw out any data points where only 1 train came that day, since it's almost guaranteed to be an outlier?

The .loc business is a bit more efficient than what we were doing before.


mathcolo (approved):
Looks good! And I agree with you on some simple filtration for n=1s 👍 Might clean up the charts a bit, too.

PR #68, opened 2021-03-30
Made deployed git commit id visible

### LINKED ISSUE #65: Embed deployed Git Commit Hash in Dashboard Beta
In which Nathan takes a crack at TypeScript for the first time in several years

nathan-weinberg:
Fixes #65

Specifically, this adds a new hidden tag to the Dashboard page that should only show when deployed locally or on Beta. The commit id at HEAD on the repo will be fetched from the backend so user's can determine what version of the Dashboard is current deployed (obviously more useful with Beta than locally)

It has been quite a while since I coded any JavaScript, so please be stringent with the reviews from a style and functionality standpoint - I could use the practice.

Here's a pic of what it looks like locally:
image

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -13,7 +13,7 @@ export AWS_PAGER=""
 [[ "$1" = "beta" ]] && FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN_BETA" || FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN"
 [[ "$1" = "beta" ]] && BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN_BETA" || BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN"
 
-git fetch
+git fetch --tags
 if [[ "$1" = "beta" ]]; then
     GIT_ID=`git describe --always --dirty --abbrev=10`
     echo "Deploying git commit id $GIT_ID"
--- a/deploy.sh
+++ b/deploy.sh
@@ -13,6 +13,7 @@ export AWS_PAGER=""
 [[ "$1" = "beta" ]] && FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN_BETA" || FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN"
 [[ "$1" = "beta" ]] && BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN_BETA" || BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN"
 
+git fetch
 if [[ "$1" = "beta" ]]; then
     GIT_ID=`git describe --always --dirty --abbrev=10`
     echo "Deploying git commit id $GIT_ID"
--- a/deploy.sh
+++ b/deploy.sh
@@ -13,16 +13,25 @@ export AWS_PAGER=""
 [[ "$1" = "beta" ]] && FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN_BETA" || FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN"
 [[ "$1" = "beta" ]] && BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN_BETA" || BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN"
 
+if [[ "$1" = "beta" ]]; then
+    GIT_ID=`git describe --always --dirty --abbrev=10`
+else
+    GIT_ID=`git describe --tags --abbrev=0`
+fi
+
 BACKEND_BUCKET=datadashboard-backend$ENV_SUFFIX
 FRONTEND_HOSTNAME=dashboard$ENV_SUFFIX.transitmatters.org # Must match in .chalice/config.json!
 CF_STACK_NAME=datadashboard$ENV_SUFFIX
 
+echo "Deploying git commit id $GIT_COMMIT_ID"
 echo "Starting $CHALICE_STAGE deployment"
 echo "Backend bucket: $BACKEND_BUCKET"
 echo "Hostname: $FRONTEND_HOSTNAME"
 echo "CloudFormation stack name: $CF_STACK_NAME"
 
+# build frontend and patch in commit id
 npm run build
+sed -i "s/version/version $GIT_ID/" ./build/index.html
 
 pushd server/ > /dev/null
 pipenv run chalice package --stage $CHALICE_STAGE --merge-template frontend-cfn.json cfn/
--- a/deploy.sh
+++ b/deploy.sh
@@ -15,15 +15,16 @@ export AWS_PAGER=""
 
 if [[ "$1" = "beta" ]]; then
     GIT_ID=`git describe --always --dirty --abbrev=10`
+    echo "Deploying git commit id $GIT_ID"
 else
     GIT_ID=`git describe --tags --abbrev=0`
+    echo "Deploying git tag $GIT_ID"
 fi
 
 BACKEND_BUCKET=datadashboard-backend$ENV_SUFFIX
 FRONTEND_HOSTNAME=dashboard$ENV_SUFFIX.transitmatters.org # Must match in .chalice/config.json!
 CF_STACK_NAME=datadashboard$ENV_SUFFIX
 
-echo "Deploying git commit id $GIT_COMMIT_ID"
 echo "Starting $CHALICE_STAGE deployment"
 echo "Backend bucket: $BACKEND_BUCKET"
 echo "Hostname: $FRONTEND_HOSTNAME"
@@ -31,7 +32,7 @@ echo "CloudFormation stack name: $CF_STACK_NAME"
 
 # build frontend and patch in commit id
 npm run build
-sed -i "s/version/version $GIT_ID/" ./build/index.html
+sed -i "s/git-id/version $GIT_ID/" ./build/index.html
 
 pushd server/ > /dev/null
 pipenv run chalice package --stage $CHALICE_STAGE --merge-template frontend-cfn.json cfn/
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "lint-backend": "cd server && pipenv run flake8 --ignore=E501,E731 --exclude .chalice",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "postinstall": "cd server && pipenv install"
+    "postinstall": "cd server && pipenv install --python=3.7"
   },
   "proxy": "http://localhost:5000",
   "eslintConfig": {
--- a/server/app.py
+++ b/server/app.py
@@ -1,7 +1,7 @@
 import json
 import os
 import subprocess
-from chalice import Chalice, Cron, CORSConfig
+from chalice import Chalice, Cron, CORSConfig, ConflictError
 from datetime import date, timedelta
 from chalicelib import data_funcs, aggregation, s3_alerts
 
@@ -112,3 +112,5 @@ def get_git_id():
     if TM_FRONTEND_HOST == "localhost":
         git_id = str(subprocess.check_output(['git', 'describe', '--always', '--dirty', '--abbrev=10']))[2:-3]
         return json.dumps({"git_id": git_id})
+    else:
+        raise ConflictError("Cannot get git id from serverless host")
--- a/server/app.py
+++ b/server/app.py
@@ -108,5 +108,7 @@ def dwells_aggregate_route():
 
 @app.route("/git_id", cors=cors_config)
 def get_git_id():
-    git_id = str(subprocess.check_output(['git', 'describe', '--always', '--dirty', '--abbrev=10']))[2:-3]
-    return json.dumps({"git_id": git_id})
+    # Only do this on localhost
+    if TM_FRONTEND_HOST == "localhost":
+        git_id = str(subprocess.check_output(['git', 'describe', '--always', '--dirty', '--abbrev=10']))[2:-3]
+        return json.dumps({"git_id": git_id})
--- a/server/app.py
+++ b/server/app.py
@@ -1,5 +1,6 @@
 import json
 import os
+import subprocess
 from chalice import Chalice, Cron, CORSConfig
 from datetime import date, timedelta
 from chalicelib import data_funcs, aggregation, s3_alerts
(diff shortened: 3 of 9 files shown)


server/app.py
```diff
@@ -89,3 +90,9 @@ def traveltime_route(user_date):
 def alerts_route(user_date):
     date = parse_user_date(user_date)
     return json.dumps(data_funcs.alerts(date, mutlidict_to_dict(app.current_request.query_params)))
+
+
+@app.route("/commit_id", cors=cors_config, methods=['POST'])
+def get_git_commit_id():
+    commit_id = str(subprocess.check_output(['git', 'rev-parse', 'HEAD']))[2:-3]
+    return json.dumps({"commit_id": commit_id})
```
austinjpaul, line 98: @mathcolo does this decorator auto-create a lambda for this? or do we have to list it in .chalice configs also?

nathan-weinberg: In all honesty, I have no idea

src/App.js
```diff
@@ -40,6 +40,19 @@ const showBetaTag = () => {
   beta_tag.innerText = "Beta";
 };
 
+async function showCommitId() {
```
idreyn, line 43: I was halfway through a rant about rewriting this as a React component before I noticed that the `commit-tag` span you've added isn't even part of the React app — so that's right out. Honestly, I think this looks fine! I also think it would be fine to inline the contents of `getCommitId` but totally okay either way.

fwiw — we don't totally follow this convention here but it's most typical to camelcase your JS variables.

src/App.js
```diff
@@ -68,6 +81,10 @@ class App extends React.Component {
       showBetaTag();
     }
 
+    if (window.location.hostname !== "dashboard.transitmatters.org") {
```
idreyn, line 84: Only real nitpick here is that this should go in `componentDidMount` below.

idreyn (commented):
Just weighing in on the JS side of this, no idea what needs to be adapted to get this to run on a lambda, but it'll be a cool feature to have!

server/app.py
austinjpaul: I'm willing to bet this won't work. Per discussion in slack, i think editing something in deploy.sh is going to be easier.
Also, I personally like using `git describe --all --long --dirty --abbrev=10` (or similar) to get a more descriptive label than `rev-parse` would return.

nathan-weinberg: Compared the output, curious about a few things:
```
$ git rev-parse HEAD                            => 0fd9cfcbba06103e8992d9eb1a83dd70d0b64408
$ git describe --all --long --dirty --abbrev=10 => heads/master-0-g0fd9cfcbba
```
Specifically, what is the significance of `-0-g`?
Also, this seems like it chops off the first char of the hash

src/App.js
nathan-weinberg: Can do, do you want me to move the Beta tag func as well? I modeled this after that

server/app.py
```diff
@@ -89,3 +90,9 @@ def traveltime_route(user_date):
 def alerts_route(user_date):
     date = parse_user_date(user_date)
     return json.dumps(data_funcs.alerts(date, mutlidict_to_dict(app.current_request.query_params)))
+
+
+@app.route("/commit_id", cors=cors_config, methods=['POST'])
```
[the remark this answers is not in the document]
nathan-weinberg: Why GET over POST in this case?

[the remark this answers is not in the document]
mathcolo: POST requests typically come with a body that contains data to update something on the server—you’re not updating anything

FYI POST can also trip the browser into “do you want to submit this form again?” on a back button press, although that won’t happen if JavaScript (as opposed to page navigation) is making the request

nathan-weinberg:
Kept the JS around just for local dev + in case the app is ever run not on Lambda; current implementation should work for both
Haven't gotten to test with Beta yet

src/App.js
```diff
@@ -40,6 +40,14 @@ const showBetaTag = () => {
   beta_tag.innerText = "Beta";
 };
 
+async function showCommitId() {
+  const commitTag = document.querySelector(".commit-tag");
+  const response = await fetch(APP_DATA_BASE_PATH + '/commit_id');
```
austinjpaul, line 45: I think we don't want to run this fetch command on lambda, since it can't do anything with it.
But this request is the only reason it's working locally, right? because `deploy.sh` isn't used to run locally.

I'm not a huge fan of behavior that is deployment-dependent (understanding that some are unavoidable), but is there any way to unify the approach here? I like the static approach (`sed`) better than the dynamic (REST request). Maybe throwing it in an npm build step? I don't want to add too much opacity though.

Apologies if I've misunderstood what this is doing.

austinjpaul: OH i also just realized running deploy would add changes in `index.html` that we would then have to be careful not to accidentally commit.

nathan-weinberg: Your first assumption is correct; honestly due to the nature of the Lambda functionality I can't think of a way to unify the two approaches but am open to ideas!

As for the `index.html` git tracking I suppose a somewhat hacky but functional solution would be a second `sed` command after the deployment has happened to reset the file contents before the shell script exits?

mathcolo: Which `index.html` are we talking about? The one at `public/index.html` gets copied to `build/` before it gets uploaded.

nathan-weinberg: I'm talking about the one in `public/` - maybe we can do the `sed` on the one in `build/`? When is it generated in the script?

mathcolo: I think you can! If you do whatever you want to the frontend in `build/` between `npm run build` and the `aws s3 sync...` you should be good 👍

nathan-weinberg: I'll give it a try!

nathan-weinberg:
Now visible on https://dashboard-beta.transitmatters.org/

nathan-weinberg:
Just realized this implementation may mean this shows in prod - need to make a few more tweaks

mathcolo:
@nathan-weinberg You know how some websites just always have a version number listed at the bottom? I wonder if it would be fine on prod too if it was smaller, grey text or something?

Not that I don't want you to tweak the script to disable it on prod, but it might actually be useful on prod too.

nathan-weinberg:
@mathcolo @austinjpaul @idreyn @arosgood Made some changes to the PR, should now show latest git tag on Prod site and Commit ID on Beta site (up now if you want to see) and localhost

server/app.py
```diff
@@ -103,3 +104,9 @@ def dwells_aggregate_route():
 
     response = aggregation.dwells_over_time(sdate, edate, stop)
     return json.dumps(response, indent=4, sort_keys=True, default=str)
+
+
+@app.route("/git_id", cors=cors_config)
+def get_git_id():
+    git_id = str(subprocess.check_output(['git', 'describe', '--always', '--dirty', '--abbrev=10']))[2:-3]
+    return json.dumps({"git_id": git_id})
```
austinjpaul, line 112: is there a way to tell lambda not to create this route, but leave it for local (or other server) deploys only?

austinjpaul (commented):
Nice improvements. I worry about tag behavior and `git describe --tags`. We should come up with an explicit plan for how to deal with these.

package.json
```diff
@@ -30,7 +30,7 @@
     "lint-backend": "cd server && pipenv run flake8 --ignore=E501,E731 --exclude .chalice",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "postinstall": "cd server && pipenv install"
+    "postinstall": "cd server && pipenv install --python=3.7"
```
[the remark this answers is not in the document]
nathan-weinberg: It knows what version it wants to user, but it will still only search for whatever interpreter `python` points at (in my case, Python 3.9)

This way it forces pipenv to look for the actual `python3.7` interpreter

deploy.sh
```diff
@@ -13,6 +13,14 @@ export AWS_PAGER=""
 [[ "$1" = "beta" ]] && FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN_BETA" || FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN"
 [[ "$1" = "beta" ]] && BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN_BETA" || BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN"
 
+if [[ "$1" = "beta" ]]; then
+    GIT_ID=`git describe --always --dirty --abbrev=10`
+    echo "Deploying git commit id $GIT_ID"
+else
+    GIT_ID=`git describe --tags --abbrev=0`
+    echo "Deploying git tag $GIT_ID"
```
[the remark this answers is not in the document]
nathan-weinberg: A `git fetch` should do it, I'll add that to `deploy.sh`

server/app.py
nathan-weinberg: @mathcolo?

mathcolo: I looked in the Chalice documentation briefly, and didn't find any way to easily do that :( Throwing a conditional on `TM_FRONTEND_HOST` in the body should work though? It defaults to "localhost" in the .get() on line 11

That should be enough to prevent `git` from getting executed on remote

deploy.sh
```diff
@@ -22,7 +31,9 @@ echo "Backend bucket: $BACKEND_BUCKET"
 echo "Hostname: $FRONTEND_HOSTNAME"
 echo "CloudFormation stack name: $CF_STACK_NAME"
 
+# build frontend and patch in commit id
 npm run build
+sed -i "s/git-id/version $GIT_ID/" ./build/index.html
```
mathcolo, line 36: It's....beautiful

src/App.js
```diff
@@ -45,6 +45,14 @@ const showBetaTag = () => {
   beta_tag.innerText = "Beta";
 };
 
+async function getGitId() {
+  const commitTag = document.querySelector(".version");
+  const response = await fetch(APP_DATA_BASE_PATH + '/git_id');
```
mathcolo, line 50: For robustness you may want to surround fetch calls in try/catch, and fail gracefully 😃

src/App.css
```diff
@@ -40,6 +40,11 @@
   color: #333;
 }
 
+.version {
+  font: 0.9rem Inconsolata, monospace;
```
mathcolo, line 44: Most probably don't have this font (myself included)...should perhaps either bring it in via web font URL or remove?

server/app.py
```diff
@@ -103,3 +104,9 @@ def dwells_aggregate_route():
 
     response = aggregation.dwells_over_time(sdate, edate, stop)
     return json.dumps(response, indent=4, sort_keys=True, default=str)
+
+
+@app.route("/git_id", cors=cors_config)
+def get_git_id():
+    git_id = str(subprocess.check_output(['git', 'describe', '--always', '--dirty', '--abbrev=10']))[2:-3]
```
mathcolo, line 111: Is there any chance that `[2:-3]` will depend on a particular version of git? If others on our team haven't made a concerted effort to use git latest, not sure what will happen?

nathan-weinberg: Don't think it should make a difference based on git version

src/App.css
nathan-weinberg: Aha this is what I get for trying to be fancy :laughing: what CSS were you originally using for this tag, I can just use that

server/app.py
nathan-weinberg: Something like `if TM_FRONTEND_HOST === "localhost":`?

src/App.css
mathcolo: LOL you are allowed! :-) https://github.com/transitmatters/t-performance-dash/blob/main/public/index.html#L38

server/app.py
mathcolo: 👍

nathan-weinberg:
Changes can be seen on https://dashboard-beta.transitmatters.org/

src/App.js
```diff
@@ -45,6 +45,21 @@ const showBetaTag = () => {
   beta_tag.innerText = "Beta";
 };
 
+async function getGitId() {
+  const commitTag = document.querySelector(".version");
+  var git_id = "unknown";
```
mathcolo, line 50: `var` is taboo these days, suggest `let` instead :)

server/app.py
```diff
@@ -103,3 +104,11 @@ def dwells_aggregate_route():
 
     response = aggregation.dwells_over_time(sdate, edate, stop)
     return json.dumps(response, indent=4, sort_keys=True, default=str)
+
+
+@app.route("/git_id", cors=cors_config)
+def get_git_id():
+    # Only do this on localhost
+    if TM_FRONTEND_HOST == "localhost":
```
mathcolo, line 112: Probably want an else here that throws a 4xx error, client will probably hang if server doesn't reply? https://aws.github.io/chalice/topics/views.html#error-handling

src/App.js
```diff
@@ -45,6 +45,21 @@ const showBetaTag = () => {
...
+  try {
+    const response = await fetch(APP_DATA_BASE_PATH + '/git_id');
+    const commitJson = await response.json();
+    git_id = commitJson.git_id;
+  }
+  catch (error) {
+    console.log(`Error fetching Git ID: ${error}`);
```
mathcolo, line 57: Prefer `console.error` for errors so it gets treated appropriately in the browser console

mathcolo (commented):
Some more comments. getting close, thanks for hanging in there! 😁

nathan-weinberg:
@mathcolo No worries thanks for the feedback!

server/app.py
nathan-weinberg: Based on what Chalice offers I'm going to go with 409 based off https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

(would like to do 405 really but I'm not seeing any native Chalice classes that support it and I'm not writing a custom one)

nathan-weinberg:
@austinjpaul Can we merge this?

deploy.sh
```diff
@@ -13,6 +13,15 @@ export AWS_PAGER=""
 [[ "$1" = "beta" ]] && FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN_BETA" || FRONTEND_CERT_ARN="$TM_FRONTEND_CERT_ARN"
 [[ "$1" = "beta" ]] && BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN_BETA" || BACKEND_CERT_ARN="$TM_BACKEND_CERT_ARN"
 
+git fetch
```
austinjpaul, line 16: i think this might need to be `git fetch --tags` in order to pull the tags.

austinjpaul (approved):
One minor fix, but then the rest looks good.

PR #75, opened 2021-04-04
Whole day x

austinjpaul:
Addresses #2 with some plugins. Hopefully it's not too ugly.
I also hard-coded the alerts timeframe, since traveltimes don't always line up (esp. with historical trip-id issues)

However.

This also introduces an error when switching out of multi-day mode into single-day mode.
Perhaps @mathcolo has some tricks to help?

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -313,16 +313,23 @@ class App extends React.Component {
   }
 
   suggestXRange() {
-    const today = `${this.state.configuration.date_start}T00:00:00`;
+    if (this.getTimescale() === 'hour') {
+      const today = `${this.state.configuration.date_start}T00:00:00`;
 
-    var low = new Date(today);
-    low.setHours(6,0);
+      var low = new Date(today);
+      low.setHours(6,0);
 
-    var high = new Date(today);
-    high.setDate(high.getDate() + 1);
-    high.setHours(1,0);
+      var high = new Date(today);
+      high.setDate(high.getDate() + 1);
+      high.setHours(1,0);
 
-    return [low, high];
+      return [low, high];
+    } else {
+      const start = `${this.state.configuration.date_start}T00:00:00`;
+      const end = `${this.state.configuration.date_end}T00:00:00`;
+
+      return [new Date(start), new Date(end)];
+    }
   }
 
   chartTimeframe() {
--- a/src/App.js
+++ b/src/App.js
@@ -316,10 +316,10 @@ class App extends React.Component {
     if (this.getTimescale() === 'hour') {
       const today = `${this.state.configuration.date_start}T00:00:00`;
 
-      var low = new Date(today);
+      let low = new Date(today);
       low.setHours(6,0);
 
-      var high = new Date(today);
+      let high = new Date(today);
       high.setDate(high.getDate() + 1);
       high.setHours(1,0);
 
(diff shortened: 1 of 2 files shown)


austinjpaul:
I forgot about the 2nd part of this: make the aggregate x always be the selected range, even when data is not available for some days. A similar approach should work, just have to work out the bugs.

src/line.js
```diff
@@ -177,7 +177,11 @@ class LineClass extends React.Component {
                       display: true,
                       fontSize: 14,
                       labelString: this.props.xFieldLabel
-                    }
+                    },
+                    // make sure graph shows /at least/ suggestedXRange, either end will extend to not hide points.
+                    afterDataLimits: (axis) => {if (this.props.isLoading) return; // prevents weird sliding animation
+                                                axis.min = Math.min(axis.min, this.props.suggestedXRange[0]);
```
mathcolo, line 183: Sorry if I'm missing something, but can these statements just be gated to daily mode only?

src/App.js
```diff
@@ -311,12 +312,37 @@ class App extends React.Component {
...
 
-  chartTimeframe() {
-    const travel_times = this.state.traveltimes;
-    if (travel_times.length > 0) {
-      return [new Date(travel_times[0].dep_dt), new Date(travel_times[travel_times.length - 1].dep_dt)];
+  suggestXRange() {
+    if (this.getTimescale() === 'hour') {
+      const today = `${this.state.configuration.date_start}T00:00:00`;
+
+      var low = new Date(today);
```
mathcolo, line 319: let 😃

src/App.js
```diff
@@ -311,12 +312,37 @@ class App extends React.Component {
...
+  }
+
+  chartTimeframe() {
+    const today = `${this.state.configuration.date_start}T00:00:00`;
+
+    var low = new Date(today);
+    low.setHours(5, 30)
+
+    var high = new Date(today);
+    high.setDate(high.getDate() + 1);
```
mathcolo, line 342: This handles `32` and `31` correctly right?

src/line.js
austinjpaul: I want to have this behavior for multi-day mode also
tmp

src/App.js
austinjpaul: yep, it rolls over month and year.

austinjpaul:
Bug fixed- this is working now!
tmp
tmp2

src/App.js
austinjpaul: alternatively, i could just call setHours(25) ?

austinjpaul: lmk if you think something like this reads better: 
```      const { date_start } = this.state.configuration;

      const low = new Date(`${date_start}T06:00:00`);
      const high = new Date(`${date_start}T24:00:00`);
      high.setHours(1);

      return [low, high];
```

src/App.js
```diff
@@ -311,12 +312,37 @@ class App extends React.Component {
...
 
-  chartTimeframe() {
-    const travel_times = this.state.traveltimes;
-    if (travel_times.length > 0) {
-      return [new Date(travel_times[0].dep_dt), new Date(travel_times[travel_times.length - 1].dep_dt)];
+  suggestXRange() {
+    if (this.getTimescale() === 'hour') {
+      const today = `${this.state.configuration.date_start}T00:00:00`;
+
+      let low = new Date(today);
```
mathcolo, line 319: Would you mind adding some comments to indicate that this is "6am today" and "1am tomorrow"?

src/App.js
mathcolo: Hmmm no I think the way you have it is probably more clear with comments than this trick, but nice though lol.

mathcolo (approved):
Looks good!! Thanks for fixing this. A few comments, but otherwise I think good to merge.

Issue #69, opened 2021-03-31
Cancel requests on config change


austinjpaul:
If you request several months worth of data, and then quickly switch to single day mode, the single day info will get erased when the aggregate requests finally return.

PR #79, opened 2021-04-11
Aggregation: Cancel existing requests on config change

mathcolo:
Fix https://github.com/transitmatters/t-performance-dash/issues/69

Add an AbortController array, so new fetches signal all existing fetches first.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -99,6 +99,7 @@ class App extends React.Component {
     this.permittedRange = this.permittedRange.bind(this);
 
     this.progressTimer = null;
+    this.fetchControllers = [];
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -256,28 +259,44 @@ class App extends React.Component {
...
   }
 
   download() {
+    // End all existing fetches
+    while(this.fetchControllers.length > 0) {
+      this.fetchControllers.shift().abort();
+    }
+
+    const controller = new AbortController();
+    this.fetchControllers.push(controller);
```
austinjpaul, line 276: it looks like fetchControllers should only ever have 1 thing it it, but i like the just-in-case list approach.

austinjpaul (approved):
This is clean and working great! LGTM

src/App.js
mathcolo: Yeah probably unnecessary futureproofing!

Issue #76, opened 2021-04-04
Back button should update configuration displayed


austinjpaul:
going from multi-day mode to single-day mode and then pressing browser's 'back' button leaves teh configuration confused about where it is.

PR #80, opened 2021-04-11
[Short and sweet] Fix station picker deserialization

mathcolo:
Fixes issue #76. This happens on `main`, so this PR is against main.

The station picker cannot deserialize station info from the URL right now, because we cannot match station indices based on objects—they're not the same. Have to compare stop names themselves instead.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -2,7 +2,7 @@ import React, { useRef } from "react";
 import classNames from 'classnames';
 
 const Select = props => {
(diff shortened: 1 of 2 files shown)


mathcolo:
Oops, by "from the URL" I actually mean "from browser state history" :-(

idreyn (approved):
LGTM! 👍

austinjpaul (approved):
yesssss thank you!! this had been bugging me for a while.

PR #83, opened 2021-04-12
Update presets March 2021

austinjpaul:
2 things:
- I picked maybe the worst possible days for the T. Do we want to be less harsh?
- I added a tiny bit of context as to why these days are interesting in the picker, which I find useful, but maybe it's too heavy-handed?


mathcolo (approved):
This is awesome! And I think the description is neat too. Let's give it a whirl

PR #84, opened 2021-04-13
Add northern OL preset April 2021

mathcolo:
Add an option to the presets for the April 2021 OL slow zone, with a dynamic end date that's always today.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1119,7 +1119,7 @@ export const configPresets = [
 		value: createConfigPresetValue("Orange", "Community College", "North Station", '2020-11-01', '2021-02-17'),
 	},
 	{
-		label: "[New!] Spring 2020 — Green Line (C branch) COVID effect",
+		label: "[New!] Spring 2020 — Green Line (C-branch) COVID-19 pandemic effect",
 		value: createConfigPresetValue("Green", "Englewood Avenue", "Saint Marys Street", '2020-01-31', '2020-07-05'),
 	},
 	{
--- a/src/constants.js
+++ b/src/constants.js
@@ -1094,18 +1094,26 @@ export const stations = {
 	}]
 };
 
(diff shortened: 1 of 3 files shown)


src/constants.js
```diff
@@ -1094,18 +1094,26 @@ export const stations = {
 	}]
 };
 
-const createConfigPresetValue = (line, fromStationName, toStationName, date_start) => {
+const createConfigPresetValue = (line, fromStationName, toStationName, date_start, date_end = undefined) => {
```
austinjpaul, line 1097: is undefined okay here? or would "" be safer?

src/constants.js
```diff
@@ -1094,18 +1094,26 @@ export const stations = {
...
 };
 
+// Returns today's EST/EDT date in a format like "2021-04-12". Thanks Canada!
+const TODAY = new Date().toLocaleString('fr-CA', { timeZone: 'America/New_York' }).split(" ")[0];
+
 export const configPresets = [
+	{
+		label: "[New!] April 2021 — Orange Line slow zone",
+		value: createConfigPresetValue("Orange", "Oak Grove", "Wellington", '2021-03-01', TODAY),
+	},
```
austinjpaul, line 1116: Let's add one more aggregation pre-set, maybe one of the ones from the blog-post?

mathcolo: Great idea; went for it!

src/constants.js
mathcolo: undefined is better here, because it matches the default state more closely...`date_end` starts out as undefined until it gets defined by either the URL or user date picker selection

austinjpaul (approved):
if orange slow zones are getting old, we could also demonstrate headway increases due to the service cuts in march. But I digress... lgtm!

PR #86, opened 2021-04-13
Correct trailing comma in Orange Line Slowdown

crschmidt:
Trailing comma in the date messes up queries.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1107,7 +1107,7 @@ const createConfigPresetValue = (line, fromStationName, toStationName, date_star
 };
 
 // Returns today's EST/EDT date in a format like "2021-04-12". Thanks Canada!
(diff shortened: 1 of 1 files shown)


austinjpaul:
Thanks for opening this, Chris! 
We tested with the `" "` last night on firefox, but the existing code is breaking on mobile chrome, at least.
What browser are you using?
Changing this to `","` causes a 500 to occur on Firefox.
Maybe we need to split on "," and " " both?

crschmidt:
I've created a new commit that just uses ISO date format, which should be consistent (at the cost of not having support for timezones, so in the evening it will wrap into the next day since it's GMT rather than Eastern time, but that seems pretty minimal.)

austinjpaul:
Nice. I agree that this is better- locale strings are very inconsistent (by design).
Wrapping to tomorrow shouldn't be an issue, i'm gonna go ahead and merge!

austinjpaul (approved):

PR #87, opened 2021-04-16
Fix s3_alerts.store_alerts to use new get_api_data argument order

mathcolo:
Automatic overnight alert pull was broken because `get_api_data` changed— quick fix! :-)


austinjpaul (approved):
Oops. Nice catch!

PR #88, opened 2021-04-18
Add backend healthcheck

mathcolo:
The train tracker issue this morning reminded me that the dashboard needs a real health check. This checks to make sure the API key has a non-zero length, and that it can get alerts from S3. 500 error if a check fails, and a number output indicating which one.

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1,8 +1,8 @@
 import json
 import os
-from chalice import Chalice, Cron, CORSConfig
+from chalice import Chalice, Cron, CORSConfig, Response
 from datetime import date, timedelta
-from chalicelib import data_funcs, aggregation, s3_alerts
+from chalicelib import data_funcs, aggregation, s3_alerts, secrets
 
 app = Chalice(app_name="data-dashboard")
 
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

austinjpaul:
This looks decent. Is the goal to point uptime bot at it or something so we get slack notifications? In general, i'd love a log of 500s (and maybe 400s?) that occur- I tried to find such things in aws, but it made me cry. Do you know of any way to get notifications of these? either in slack or a weekly digest email or something?

server/app.py
```diff
@@ -47,6 +47,31 @@ def mutlidict_to_dict(mutlidict):
     return res_dict
 
 
+@app.route("/healthcheck", cors=cors_config)
+def healthcheck():
+    # These functions must return True or False :-)
+    checks = [
+        lambda: len(secrets.MBTA_V2_API_KEY) > 0,
+        lambda: "2020-11-07 10:33:40" in json.dumps(data_funcs.headways(date(year=2020, month=11, day=7), ["70061"]))
```
austinjpaul, line 55: this is an old date, so testing s3 access i presume. Should we test mbta's api as well? and/or alert cache thingys?

austinjpaul (approved):

nathan-weinberg:
Any way to incorporate this into a CI check? We can always brainstorm this later ofc

server/app.py
mathcolo: Yes we can! I'll merge this and can do a subsequent PR with some additional features

mathcolo:
@nathan-weinberg You bet; let's add more soon

Issue #82, opened 2021-04-12
Set API key through deploy.sh deployment


mathcolo:
We're manually setting the API key environment variable on lambdas right now. This should be handled through deploy.sh.

nathan-weinberg:
Could we just have the API key set as an environmental variable/GitHub Secret and have `deploy.sh` set it in Lambda with [this](https://docs.aws.amazon.com/cli/latest/reference/lambda/update-function-configuration.html) or am I oversimplifying?

mathcolo:
@nathan-weinberg No that's almost it! It should be set as a part of the initial configuration though; doesn't have to be a separate usage of `update-function-configuration`. Similarly to how ansible is declarative, the goal with this serverless setup is also to do things in a declarative manner and prevent drift between the CloudFormation config and reality.

The chalice system here works by first assuming a series of defaults (configurable by chalice's config in `server/.chalice/config.json`). That's enough for backend-only projects, but we have an entirely separate frontend as well in here, so that's where [frontend-cfn.json](https://github.com/transitmatters/t-performance-dash/blob/main/server/frontend-cfn.json) comes in. `deploy.sh` [asks chalice](https://github.com/transitmatters/t-performance-dash/blob/main/deploy.sh#L39) to make a serverless config, but ALSO overlaps frontend-cfn.json on top of it, to add the necessary frontend resources that chalice won't do for us. That also makes for a handy place to inject the API key into the lambda, so it doesn't have to be hard coded in `config.json`.

PR #94, opened 2021-08-14
Set API key through deploy.sh

mathcolo:
Deployments sometimes cause the backend to break because the API key isn't set automatically. It's now passed in through GitHub secrets into deploy.sh like it's supposed to be. Sorry this took so long @nathan-weinberg!!

Fixes https://github.com/transitmatters/t-performance-dash/issues/82


nathan-weinberg (approved):
LGTM - if you want to test you can make changes to the deploy beta workflow but I think this should be fine either way 👍

mathcolo:
@nathan-weinberg Already did; thanks for the approval!

PR #99, opened 2021-09-24
flexible direction words

austinjpaul:
Slowly paving the way for bus.
also note: green line is now east/westbound instead of north/south.
(perhaps we reconsider once GLX opens?)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -184,10 +184,11 @@ class App extends React.Component {
       date_start,
       date_end,
     } = this.state.configuration;
+    const { fromStopIds, toStopIds } = get_stop_ids_for_stations(from, to);
     const parts = [
       line,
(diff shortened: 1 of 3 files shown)


nathan-weinberg:
I uh have no idea why the deploy action ran

austinjpaul:
I accidentally pushed this commit to main at first, which spawned deploy job 47, which I cancelled. Preston and I fixed the main branch (deploy 48 was successful), and then I pushed the commit to this branch and opened the PR, where it spawned lint job 179.
My guess is since the git hash is the same, github thinks job 47 belongs to this PR even though it was from a different branch.

idreyn (approved):
looks good to me (but it's been a while since I've seen any of this)

mathcolo (approved):

PR #100, opened 2021-10-12
Aggregation refactor

### LINKED ISSUE #63: Separate aggregation and non-aggregation code paths

austinjpaul:
This has been a bit overdue- it should resolve #63 
Chart types are now separate, which should make adding a new chart type for buses and peak/off-peak stuff easier. 

Hopefully things are cleaner/more readable, but let me know if I've gone too far.

Major things include:
- turning our `Line` chart into two types: `SingleDayLine` and `AggregateLine`. Most common settings are factored out into defaults, and so each chart can have more specific logic built in. Git diff is making a mess of this, it might be easier to read just the final file.
- E.g for the above: the suggestedXRange logic was moved out of App.js and into each chart's axis settings (as an `afterDraw` function).
- There was a bug where picking a date range >8months and then quitting multi-day mode would leave you stuck with the 8-month error, even on refresh. A slight logic change has solved that issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "classnames": "^2.2.6",
     "concurrently": "^5.2.0",
     "flatpickr": "4.5.7",
+    "lodash.merge": "^4.6.2",
     "react": "^16.13.1",
     "react-chartjs-2": "^2.11.1",
     "react-dom": "^16.13.1",
--- a/src/App.js
+++ b/src/App.js
@@ -113,7 +113,7 @@ class App extends React.Component {
     this.setIsLoadingDataset = this.setIsLoadingDataset.bind(this);
     this.getIsLoadingDataset = this.getIsLoadingDataset.bind(this);
     this.getDoneLoading = this.getDoneLoading.bind(this);
-    this.getTimescale = this.getTimescale.bind(this);
+    this.isAggregation = this.isAggregation.bind(this);
     this.progressBarRate = this.progressBarRate.bind(this);
     this.restartProgressBar = this.restartProgressBar.bind(this);
     this.permittedRange = this.permittedRange.bind(this);
@@ -258,11 +258,11 @@ class App extends React.Component {
     return this.state.datasetLoadingState[name];
   }
 
-  getTimescale() {
+  isAggregation() {
     if (this.state.configuration.date_end) {
-      return "day";
+      return true;
     }
-    return "hour";
+    return false;
   }
 
   restartProgressBar() {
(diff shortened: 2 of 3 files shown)


src/App.js
```diff
@@ -258,11 +258,11 @@ class App extends React.Component {
     return this.state.datasetLoadingState[name];
   }
 
-  getTimescale() {
+  isAggregation() {
```
idreyn, line 261: Or just `return !!this.state.configuration.date_end`

src/App.js
```diff
@@ -439,65 +416,80 @@ class App extends React.Component {
...
-        xField={is_aggregation ? 'arr_dt' : 'service_date'}
-        xFieldLabel={is_aggregation ? 'Time of day' : 'Day'}
-        xFieldUnit={timescale}
-        suggestedXRange={this.suggestXRange()}
-        yField={is_aggregation ? 'dwell_time_sec' : '50%'}
-        yFieldLabel={'Minutes'}
-        benchmarkField={null}
-      />
-    </div>
+    if (!this.isAggregation()) {
```
idreyn, line 419: Why not put the positive condition in the `if` clause for a slight boost in readability?

```ts
if (this.isAggregation()) {

}
```

idreyn (approved):
I bet Preston is more up to date on the overall context of this change than I am, but looks good to me!

package.json
```diff
@@ -8,6 +8,7 @@
     "classnames": "^2.2.6",
     "concurrently": "^5.2.0",
     "flatpickr": "4.5.7",
+    "lodash.merge": "^4.6.2",
```
mathcolo, line 11: Is this a package that contains just lodash merge?

src/App.js
mathcolo: +1

src/line.js
```diff
@@ -56,147 +57,216 @@ const point_colors = (data, metric_field, benchmark_field) => {
...
+      </div>
+      <LegendLongTerm />
       </div>
     );
   }
 }
 
-export default LineClass;
+export { SingleDayLine, AggregateLine };
+
```
mathcolo, line 272: Note the lack of newline

src/line.js
```diff
@@ -56,147 +57,216 @@ const point_colors = (data, metric_field, benchmark_field) => {
   });
 }
 
-class LineClass extends React.Component {
+merge(defaults, {
```
mathcolo, line 60: This is awesome, thank you!!

mathcolo (approved):
Austin this is awesome! 🚀 Few little random comments

package.json
austinjpaul: i guess so? lodash.merge is what was recommended by the react2-chartjs docs for this task.

PR #102, opened 2021-11-14
hotfix for dst

austinjpaul:
This might could be improved later, but works for now.
Also switched requests so we can log the useful message from the server along with the 400 client error info. e.g. `{ "error": { "message": "Missing required query parameter: api_key" }}`


austinjpaul:
Merging so the site works again. Feel free to comment/revisit code for cleanup tho

PR #106, opened 2021-11-25
New Daily Aggregate chart type

austinjpaul:
So here's the deal:
There are a TON of code changes flying every which way from the bus endeavor: (lots of new backend code, minor refactors everywhere, there might be an overhaul to `configuration` coming up *shudders*). I'm slowly learning my way around Reactjs, but I'm still very much a novice.
So with all that in mind, I'm creating a `bus_staging` branch that I will gradually PR stuff into before it is ready for a deployment to main. (This way `beta` can keep the existing bus code without me having to ask for a review of everything at once.)

Since this is an an independent piece from everything else, so I figure now's a good time to ask for feedback before I take it any further.

image

As @friendchris and @ofsevit have requested, we are aggregating data not over time, but by time of day to generate a chart like the above.

Some technical notes:
- There is a new endpoint `traveltimes2` that returns aggregate both over time and daily (hourly?). I have left the existing backend endpoint as is, since slowzones is consuming it and I don't want to thinking about having too many things in flight.
- I have pulled the long list of charts to render out of `App.js` and into `ChartSet.js` just to help improve readability. However, I'm not sure it helps right now, but it will help separate logic when buses get different chartsets.
- Our Aggregation chart now has 2 types: AggregationOverTime and AggregationDaily. Both of these include a RadioField selector and their AggregationOverTimeLine/AggregationDailyLine chart component, respectively. Hopefully this isn't tremendously confusing, I'm having trouble naming things.
- I've done my best with the css, but it's not perfect. Particularly, the colorful legend for single-day mode in mobile could still use some work, but maybe it's good enough for now.

I know this is a lot... Thanks so much for your time!!


austinjpaul:
Bit of an update to how things are looking. 
Again, not asking for this to go straight into main, just that it's ready for more bus changes to be built on top of it.
image

friendchris (approved):
This looks good to me. Only change: I would make “Work days” —> “Weekday” and remove the S from “Weekends/Holidays” to just “Weekday/Holiday”

server/chalicelib/aggregation.py
```diff
@@ -34,16 +35,24 @@ def faster_describe(grouped):
     q3.name = '75%'
     # TODO: we can take this out if we filter for 'median' in the front end
     stats.rename(columns={'median': '50%'}, inplace=True)
-    stats = pd.concat([stats, q1, q3, std], axis=1)
+    stats = pd.concat([stats, q1, q3, std], axis=1).reset_index()
```
mathcolo, line 38: Could you explain why you did this?

src/AggregateCharts.js
```diff
@@ -0,0 +1,135 @@
...
+    location
+    isLoading
+
+  */
+  constructor(props) {
+    super(props);
+    this.onChangeValue = this.onChangeValue.bind(this);
+
+    this.state = {
+      title: `${props.title}`,
```
mathcolo, line 94: String interpolation not needed?

src/AggregateCharts.js
```diff
@@ -0,0 +1,135 @@
...
+
+  options = [
+    {value: "all", label: "All times", titleSuffix: ""},
+    {value: "am_peak", label: "AM Peak", titleSuffix: "(AM Peak only)"},
+    {value: "pm_peak", label: "PM Peak", titleSuffix: "(PM Peak only)"},
+    {value: "off_peak", label: "Off-peak", titleSuffix: "(Off-peak only)"}
+  ]
+
+  onChangeValue(value) {
+    let titleSuffix = this.options.find(x => x.value === value).titleSuffix || ""
```
mathcolo, line 107: const?

src/AggregateCharts.js
```diff
@@ -0,0 +1,135 @@
...
+        <RadioForm onChange={this.onChangeValue}
+          options={this.options}
+          defaultValue={this.state.selected}
+        />
+      </AggregateOverTimeLine>
+    )
+  }
+}
+
+export { AggregateDaily, AggregateOverTime }
```
mathcolo, line 135: newline

src/App.css
```diff
@@ -258,7 +262,9 @@ button:disabled,
 @media screen and (min-width: 770px) {
   .chart {
     display: grid;
-    grid-template: 'chart legend';
+    grid-template: 
+      'chart legend'
```
mathcolo, line 266: Does this look ok on mobile?

src/App.js
```diff
@@ -201,7 +201,13 @@ class App extends React.Component {
     if (this.state.configuration.date_end) {
       options["start_date"] = this.state.configuration.date_start;
       options["end_date"] = this.state.configuration.date_end;
-      url = new URL(`${APP_DATA_BASE_PATH}/aggregate/${name}`, window.location.origin);
+
+      var method = name;
```
mathcolo, line 205: Prefer let

server/chalicelib/aggregation.py
```diff
@@ -56,23 +65,62 @@ def travel_times_over_time(sdate, edate, from_stop, to_stop):
...
+
+def travel_times_all(sdate, edate, from_stop, to_stop):
+    df = aggregate_traveltime_data(sdate, edate, from_stop, to_stop)
+    if df is None:
+        return {
+            'daily': [],
+            'overtime': []
+        }
+    daily = calc_travel_times_daily(df)
+    overtime = calc_travel_times_over_time(df)
```
mathcolo, line 105: I'm finding myself a bit confused by the "over time" nomenclature... could this be renamed to be more clear perhaps?

And to clarify, over time is the previous functionality and `daily` is the by hour?

server/app.py
```diff
@@ -111,6 +111,17 @@ def traveltime_aggregate_route():
     return json.dumps(response, indent=4, sort_keys=True, default=str)
 
 
+@app.route("/aggregate/traveltimes2", cors=cors_config)
```
mathcolo, line 114: Thank you for making this backwards compatible!

mathcolo (approved):
A few comments, but looks good to merge! Especially since it isn't into main anyway.

server/chalicelib/aggregation.py
austinjpaul: everyone was calling `reset_index` on the result of fast_describe, so I decided to move it inside the function.

src/AggregateCharts.js
austinjpaul: lol oops

src/App.css
austinjpaul: its... passable? Could use some improvement yet.
image
image

server/app.py
austinjpaul: 👍

server/chalicelib/aggregation.py
austinjpaul: Ah yes- I decided that and then stopped thinking about it. They're both grouping by time though, so I see the confusion too.
I think I'll change everything to "Aggregation by date" (existing) and "aggregation by time" (new). That reads sensibly and is hopefully quite descriptive of how it's being grouped.

src/App.js
```diff
@@ -201,7 +201,13 @@ class App extends React.Component {
     if (this.state.configuration.date_end) {
       options["start_date"] = this.state.configuration.date_start;
       options["end_date"] = this.state.configuration.date_end;
-      url = new URL(`${APP_DATA_BASE_PATH}/aggregate/${name}`, window.location.origin);
+
+      let method = name;
```
idreyn, line 205: More idiomatic to say 

```js
const method = name === "traveltimes" ? "traveltimes2" : name;
```

src/ChartSets.js
```diff
@@ -0,0 +1,113 @@
...
+        <SingleDayLine
+          title={"Travel times"}
+          data={props.traveltimes}
+          seriesName={"travel time"}
+          xField={'dep_dt'}
+          yField={'travel_time_sec'}
+          benchmarkField={anyTravelBenchmarks ? 'benchmark_travel_time_sec' : null}
+          location={locationDescription}
+          titleBothStops={true}
+          isLoading={props.isLoadingTraveltimes}
```
idreyn, line 83: Tiny nit but this looks like a typo to me because it feels like it should be `isLoadingTravelTimes` with the full camel-casing. Maybe that should be the prop name?

src/ChartSets.js
```diff
@@ -0,0 +1,113 @@
...
+          date={props.startDate}
+        />
+        <SingleDayLine
+          title={'Time between trains (headways)'}
+          data={props.headways}
+          seriesName={"headway"}
```
idreyn, line 89: I wonder if there's a kind of object that holds:

```
{ seriesName, xField, yField, benchmarkField }
```

that we could pass around instead of specifying these as four separate scalar values?

src/SelectableCharts.js
```diff
@@ -0,0 +1,135 @@
+import React from 'react';
+import { useState } from 'react';
+import classNames from 'classnames';
+import { AggregateByTime, AggregateByDate } from './line';
+
+const RadioForm = props => {
```
idreyn, line 6: Instinct tells me that `checked` should be a prop — if the component is providing its state to a parent via an `onChange` callback, the parent should _probably_ provide that state back as a prop — in short, the state should live in exactly one place. There are all sorts of reasons to bend this rule that I might not be picking up on, though.

src/SelectableCharts.js
```diff
@@ -0,0 +1,135 @@
...
+    super(props);
+    this.onChangeValue = this.onChangeValue.bind(this);
+
+    this.state = {
+      title: `${props.title} (Weekday)`,
+      selected: "weekday"
+    }
+  }
+
+  options = [
```
idreyn, line 49: This feels like it can be factored out of the component as module-level `const` declaration to signal to the reader that the instance is not going to try to mutate it.

src/SelectableCharts.js
```diff
@@ -0,0 +1,135 @@
...
+    super(props);
+    this.onChangeValue = this.onChangeValue.bind(this);
+
+    this.state = {
+      title: props.title,
+      selected: "all"
+    }
+  }
+
+  options = [
```
idreyn, line 99: (Same here)

idreyn (approved):
The Pandas stuff is greek to me but I took a look at the React side and things look really good! 👍

src/SelectableCharts.js
austinjpaul: absolutely right, i don't need to be storing the state in both places. Great catch!

PR #107, opened 2021-12-03
Peak and bus

austinjpaul:
Alrighty, here we go.

This is the main "bus logic" PR. There will be one more coming with configuration/toggle switch/ui changes, but this probably has most of the meat.

@mathcolo : I think you're in the best place to review the backend code. bus2train probably looks very different from when you last touched it, but it's pretty much based on the framework you laid, just in pandas.
@idreyn : Possibly still value in a review, even if pandas is not your jam
@crschmidt and @friendchris  : mostly FYI, but any feedback on readability, stop names, anything really is welcome.

This is what's up on dashboard-beta.transitmatters.org right now, if that helps go through the stop names.

- One of the main backend changes is the ability to pass a list of stop_ids as a urlParameter and the backend will appropriately fetch data for all of them. (e.g. different bus berths at the same station can have different stop_ids, despite being the same)
- New processing files: bus2train.py will generate the "event" type files for s3. I've been putting them in a separate "folder" in s3 (yes yes, it's just in a bucket, i know) in case we need to wipe them at some point.
- New processing file manifest.py will create a best guess for the bus's stations_json file to then be manually curated. 
- Stop names in the .json files are coming straight from GTFS's checkpoints.txt, but we can manually clean these up if we want (some are a little funky, albeit descriptive). 

Some work still to be done to make uploading a new month of data easier.


server/chalicelib/s3.py
```diff
@@ -13,7 +13,8 @@
 def download(key, encoding="utf8"):
     obj = s3.get_object(Bucket=BUCKET, Key=key)
     s3_data = obj["Body"].read()
-    decompressed = zlib.decompress(s3_data).decode(encoding)
+    # 32 should detect zlib vs gzip
+    decompressed = zlib.decompress(s3_data, zlib.MAX_WBITS | 32).decode(encoding)
```
mathcolo, line 17: Wait. Did you have to do this because I accidentally Zlib wrapped T alerts and called them .gzs even though they aren't?

src/AlertFilter.js
```diff
@@ -4,10 +4,21 @@ const known = [
...
+  /shuttle/,  // might want to worry about this one...
 ];
 
+const anti = [
+  / stop .* move /i // "The stop X will permanently move to Y"
+]
+
 const findMatch = (alert) => {
   const text = alert.text;
+  for (const exp of anti) {
```
mathcolo, line 16: I wonder if this whole thing can instead be ```
known.every(exp => text.match(exp)) && !anti.some(exp => text.match(exp))
```
or something?

Might be easier to read IDK

File: src/AlertFilter.js [hunk 3]
```diff
@@ -4,10 +4,21 @@ const known = [
   /[A-Za-z]+ speeds/,     // "Reduced speeds"
   /delay/,
   /notice/,
+  /shuttle/,  // might want to worry about this one...
```
mathcolo, line 7: Wonder if this should be [Ss]huttle; doesn't capture today's alert:

```
Shuttle buses replace Red Line service between JFK/UMass and Broadway on weekends through Dec 19.
```

Perhaps previous two also need the same treatment?

src/constants.js
```diff
@@ -1,8 +1,9 @@
 export const colorsForLine = {
-  Red: '#d13434',
-  Orange: '#e66f00',
-  Blue: '#0e3d8c',
-  Green: '#159765',
+	Red: '#d13434',
```
mathcolo, line 2: Are these tabs? I cannot tell

Also I know this show is kinda dumb but https://www.youtube.com/watch?v=SsoOG6ZeyUI lol

mathcolo (approved):
WOW I am so amped for this. Awesome work!! Few comments.

server/bus/manifest.py
```diff
@@ -0,0 +1,100 @@
...
+    # use checkpoint file to map time_point_id to stop_name
+    station_names = load_checkpoints(checkpoint_file)
+    tpts['stop_name'] = tpts['time_point_id'].map(station_names)
+
+    # Create full stop id e.g. '66-0-64000'
+    tpts['full_stop_id'] = tpts[['route_id', 'direction_id', 'stop_id']].astype(str).agg('-'.join, axis=1)
+
+    # Must be arranged by inbound direction. Use most common (mode) as guess (w/ max in case 2 modes)
+    orders = df.loc[df.direction_id == 1].groupby("time_point_id", dropna=False)['time_point_order'].agg(lambda x: x.mode().max())
+    tpts['order_guess'] = tpts['time_point_id'].map(orders).fillna(0).astype(int)
```
mathcolo, line 56: Can we get these from GTFS instead?

server/Pipfile
```diff
@@ -14,7 +14,7 @@ chalice = "*"
 pytz = "*"
 boto3 = "*"
 flake8 = "*"
-pandas = "*"
+pandas = ">=1.3"
```
mathcolo, line 17: Also needs to be updated in requirements.txt

server/bus/gen_manifests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+route="111"
+
+mkdir data/output/manifests/$route
+
+for f in $(find data/input/MBTA_Bus_Arrival_Departure_Times_2021/ -name *.csv); do
```
mathcolo, line 7: Wonder if route and year could be arguments to make this easier to use?

server/bus/bus2train.py
```diff
@@ -0,0 +1,125 @@
...
+    fname.parent.mkdir(parents=True, exist_ok=True)
+    # set mtime to 0 in gzip header for determinism (so we can re-gen old routes, and rsync to s3 will ignore)
+    events.to_csv(fname, index=False, compression={'method': 'gzip', 'mtime': 0} if not nozip else None)
+
+
+def to_disk(df, root, nozip=False):
+    """
+    For each service_date/stop_id group, we call the helper that will write it to disk.
+    """
+    df.groupby(['service_date', 'stop_id', 'direction_id', 'route_id']).apply(lambda e: _write_file(e, root, nozip))
```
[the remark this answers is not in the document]
austinjpaul: haha that could be a good or a bad thing. I guess similarly to how I feel about this use of apply.
A for loop might be easier to read actually, now that you mention it.

server/bus/bus2train.py
```diff
@@ -0,0 +1,125 @@
...
+
+def process_events(df):
+    """
+    Take the tidied input data and rearrange the columns to match rapidtransit format.
+    - Rename columns (trip_id, stop_sequence, event_time)
+    - Remove extra columns
+    - Add empty vehicle columns
+    - Calculate event_type column with ARR and DEP entries
+    """
+    CSV_HEADER = ["service_date", "route_id", "trip_id", "direction_id", "stop_id",
```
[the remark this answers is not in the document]
austinjpaul: bah, i'm really bad at staying consistent apparently.

server/chalicelib/data_funcs.py
```diff
@@ -13,17 +13,22 @@ def stamp_to_dt(stamp):
     return dt.strftime(DATE_FORMAT)
 
 
-def use_S3(date):
-    return (date.today() - date).days >= 90
+def is_bus(stops):
+    return '-' in stops[0]
```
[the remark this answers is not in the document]
austinjpaul: I could make it check for two dashes? Or a regex of `"*-[01]-[0-9]+"` ?
Anything stricter would require changing the api i think.

server/chalicelib/s3.py
austinjpaul: If you look at `upload` in this file, it's using zlib compression on the alerts. But our events uploaded by Chris (and now me) are all gzip.
Both types of things get retrieved through `download`, but thankfully zlib can decompress either if you pass it this flag.

src/constants.js
austinjpaul: yeah, i fixed this in the next set of changes to me merged

server/bus/manifest.py
austinjpaul: not easily. I don't think. Since they're "timepoints" and not stops?
And we could try and map the stop ids, but there can be multiple...
I guess I was thinking this would be quick-and-dirty but good enough, since there will have to be manual oversight of this step anyway.

server/Pipfile
austinjpaul: ah, I can update there too.
I'm only using 1.3 features in bus2train, so nothing "online". i.e. chalice is probably fine.
dep issues are my biggest fear, so i really hope this doesn't need much attention.

server/bus/gen_manifests.sh
austinjpaul: a good plan- do i keep it a bash script then?

src/AlertFilter.js
austinjpaul: /shuttle/i is doing the trick. I'm planning to go through these in more detail... not that I want to though.

src/AlertFilter.js
austinjpaul: yeah, good call. I opted for not short-circuiting with && just so it's super easy to read.

src/bus_constants/111.json
```diff
@@ -0,0 +1,155 @@
...
+                    "1": [
+                        "111-1-2829"
+                    ],
+                    "0": [
+                        "111-0-2832"
+                    ]
+                }
+            },
+            {
+                "stop_name": "Haymarket",
```
crschmidt, line 138: We should remove this stop from the list. It only gets 3 daily trips, which is just slightly over the trigger for displaying, but it looks silly.

src/bus_constants/28.json
```diff
@@ -0,0 +1,166 @@
...
+                    "1": [
+                        "28-1-64000"
+                    ],
+                    "0": [
+                        "28-0-64000"
+                    ]
+                }
+            },
+            {
+                "stop_name": "Roxbury Crossing Station (Bus)",
```
crschmidt, line 136:
```suggestion
                "stop_name": "Roxbury Crossing Station",
```

(Looks like you already fixed this in the code, but didn't rerun the manifest generation fully for this one.)

src/bus_constants/114-116-117.json
```diff
@@ -0,0 +1,203 @@
...
+                        "117-1-5720"
+                    ],
+                    "0": [
+                        "116-0-5755",
+                        "117-0-5755"
+                    ]
+                }
+            },
+            {
+                "stop_name": "Market Basket",
```
crschmidt, line 100: This only has ~5 daily trips and only on the 114; not clear it adds a lot of value, I'd probably remove it.

server/bus/gen_manifests.sh
austinjpaul: I bet @crschmidt  has a plan to update the original scripts, so not gonna worry for now. (can perhaps remove these entirely even)

src/bus_constants/114-116-117.json
austinjpaul: I'm seeing 8 or 9 daily trips, but only on weekdays. Wondering now if we should remove 114 from this set?

server/Pipfile
austinjpaul: Requirements.txt has been bumped, server didn't seem to break.

PR #112, opened 2021-12-08
Improve local development API path code.

crschmidt:
Previously, when loading the dashboard on a host other than localhost (e.g.
a development server name), the lookup would fail and return "undefined",
which would break the API requests in a confusing way. Rather than hardcoding
localhost, just fall back to the relative path option for any falsey thing.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,12 +12,13 @@ import Select from './Select';
 import { configPresets } from './constants';
 
 const FRONTEND_TO_BACKEND_MAP = new Map([
-  ["localhost", ""], // this becomes a relative path that is proxied through CRA:3000 to python on :5000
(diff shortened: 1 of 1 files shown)


austinjpaul:
I'm not overly familiar with what's happening here. I recognize that there is no functional difference in code, but worry about losing the comment about ports, and the readability of "oh, localhost is accounted for thusly" . (An additional comment could address both of these concerns)

mathcolo:
Agreed with @austinjpaul re: a comment, but otherwise this is a good change.

nathan-weinberg:
@crschmidt Thanks for the change, do you think we could resolve conflicts/add that comment so we can merge this in?

PR #118, opened 2021-12-15
Adding new routes + improved names

crschmidt:
Adding additional routes into the bus data:

 -15, 22, 32, 39, 71, 73, 77 (popular according to recovery dash)
 - Shorten names by using postal abbreviations for street intersections
 - Add some prior-year stops to a couple of routes.


austinjpaul (commented):
A first pass. I might look a bit more at the stop ids before we merge.

Could you update Route 1 with "Mass Ave" etc. also?

Other abbreviation thoughts to take or leave as you see fit: 
Square -> Sq
Commonwealth Avenue -> Comm Ave
Center -> Ctr
Station -> Sta
Mount -> Mt

Also, do you mind addressing the 111 `thche|bwybc` issue? I think you're right that they can be two different stops labeled (Inbound) and (Outbound). We can't really autoselect for the user without much more complexity, but we _could_ put them in different branches so they can't select both simultaneously. Though that said, I'm happy to just rely on the user to not be silly.

src/bus_constants/111.json
```diff
@@ -133,6 +134,22 @@
...
+                "stops": {
+                    "1": [
+                        "111-1-8309",
+                        "111-1-8310"
+                    ],
+                    "0": [
+                        "111-0-8309",
+                        "111-0-8310"
+                    ]
+                }
```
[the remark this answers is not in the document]
crschmidt: Hmm, now that i think about it, this might actually be differently concerning: Haymarket isn't used in the current timepoints data, I don't think, so maybe this is actually not a new stop, just a new name? I only saw this turn up in older datasets. I'll take a second look to make sure.

src/bus_constants/32.json
```diff
@@ -0,0 +1,97 @@
...
+                    "1": [
+                        "32-1-6474"
+                    ],
+                    "0": [
+                        "32-0-6500"
+                    ]
+                }
+            },
+            {
+                "stop_name": "Hyde Park Ave @ Cummins Highway",
```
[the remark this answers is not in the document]
crschmidt: I thought about it, but it feels uncomfortable to me; I know it's standard, but it's not common.

src/bus_constants/39.json
```diff
@@ -0,0 +1,150 @@
...
+                    "1": [
+                        "39-1-1128"
+                    ],
+                    "0": [
+                        "39-0-1939"
+                    ]
+                }
+            },
+            {
+                "stop_name": "Centre St & South Huntington Ave",
```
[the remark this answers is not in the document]
crschmidt: Yeah, will probably go to "S" and drop the last Ave entirely.

src/bus_constants/111.json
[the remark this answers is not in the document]
austinjpaul: The 111 always goes to Haymarket, so maybe there's missing data. 
The stop did just move due to construction, but it's accurately showing up for rte 117, so that shouldn't be the issue.

src/bus_constants/111.json
```diff
@@ -133,6 +134,22 @@
                         "111-0-2832"
                     ]
                 }
+            },
```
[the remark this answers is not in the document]
crschmidt: Done.

src/bus_constants/57.json
```diff
@@ -94,7 +94,7 @@
                 }
             },
             {
-                "stop_name": "Commonwealth Avenue @ BU Bridge",
+                "stop_name": "Commonwealth Ave @ BU Bridge",
```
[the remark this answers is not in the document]
crschmidt: Dropped this and the other to just "Comm Ave".

src/bus_constants/111.json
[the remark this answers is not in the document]
crschmidt: Okay, I've just checked, and running the manifest.py file against 111 definitely includes the hayms stop, so I also have no idea how you missed this one :)

crschmidt:
I took the Mount Auburn -> Mt Auburn change and Commonwealth Ave -> Comm Ave. I split the 111 IB/OB into two stops. I think that this addresses all the things mentioned here, and we can pull this in once we have data uploaded.

austinjpaul:
Awesome, I think that's a good compromise on abbreviations. I did adjust the mobile screen formatting so things should just shrink instead of getting cut off, so it's not crucial that everything be super short - just preferable.

src/bus_constants/111.json
```diff
@@ -78,15 +79,22 @@
...
                         "111-1-5607"
-                    ],
+                    ]
+                }
+            },
+            {
+                "stop_name": "Broadway & Beacon (Outbound)",
+                "branches": null,
+                "station": "bwybc",
+                "order": 6,
```
austinjpaul, line 96: It may seem silly but `bwybc` and `thche` should have different stop orders to prevent undefined sorting behavior. 
If the outbound stop gets order 6.5 I think that should work- it will even show headways when the pair is selected, unless i've done that logic backwards (but no traveltimes ofcourse).
Oh, and I think we need `"stops": { "0": [] ...}` to stay in, or we might get some errors.

crschmidt: Done and done!

src/bus_constants/111.json
```diff
@@ -86,18 +86,20 @@
                 "stops": {
                     "1": [
                         "111-1-5607"
-                    ]
+                    ],
+		    "0": []
```
austinjpaul, line 90: eek we got some tabs

crschmidt: fixed

src/bus_constants/15.json
```diff
@@ -0,0 +1,139 @@
...
+                "order": 1,
+                "stops": {
+                    "1": [
+                        "15-1-15100"
+                    ],
+                    "0": [
+                        "15-0-15100"
+                    ]
+                }
+            },
```
austinjpaul, line 37: These shouldn't both have order 1. 
The route goes `fldcr -> genbo -> peter -> kane -> ...` but there's a couple variants.
Kane buses start at `kane`.
Peter's Sq buses go `peter -> kane`.
Fields Corner buses go `fldcr -> genbo -> kane`. (they presumably stop at `peter`, but don't use it as a timepoint)
Given that only the less common Peter's Sq trips use it as a timepoint, I propose we simplify and just remove `peter` from the listing. (it's not too far from kane anyhow)

src/stations.js
```diff
@@ -1,19 +1,33 @@
 import { stations as rt_stations } from './constants';
 
 import bus_1 from './bus_constants/1.json';
+import bus_15 from './bus_constants/15.json';
+import bus_22 from './bus_constants/22.json';
 import bus_23 from './bus_constants/23.json';
+import bus_32 from './bus_constants/32.json';
```
austinjpaul, line 7: 32 is out of order

src/bus_constants/15.json
crschmidt: Removed Peter.

src/stations.js
crschmidt: reordered import.

src/bus_constants/39.json
```diff
@@ -0,0 +1,150 @@
...
+                    "1": [
+                        "39-1-11389"
+                    ],
+                    "0": [
+                        "39-0-11388"
+                    ]
+                }
+            },
+            {
+                "stop_name": "Copley",
```
austinjpaul, line 123:
```suggestion
                "stop_name": "Copley (Inbound)",
```
Since only inbound buses stop here.
2 other points:
We're losing legacy data from bdal by not including it, and hunbv won't have inbound data before June 2020.
I think this is the correct treatment for now, but I'm going to open an issue in case we figure out a way to cleanly solve this later.
(from http://roster.transithistory.org/MBTARouteHistory.pdf : "In June 2020 the inbound route was changed to use Ring Road in place of Belvidere and Dalton Street between Huntington Ave. and Boylston St.")

src/bus_constants/111.json
austinjpaul: So... this still might throw errors it turns out. Will have to test once everything is merged together.

austinjpaul (approved):
Once `Copley (Inbound)` is fixed, it's good to merge.
I have lingering concerns about the 114/116/117 combo, but that's not your doing, so I'll address it later.

PR #130, opened 2021-12-17
Progress Bar refactor

austinjpaul:
I wanted this to be a "minor bug fix", and functionality-wise, it is. But it's structurally different enough that it probably warrants a separate review. Thanks, and sorry!

==========
ProgressBar now encapsulates its own "progress" state. This prevents everything in App from rerendering with the timer (since state was being updated). 
The `key` is used to tell React when to "refresh" the progress bar (i.e. on configuration change), which it does by mounting a brand new one. Visibility is controlled by presence or absence in the DOM instead of css.

### DIFF
--- a/src/ui/ProgressBar.js
+++ b/src/ui/ProgressBar.js
@@ -1,19 +1,81 @@
 import React from "react";
 import './ProgressBar.css';
 
-const ProgressBar = (props) => {
-  const { progress } = props;
-  if (!(progress >= 0 && progress < 100)) {
-    return (null);
+function progressBarRate(date_start, date_end) {
+  // Single day: no rate
+  if(!date_end) {
+    return null;
+  }
+  // Aggregation: fake rate based on how many days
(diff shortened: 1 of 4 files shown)


src/ui/ProgressBar.js
```diff
@@ -1,19 +1,81 @@
...
+  const total_seconds_expected = 3.0 * months;
+  return 100 / total_seconds_expected;
+}
+
+
+class ProgressBar extends React.Component{
+  /**
+   * This is now a completely uncontrolled component.
+   * Once mounted, it will run until 95% or it's unmounted.
+   * key is used by App.js to make React mount a new one each time.
```
mathcolo, line 22: Is this functionality that you get from React for free?

mathcolo (approved):
oh thank heavens

src/ui/ProgressBar.js
austinjpaul: yep! A nice sol'n to the very problem I had.

PR #132, opened 2022-01-13
Version bump tsunami (but primarily Python 3.9)

mathcolo:
⚠️ For after bus goes in!

- Python 3.9
- node 16
- Make sure Pipfile and requirements.txt match


package.json
```diff
@@ -31,7 +31,7 @@
     "lint-backend": "cd server && pipenv run flake8 --ignore=E501,E731 --exclude .chalice",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "postinstall": "cd server && pipenv install --python=3.7"
+    "postinstall": "cd server && pipenv install --dev"
```
mathcolo, line 34: @nathan-weinberg Unfortunately `--python=3.7` causes every `pipenv install` to delete and re-create the virtualenv, so I removed it

nathan-weinberg: Gotcha, that should be fine if it worked for you locally :+1:

nathan-weinberg (approved):
Assuming this was all tested LGTM, thanks for doing this @mathcolo!

mathcolo:
@nathan-weinberg nice ^ lol

mathcolo:
Closing because we're going to do python and js parts separately per @austinjpaul

PR #119, opened 2021-12-15
Better tooling for bus data uploads

crschmidt:
Creating a cleaner set of tools to run bus data generation start to finish, including manifest comparison, data downloads, and stop extraction.

PR #117, opened 2021-12-14
Bus Mode

austinjpaul:
Alright everyone, this is it: bus mode is here!
This branch is currently live on beta, so go play with it there and see if anything is amiss.
Use all the browsers, try to break it, tell me the bugs below.

@mathcolo @idreyn I'm getting some react warnings about switching an uncontrolled input to controlled about how i'm doing the Flatpickr when switching between rail and bus modes. But it finally has the behavior I think I want, and i'm not sure how worried i should be about this. (well, apparently mobile is still not working ,but it's close)

Issue #109, opened 2021-12-03
Peak chart options


austinjpaul:
The am-peak/pm-peak/off-peak options should be expanded to headways and dwells (server side), and incorporated into AggregateByDateSelectable charts (client side)
Also, the current suggestion is that these options are hidden unless you're in "advanced mode".
I'm thinking a normal-looking checkbox in the configuration for "advanced options"?

Issue #108, opened 2021-12-03
this.props.traveltimes can be obj or array


austinjpaul:
When switching from agg mode to single day mode, the array functions won't work on this.props.traveltimes since it's still an object.
Clicking on the (X) should clear the data in updateConfiguration.

devinmatte:
This is no longer an issue in v4, going to close now, since we won't fix this in v3 and instead just launch v4

PR #134, opened 2022-01-20
BUS MODE!

austinjpaul:
Alright folks, this is the one.
Changes here have been reviewed previously in #106 #107 #108 #109 #117 #118 #119 #130
Aim to merge Monday night for debugging so the site is live on Tuesday?


server/Pipfile
```diff
@@ -14,7 +14,7 @@ chalice = "*"
 pytz = "*"
 boto3 = "*"
 flake8 = "*"
-pandas = "*"
+pandas = ">=1.3"
```
mathcolo, line 17: I would put whatever is in `requirements.txt` in here so there are no surprises; just remember that `Pipfile` is used for pipenv/local dev but `requirements.txt` is what actually gets packaged and uploaded

mathcolo (approved):
Seen most of this before, lets do it 👍

server/Pipfile
austinjpaul: ah yes, thanks this got lost at some point. the only 1.3 required feature was only used for offline work, but good to have the same version on the server as well.
I'm starting to think the performance issues i was having with `pandas.describe` was a bug only in one version and has since been fixed, but i'm too scared to go back to using that builtin.

nathan-weinberg (approved):
This might be a TM Labs record for amount of code changed :1st_place_medal: 

Cursory glance at the code and a bit of playing with the beta site -> LGTM! Lmk if there are any CI enhancements you might want alongside this functionality :+1:

server/Pipfile
austinjpaul: requirements.txt is already 1.3.4 on my end?

mathcolo: I didn't check—if so, probably good then! 👍

PR #136, opened 2022-01-23
Switch from Google Analytics to GoatCounter

mathcolo:
The time has come.

NOTE: ⚠️ This is against `bus-staging`

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -117,7 +117,7 @@ class App extends React.Component {
       progressBarKey: 0,
     };
 
-    if(window.location.hostname === PRODUCTION) {
+    if (window.location.hostname === PRODUCTION) {
       window.goatcounter.count({
         path: props.location.pathname,
       });
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import ReactGA from 'react-ga';
 import { SingleDaySet, AggregateSet } from './ChartSets';
 import StationConfiguration from './StationConfiguration';
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -116,8 +117,11 @@ class App extends React.Component {
       progressBarKey: 0,
     };
 
-    ReactGA.initialize("UA-71173708-2");
-    ReactGA.pageview(props.location.pathname);
+    if(window.location.hostname === PRODUCTION) {
```
austinjpaul, line 120:
```suggestion
    if (window.location.hostname === PRODUCTION) {
```

src/App.js
```diff
@@ -328,7 +332,12 @@ class App extends React.Component {
           route: configuration.line,
         });
       }
-      ReactGA.pageview(window.location.pathname + window.location.search);
+
+      if (window.location.hostname === PRODUCTION) {
+        window.goatcounter.count({
+          path: window.location.pathname + window.location.search,
+        });
```
austinjpaul, line 339: trying to wrap my head around this API. It seems like pathname + location is the default? is there a way to let goatcounter get that info on its own so we don't have to specify?

austinjpaul (approved):
THANK YOU THANK YOU!
This looks awesome. Thrilled to ditch GA.
2 questions:
- Do we need to update package-lock bc package changed?
- have we tested this?

src/App.js
mathcolo: I don't think so; it requires a `path`: https://www.goatcounter.com/help/path

austinjpaul: Fair enough- this is very clear

PR #141, opened 2022-01-26
hotfix: force flatpickr for everyone

austinjpaul:
Fixing bug on iOS because apple decided html5 <input type=date>'s max: prop shouldn't do anything.
New version is currently deployed on beta if you have an iPhone and want to confirm.
(thanks @purcellif17 !!)


mathcolo (approved):
okayyyyy

PR #149, opened 2022-01-28
switch back to mobile-native dates, but with errors for iOS

austinjpaul:
@friendchris has already expressed his approval, but the code should still be looked at.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,6 +3,7 @@ import { SingleDaySet, AggregateSet } from './ChartSets';
 import StationConfiguration from './StationConfiguration';
 import { withRouter } from 'react-router-dom';
 import { lookup_station_by_id, get_stop_ids_for_stations, line_name } from './stations';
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -24,6 +25,8 @@ const RAPIDTRANSIT_PATH = "/rapidtransit";
 const BUS_PATH = "/bus";
 
 const MAX_AGGREGATION_MONTHS = 8;
+const TOO_EARLY_ERROR = (date) => `Our archives only go back so far. Please select a date no earlier than ${date}.`;
```
mathcolo, line 28: It's slightly weird that some of these are functions and some aren't, but that's ok

mathcolo (approved):

src/App.js
austinjpaul: I agree, but everything else I could think of was messy.

PR #152, opened 2022-01-30
Python 3.9 + drop pipenv in favor of poetry

mathcolo:
- Upgrade everything and switch to Python 3.9
- Drop pipenv in favor of poetry

Poetry's dependency resolution is night and day compared to pipenv. Plus a much nicer interface. It's awesome, seriously! Also— poetry can export a `requirements.txt` for chalice so we no longer have to update package versions in two places.


nathan-weinberg (approved):
LGTM

server/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "1.0.0"
+description = ""
+authors = ["TransitMatters Labs Team"]
+license = "MIT"
+
+[tool.poetry.dependencies]
+python = "^3.9"
+json-api-doc = "0.15.0"
+requests = "2.27.1"
+pytz = "2021.3"
```
austinjpaul, line 12: I believe py3.9 has standard libraries that make pytz obsolete. I'll open an issue to address this.

austinjpaul (approved):
looks good, and works like a charm. (fingers crossed nothing breaks)

PR #156, opened 2022-02-06
add todayDiclaimer for messy day-of data

austinjpaul:
Adds disclaimer for current day's data which may be messy.
It turn out that the T's overnight processing doesn't clean up all the mess, so we may want to consider other approaches in addition to this.
image

This also won't handle the midnight-3am scenario without more complicated logic... I can add if desired.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -509,3 +509,8 @@ button:disabled,
   font-style: italic;
   padding-top: 30px;
 }
+
+.today-disclaimer {
+  font-style: italic;
+  padding-top: 30px;
+}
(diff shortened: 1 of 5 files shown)


mathcolo (approved):

nathan-weinberg (approved):
Maybe we should center it in the page?

austinjpaul:
"Centering" is complicated: the chart-centered x-axis labels are not aligned with the page-centered footer, and i feel like centering the disclaimer too just brings attention to that. sorry it's not in my screengrab.

PR #157, opened 2022-02-06
Alerty

austinjpaul:
- add script that can be used to manually refill slowzones on s3 if/when the lambda fails (has been okay this month, but in dec there were many issues)
- differentiate between "no alerts" and "alerts not found"-- an important distinction!!
- allow alert fetching for 114/6/7

### DIFF
--- a/server/refill_slowzones.py
+++ b/server/refill_slowzones.py
@@ -1,0 +1,30 @@
+import datetime
+
+from chalicelib.parallel import date_range
+from chalicelib.s3_alerts import store_alerts, get_alerts
+
+
+START = datetime.date(2022,1,3)
+END = datetime.date.today() - datetime.timedelta(days=2)
+
+def do_alerts_exist(d):
+    try:
+        get_alerts(d, ["Red"])
+    except Exception as err:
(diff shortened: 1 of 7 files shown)


nathan-weinberg (approved):
LGTM

server/refill_slowzones.py
```diff
@@ -0,0 +1,31 @@
...
+def do_alerts_exist(d):
+    try:
+        get_alerts(d, ["Red"])
+    except Exception as err:
+        if err.response["Error"]["Code"] == "NoSuchKey":
+            return False
+    return True
+
+
+for d in date_range(START, END):
```
mathcolo, line 20: Should this have a sleep in it?

server/refill_slowzones.py
```diff
@@ -0,0 +1,31 @@
+import datetime
+
+from chalicelib.parallel import date_range
+from chalicelib.s3_alerts import store_alerts, get_alerts
+
+
+START = datetime.date(2022, 1, 3)
```
mathcolo, line 7: Should these be CLI args?

mathcolo (approved):
Few questions, but otherwise 👍

PR #159, opened 2022-02-18
Slow Zones

Smilerk:
#### This PR
- Adds slow zone information at `/slowzones`
- Has two graphs, one for total "slow" time per day, and one for all slow zones.
- There is also a dropdown bar on the top right that allows you to save the charts as `png/jpg`

#### Notes
- You'll have to add `all_slow.json` to `public/static/slowzones` locally to test (let me know if you have any troubles)
- I'd love some input on design / copy
- I could add a date filter if needed. Right now I'm starting at 1/1/2021
#### Gifs / Screenshots
Screen Shot 2022-02-18 at 10 23 13 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "highcharts": "^9.3.3",
     "highcharts-react-official": "^3.1.0",
     "lodash.merge": "^4.6.2",
+    "moment": "^2.29.1",
     "react": "^17.0.0",
     "react-chartjs-2": "^2.11.1",
     "react-dom": "^17.0.0",
--- a/package.json
+++ b/package.json
@@ -4,17 +4,21 @@
   "private": true,
   "dependencies": {
     "@tippyjs/react": "^4.0.2",
+    "@types/react": "^17.0.39",
     "chart.js": "^2.9.4",
     "classnames": "^2.2.6",
     "concurrently": "^7.0.0",
     "flatpickr": "^4.6.9",
+    "highcharts": "^9.3.3",
+    "highcharts-react-official": "^3.1.0",
     "lodash.merge": "^4.6.2",
     "react": "^17.0.0",
     "react-chartjs-2": "^2.11.1",
     "react-dom": "^17.0.0",
     "react-router-dom": "^5.1.2",
     "react-scripts": "^5.0.0",
-    "react-select": "^4.0.0"
+    "react-select": "^4.0.0",
+    "typescript": "^4.5.5"
   },
   "devDependencies": {
     "eslint": "^8.0.0"
--- a/src/App.js
+++ b/src/App.js
@@ -392,7 +392,7 @@ class App extends React.Component {
     return <div className="main-column">
       <div className="empty-state">
         {error_message && <>{error_message}</>}
-        <div id="slowzone-container"> Check our our new <Link to='/slowzones'><button id="slowzone-button">Slow Zone Tracker</button></Link></div>
+        <div id="slowzone-container"> Check out our new <Link to='/slowzones'><button id="slowzone-button">Slow Zone Tracker</button></Link></div>
         {!error_message && <>See MBTA rapid transit performance data, including travel times between stations, headways,
         and dwell times, for any given day. <span style={{fontWeight: "bold"}}>Select a line, station pair, and date above to get started.</span><div style={{marginTop: 10}}>Looking for something interesting? <span style={{fontWeight: "bold"}}>Try one of these dates:</span></div>
         <Select
--- a/src/App.js
+++ b/src/App.js
@@ -10,6 +10,7 @@ import { ProgressBar, progressBarRate } from './ui/ProgressBar';
 import './App.css';
 import Select from './inputs/Select';
 import { configPresets } from './presets';
+import { Link } from 'react-router-dom/cjs/react-router-dom.min';
 
 const PRODUCTION = "dashboard.transitmatters.org";
 
@@ -392,6 +393,7 @@ class App extends React.Component {
     return <div className="main-column">
       <div className="empty-state">
         {error_message && <>{error_message}</>}
+        <div id="slowzone-container"> Check our our new <Link to='/slowzones'><button id="slowzone-button">Slow Zone Tracker</button></Link></div>
         {!error_message && <>See MBTA rapid transit performance data, including travel times between stations, headways,
         and dwell times, for any given day. <span style={{fontWeight: "bold"}}>Select a line, station pair, and date above to get started.</span><div style={{marginTop: 10}}>Looking for something interesting? <span style={{fontWeight: "bold"}}>Try one of these dates:</span></div>
         <Select
--- a/src/App.js
+++ b/src/App.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { SingleDaySet, AggregateSet } from './ChartSets';
 import StationConfiguration from './StationConfiguration';
-import { withRouter, Link } from 'react-router-dom';
+import { withRouter } from 'react-router-dom';
 import { lookup_station_by_id, get_stop_ids_for_stations, line_name } from './stations';
 import { trainDateRange, busDateRange } from './constants';
 import { recognize } from './alerts/AlertFilter';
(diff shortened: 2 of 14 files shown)


austinjpaul:
A few comments, take all with much salt:
- We really need to make sure this stays legible on smaller screens, and I don't have any good ideas. Limiting how many months are shown to limit the number of bars could help. Also, setting a height minimum could be good so it doesn't collapse too much? I think having to scroll to see the bottom legend would be fine (esp with vertical gridlines [see note below], and maybe an x-axis on the top as well?)
- Speaking of range limits, have you tried highcharts built-in range selection for gantt? it looks pretty decent from afar, could be worth enabling.
- Personally, the multi-year data is cool to reflect on, but i think highlighting the most recent 3-4 months is maybe a better default.
- Also, I like vertical gridlines at the month boundaries- i think it helps mentally place these bars a little better.
- Are you targeting the navigation enhancements for a Mar 1 release? If not, i think we need to do something else in the meantime so options are more clear.

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+    },
+    categories: getRoutes(data),
+    reversed: true,
+  },
+  plotOptions: {
+    series: {
+      cursor: "pointer",
+      events: {
+        // @ts-ignore
+        click: function (event) {
```
idreyn, line 149: Might be preferable to write `event: any` instead of using a `@ts-ignore` here.

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+const getDirection = (to: any, from: any) => {
+  const toOrder = to.order;
+  const fromOrder = from.order;
+  return toOrder > fromOrder ? "southbound" : "northbound";
+};
+
+// Data formatting & cleanup
+export const formatSlowZones = (data: any) =>
+  data.map((x: any) => {
+    // @ts-ignore
```
idreyn, line 37: These two `@ts-ignores` seem like they're not suppressing any error.

idreyn (approved):
Giving this a tentative approval — I defer to Austin and Preston for a final thumbs-up, but the code looks great and TypeScript will be a very helpful addition as this project continues to grow! Most of my comments are about TS, actually.

Otherwise, very excited to see this get released soon!

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  let now: any = new Date(d.x2);
+  now.setDate(now.getDate() + 14); // two weeks after for comparison.
+  const today = new Date();
+  if (today < now) {
+    now = today;
+  }
+  now = now.toISOString().split("T")[0];
+
+  return `https://dashboard.transitmatters.org/rapidtransit?config=${d.custom.color},${d.custom.fr_id},${d.custom.to_id},${then},${now}`;
```
mathcolo, line 25: Since this is on the same domain, can this URL be relative? i.e. `return "/rapidtransit..."`

src/App.js
```diff
@@ -10,6 +10,7 @@ import { ProgressBar, progressBarRate } from './ui/ProgressBar';
 import './App.css';
 import Select from './inputs/Select';
 import { configPresets } from './presets';
+import { Link } from 'react-router-dom/cjs/react-router-dom.min';
```
mathcolo, line 13: Out of curiosity is there a reason this is pointing directly to the minified react-router-dom? I would have expected webpack to handle the loading of this or something, IDK.

package.json
```diff
@@ -4,17 +4,21 @@
   "private": true,
   "dependencies": {
     "@tippyjs/react": "^4.0.2",
+    "@types/react": "^17.0.39",
     "chart.js": "^2.9.4",
     "classnames": "^2.2.6",
     "concurrently": "^7.0.0",
     "flatpickr": "^4.6.9",
+    "highcharts": "^9.3.3",
```
mathcolo, line 12: Just another reminder to get this license in the shared drive

src/constants.ts
```diff
@@ -1166,4 +1184,4 @@ export const stations = {
 			}
 		]
 	}
-};
+};
```
mathcolo, line 1187: Nit: newline

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+
+export const groupByRoute = (data: SlowZone[]) =>
+  data.reduce((series: any, sz: SlowZone) => {
+    const key = sz.id;
+    const s = (series[key] || []).concat(sz);
+    series[key] = s;
+    return series;
+  }, {});
+
+export const groupByLine = (data: SlowZone[]) =>
```
mathcolo, line 66: Not sure if lodash `groupBy` would have helped make this cleaner? https://lodash.com/docs/4.17.15#groupBy

Probably not worth changing now but I just wanted to make sure you knew this existed lol

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+  return Object.keys(groupObj);
+};
+
+// Xrange options
+export const generateXrangeSeries = (data: any) => {
+  const routes = getRoutes(data);
+  const groupedByLine = groupByLine(data);
+  return Object.entries(groupedByLine).map((line: any) => {
+    return {
+      name: line[0],
```
mathcolo, line 85: Can `line` be de-structured here to avoid magic numbers?

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+      color: colorsForLine[line[0]],
+      data: line[1].map((d: SlowZone) => ({
+        x: Date.UTC(
```
mathcolo, line 88: Should this be Eastern timezone? Is the chart actually right?

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+  setDirection: Function
+): any => ({
+  chart: {
+    type: "xrange",
+  },
+  credits: {
+    enabled: false,
+  },
+  title: {
+    text: `${capitalize(direction)} Slow Zones`,
```
mathcolo, line 126: This is a huge nit but should the S and Z in slow zones be capitalized? I'm imagining

`Southbound slow zones` vs `Southbound Slow Zones` vs `Southbound Slow zones`

And I'm wondering if the first is actually the most correct.

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+  data: SlowZone[],
+  setChartView: Function
+): any => ({
+  title: {
+    text: `Slow Zones`,
+  },
+  xAxis: { type: "datetime", title: { text: "Date" } },
+  yAxis: {
+    title: {
+      text: "Slow Time Per Day (minutes)",
```
mathcolo, line 208: Same with the grammar here, perhaps `Time increase per day (min)`?

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+    },
+  },
+  series: groupByLineDailyTotals(data),
+  plotOptions: {
+    series: {
+      pointStart: Date.UTC(
+        2021,0,1
```
mathcolo, line 215: Another case of what do we do about UTC?

src/slowzones/formattingUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+      ),
+      pointInterval: DAY_MS,
+    },
+  },
+  exporting: {
+    buttons: {
+      customButton: {
+        text: "Detailed View",
```
mathcolo, line 223: `Detailed view`? IDK

mathcolo (commented):
A few comments; I don't think show stopping though? Mainly just wondering about the UTC vs EST cases

src/App.js
Smilerk: Looks like just a wierd import. I'll fix

package.json
Smilerk: done! its in the data dashboard folder

src/slowzones/formattingUtils.ts
Smilerk: Highcarts has a way to set a timezone! I'll push a change for it :)

mathcolo (approved):
woo let's do it!

PR #160, opened 2022-03-04
Bus2022

austinjpaul:
Jan and Feb added.


austinjpaul:
It sort of seems like MBTA is keeping 'dudly' as the checkpoint name for Nubain Station. 
(I mean, these OPMI files switched to nubn at some point, but then switched back)
I'm tempted to change our internal "station" id to match, just so we don't get as many warnings when loading new files. It won't be user-visible anywhere.

mathcolo (approved):
So wait, how is switching that internal name not a necessity for things to work correctly?

👍 otherwise

austinjpaul:
> So wait, how is switching that internal name not a necessity for things to work correctly?
The internal name is only used as a reference when generating the files, but isn't used by the front-end or server logic.
I'll hold off the switch for now, but will keep it in mind for future updates. (and hold out hope they'll switch to nubn)

PR #165, opened 2022-03-28
Update the Chris Friend new train run csvs automatically every morning

mathcolo:
This:

- Updates (at 10:05am UTC, every day) the csvs we already have with what happened the previous day
- Can be consumed easily by NTT
- Only supports pulling from MBTA-performance right now, not S3 events

Here's what those csvs look like, for reference. (simple!)
```
service_date,run_count
8/10/2019,16
8/11/2019,1
8/12/2019,2
8/13/2019,1
8/14/2019,5
8/15/2019,5
8/16/2019,6
[...]
```


nathan-weinberg (approved):
LGTM! Just needs lint errors fixed :D

austinjpaul (changes requested):
Ugh, I hate to do this since the code is already written, but in the interest of not confusing people in the future (myself included), I'd feel a lot better if we just put this in the NTT app and saved ourselves some technical debt to deal with later.

mathcolo:
@austinjpaul Point taken! I'll redo

PR #174, opened 2022-04-22
Update the new train run count csvs automatically every morning

mathcolo:
(Same as https://github.com/transitmatters/t-performance-dash/pull/165 but reopened)

We originally closed #165 because we thought the dashboard wasn't the right place for it to land. I convinced @austinjpaul it's probably OK for now, because a) it's not a big change, b) in the long term it'll probably get moved elsewhere, and c) it unblocks @thiteixeira.


server/app.py
```diff
@@ -25,6 +32,26 @@ def store_two_days_ago_alerts(event):
...
+        yesterday = date.today() - timedelta(days=1)
+        for route, spec in new_trains.ROUTES.items():
+            print(f"Storing new train runs for {route}...")
+            try:
+                run_count = len(new_trains.train_runs(
+                    spec["is_new"],
+                    spec["core_stations"],
+                    yesterday
+                ))
+                new_trains.update_statistics_file(route, yesterday, run_count)
```
austinjpaul, line 49: Can this logic be wrapped in a function in `new_trains.py` so we don't have to go back and forth between files to read it? The handler just calls `update` or something, and the update fn can do the for loop on the object.

server/app.py
```diff
@@ -25,6 +32,26 @@ def store_two_days_ago_alerts(event):
...
+            except Exception:
+                print(f"Unable to store new train run count for route={route}", file=sys.stderr)
+                continue
```
austinjpaul, line 52: It's disappointingly common that the MBTA api doesn't return. Can we forward this alert to slack somehow? nobody is going to read our lambda logs and I don't want data to get lost.

server/chalicelib/new_trains.py
```diff
@@ -0,0 +1,41 @@
...
+
+KEY = "NewTrains/AggData/{}LineAgg.csv"
+
+
+def train_runs(label_filter, stations, date):
+    api_data = MbtaPerformanceAPI.get_api_data("events",
+                                               {"stop": stations},
+                                               date)
+    all_events = sum([stop["events"] for stop in api_data], [])
+    deps = list(filter(lambda event: event["event_type"] in EVENT_DEPARTURE, all_events))
```
austinjpaul, line 24: It is possible that the same train gets both a PRD and DEP. This should be rare, but maybe we want to do some filtering for distinct trip_id.

server/chalicelib/new_trains.py
```diff
@@ -0,0 +1,41 @@
...
+    "Orange": {
+        "is_new": lambda x: int(x) >= 1400 and int(x) <= 1551,
+        "core_stations": [70014, 70015]  # Back Bay
+    }
+}
+
+KEY = "NewTrains/AggData/{}LineAgg.csv"
+
+
+def train_runs(label_filter, stations, date):
```
austinjpaul, line 19: Can this function take a color and do the dictionary lookup for stations and vehicles itself? I feel like that might be more readable (though it's probably just a personal preference)

server/chalicelib/new_trains.py
```diff
@@ -0,0 +1,41 @@
...
+                                               {"stop": stations},
+                                               date)
+    all_events = sum([stop["events"] for stop in api_data], [])
+    deps = list(filter(lambda event: event["event_type"] in EVENT_DEPARTURE, all_events))
+    return list(filter(lambda event: label_filter(event["vehicle_label"]), deps))
+
+
+def update_statistics_file(route, date, count):
+    csv_row = "{formatted_date},{count}\n".format(
+        formatted_date=date.strftime("%m/%d/%Y"),
```
austinjpaul, line 30: ew. can we do something more iso-like?

server/chalicelib/new_trains.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    key = KEY.format(route)
+    try:
+        data = s3.download(key, compressed=False) + csv_row
+    except ClientError as ex:
+        if ex.response['Error']['Code'] != 'NoSuchKey':
+            raise
+        data = "service_date,run_count\n" + csv_row
+
+    s3.upload(key, data.encode(), False)
```
austinjpaul, line 41: `compression=False` for readability.
Do we care what happens in 3 years when this file gets to be 1000 lines?

server/chalicelib/new_trains.py
```diff
@@ -0,0 +1,41 @@
+from chalicelib import MbtaPerformanceAPI, s3
+from chalicelib.constants import EVENT_DEPARTURE
+from botocore.exceptions import ClientError
+
+ROUTES = {
+    "Red": {
+        "is_new": lambda x: int(x) >= 1900 and int(x) <= 2151,
```
austinjpaul, line 7: can these just be `range(1900, 2152)`, and we do a `filter(lambda: int(event[vehicle_label]) in dict[new_vehs]` for the filter?

server/chalicelib/new_trains.py
austinjpaul, line 24: don't strictly need to convert this filter to a list since it just gets filtered again (tiny performance enhancement)

austinjpaul (approved):
I'm nitpicking this to point out possible issues, but it's really fine. Address only as many of these as you care about, since some are kind of icky.

server/app.py
thiteixeira: How long does the MBTA store/provide data for? We can check if the previous day is missing data and try to fetch it. If that fails we could send to Slack

thiteixeira (approved):
Besides Austin comments, lgtm

friendchris:
One idea—read the CSV, pull the last date, and pull every date from the MBTA afterward

PR #179, opened 2022-04-24
Add Deep Linking for Slowzones

### LINKED ISSUE #177: Slowzones: deep linking/query params
URL query parameters are not saved on configuration change, making sharing less than optimal

Smilerk:
#### This Pr
- Adds deep linking for slow-zones
- Isn't the prettiest thing I've ever written, but it works
- A test URL https://dashboard-beta.transitmatters.org/slowzones?chartView=xrange&startDate=2021-04-24&selectedLines=Red%2COrange
- Closes #177

### DIFF
--- a/src/slowzones/SlowZones.tsx
+++ b/src/slowzones/SlowZones.tsx
@@ -1,5 +1,7 @@
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { ChartView, Direction, SlowZone } from "./types";
+// @ts-ignore
+import { useHistory, useLocation, useParams } from "react-router-dom";
 import Highcharts from "highcharts";
 import HighchartsReact from "highcharts-react-official";
 import xrange from "highcharts/modules/xrange";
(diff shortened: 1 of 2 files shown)


src/slowzones/SlowZones.tsx
```diff
@@ -29,19 +30,37 @@ export const optionsForSelect = () => {
   });
 };
 
+function useQuery() {
+  const { search } = useLocation();
+
+  return useMemo(() => new URLSearchParams(search), [search]);
```
mathcolo, line 36: It works! But, in reading up on `useMemo` I found this:
> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

Are you sure this is a good idea

Smilerk: `useMemo` is a commonly used `hook` in React to prevent reinitialization of objects/functions/etc on every render. Here this hook is used to create a new `URLSearchParams` object when the `search` part of `Location` is changed. 
Without it, the `URLSearchParams` object won't react properly to changes and the components may have an incorrect representation of the current search param state.

mathcolo: Alrighty! Sounds good 👍

mathcolo (approved):

PR #180, opened 2022-04-26
Switch to monthly event files

austinjpaul:
Per @crschmidt's suggestion, we can cut down on data loading overhead by switching to monthly data event files instead of daily.

- Increases displayable range from 8 months of data to 18 months. (30s being a hard cap on API timeout)
- Incorporates @friendchris's data ingestion [scripts](https://github.com/transitmatters/performancedashbackfill) into the repo (adjusted for monthly files and to match the bus upload scripts)
- Updates the bus data upload scripts to output monthly files.

I would encourage everyone to check this out on beta to make sure the new progress bar timing is accurate.

Heads up: While the lambda now serves data 50% faster, local development server is pretty slow, at least on my laptop. My guess is too much concurrency- a single api request is quick, but loading a page (3 simultaneous requests with 5 threads each) creates a bottleneck. Open to suggestions on how to improve here.

### DIFF
--- a/server/bus/bus2train.py
+++ b/server/bus/bus2train.py
@@ -61,7 +61,6 @@ def load_data(input_csv, routes):
     OFFSET = datetime(1900, 1, 1, 0, 0, 0)
     df.scheduled = df.service_date + (df.scheduled - OFFSET)
     df.actual = df.service_date + (df.actual - OFFSET)
-    df.service_date = df.service_date.dt.date
 
     df.direction_id = df.direction_id.map({"Outbound": 0, "Inbound": 1})
 
@@ -99,18 +98,18 @@ def to_disk(df, outdir, nozip=False):
     """
     For each service_date/stop_id/direction/route group, we write the events to disk.
     """
-    grouped = df.groupby(["service_date", "stop_id", "direction_id", "route_id"])
+    monthly_service_date = pd.Grouper(key="service_date", freq="1M")
+    grouped = df.groupby([monthly_service_date, "stop_id", "direction_id", "route_id"])
 
     for name, events in grouped:
         service_date, stop_id, direction_id, route_id = name
 
         fname = pathlib.Path(outdir,
                              "Events",
-                             "daily-bus-data",
+                             "monthly-bus-data",
                              f"{route_id}-{direction_id}-{stop_id}",
                              f"Year={service_date.year}",
                              f"Month={service_date.month}",
-                             f"Day={service_date.day}",
                              "events.csv.gz")
         fname.parent.mkdir(parents=True, exist_ok=True)
         # set mtime to 0 in gzip header for determinism (so we can re-gen old routes, and rsync to s3 will ignore)
--- a/server/chalicelib/aggregation.py
+++ b/server/chalicelib/aggregation.py
@@ -89,7 +89,7 @@ def calc_travel_times_by_date(df):
     summary_stats_peak = faster_describe(df.groupby(['service_date', 'peak'])['travel_time_sec'])
 
     # combine summary stats
-    summary_stats_final = summary_stats.append(summary_stats_peak)
+    summary_stats_final = pd.concat([summary_stats, summary_stats_peak])
 
     return summary_stats_final
 
@@ -145,7 +145,7 @@ def headways_over_time(sdate, edate, stops):
     summary_stats_peak = faster_describe(df.groupby(['service_date', 'peak'])['headway_time_sec'])
 
     # combine summary stats
-    summary_stats_final = summary_stats.append(summary_stats_peak)
+    summary_stats_final = pd.concat([summary_stats, summary_stats_peak])
 
     # filter peak status
     results = summary_stats_final.loc[summary_stats_final['peak'] == 'all']
@@ -177,7 +177,7 @@ def dwells_over_time(sdate, edate, stops):
     summary_stats_peak = faster_describe(df.groupby(['service_date', 'peak'])['dwell_time_sec'])
 
     # combine summary stats
-    summary_stats_final = summary_stats.append(summary_stats_peak)
+    summary_stats_final = pd.concat([summary_stats, summary_stats_peak])
 
     # filter peak status
     results = summary_stats_final.loc[summary_stats_final['peak'] == 'all']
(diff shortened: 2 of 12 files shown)


server/bus/bus2train.py
```diff
@@ -99,18 +98,18 @@ def to_disk(df, outdir, nozip=False):
     """
     For each service_date/stop_id/direction/route group, we write the events to disk.
     """
-    grouped = df.groupby(["service_date", "stop_id", "direction_id", "route_id"])
+    monthly_service_date = pd.Grouper(key="service_date", freq="1M")
```
mathcolo, line 101: `Grouper`?

Dang, pandas is nice

src/ui/ProgressBar.js
```diff
@@ -10,7 +10,7 @@ function progressBarRate(date_start, date_end) {
   const ms = (new Date(date_end) - new Date(date_start));
   const days = ms / (1000*60*60*24);
   const months = days / 30;
-  const total_seconds_expected = 3.0 * months;
+  const total_seconds_expected = 1.25 * months;
```
mathcolo, line 13: Rock on, my friend

server/chalicelib/s3.py
```diff
@@ -57,14 +59,13 @@ def download_one_event_file(date, stop_id):
...
+    return download_one_event_file(date, stop)
 
 
-# signature: (date_iterable, [stop_id])
-# Will download events for multiple stops, over a date range.
-download_event_range = parallel.make_parallel(download_single_day_events)
+def download_events(sdate, edate, stops):
+    datestops = itertools.product(parallel.month_range(sdate, edate), stops)
+    result = parallel_download_events(datestops)
+    return filter(lambda row: sdate.strftime("%Y-%m-%d") <= row['service_date'] <= edate.strftime("%Y-%m-%d"), result)
```
mathcolo, line 71: This is clean, nice

server/chalicelib/aggregation.py
```diff
@@ -89,7 +89,7 @@ def calc_travel_times_by_date(df):
     summary_stats_peak = faster_describe(df.groupby(['service_date', 'peak'])['travel_time_sec'])
 
     # combine summary stats
-    summary_stats_final = summary_stats.append(summary_stats_peak)
+    summary_stats_final = pd.concat([summary_stats, summary_stats_peak])
```
mathcolo, line 92: ?

server/chalicelib/parallel.py
```diff
@@ -22,7 +22,14 @@ def parallel_func(iterable, *args, **kwargs):
 
 
 def date_range(start, end):
```
mathcolo, line 24: I wonder if these could go in a helper module or something? idk

server/chalicelib/s3.py
```diff
@@ -1,12 +1,14 @@
...
 from botocore.exceptions import ClientError
 import csv
+import itertools
 import zlib
 
 from chalicelib import parallel
 
 BUCKET = "tm-mbta-performance"
-s3 = boto3.client('s3')
+s3 = boto3.client('s3', config=botocore.client.Config(max_pool_connections=15))
```
mathcolo, line 11: Is this higher or lower than the default?

server/chalicelib/s3.py
```diff
@@ -57,14 +59,13 @@ def download_one_event_file(date, stop_id):
...
 
-def download_single_day_events(date, stops):
-    """Will download events for single day, but can handle multiple stop_ids"""
-    result = []
-    for stop_id in stops:
-        result += download_one_event_file(date, stop_id)
-    return sorted(result, key=lambda row: row["event_time"])
+@parallel.make_parallel
+def parallel_download_events(datestop):
+    (date, stop) = datestop
```
mathcolo, line 64: I wish this line made more sense

mathcolo (commented):
Awesome work!! Some comments; still digesting.

server/chalicelib/aggregation.py
austinjpaul: Pandas `append` is marked as deprecated, they want us to use concat instead.

server/chalicelib/s3.py
austinjpaul: Default is 10. This only matters for local development- and perhaps something else needs to be refined.

server/chalicelib/s3.py
austinjpaul: `make_parallel` takes a `function(A, *args, **kwargs)` and returns a `function([A], *args, **kwargs)`. (It runs in parallel over the first parameter, i.e. `[A]`).
If I used `make_parallel(download_one_event_file)` directly, it would only multiplex over dates, but not stop ids, which is less than ideal if you request 2 months of bus data from a stop that has 5 ids. To run all 10 requests in parallel, I had to conglomerate (date, stop) into the first param of this function, so we can use make_parallel(parallel_download_event), which takes a list of datestops despite what it looks like, because decorators.

Does that help? or just confuse further...

mathcolo: Do you mean it takes a `function([A], *args, **kwargs)` instead of `function(A, *args, **kwargs)`?

src/ui/ProgressBar.js
mathcolo, line 13: In playing on beta I don't know if it's quite this fast, even after warmup. Maybe nudge?

mathcolo (approved):
A remaining comment regarding speed, but otherwise nice!

server/chalicelib/s3.py
austinjpaul: no- backwards. unless we're talking in circles.
`parallel_download_events` has signature `function(A)`, and `@make_parallel` turns it into `function([A])`

src/ui/ProgressBar.js
austinjpaul: :( You might be right. Did you notice it timing out at all? Do i have to lower the 18 month limit also? or just the progress bar speed

mathcolo: Haven't noticed timeouts. I think you might be good with just progress bar speed

austinjpaul: 1.33 seconds per month matches the 8 to 18 month ratio expansion. We can keep an eye on if it's reasonable or not once it's live.

PR #182, opened 2022-04-29
Sort Slowzones

Smilerk:
#### This Pr
- Sorts slow zones by `order` of station id per line
- Runs `prettier` on a few files

#### Before
image_from_ios

#### After
Screen Shot 2022-04-29 at 5 18 28 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "typescript": "^4.5.5"
   },
   "devDependencies": {
+    "@types/react-router-dom": "^5.3.3",
     "eslint": "^8.0.0"
   },
   "homepage": ".",
--- a/src/slowzones/formattingUtils.ts
+++ b/src/slowzones/formattingUtils.ts
@@ -13,7 +13,7 @@ const getDashUrl = (d: any) => {
   const dateDiff = moment(d.custom.startDate).diff(d.x2, "months");
   let then;
   if (dateDiff <= -7) {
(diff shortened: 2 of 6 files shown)


tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "downlevelIteration": true
```
Smilerk, line 18: needed to add this for `[... new Set()]` to work

src/slowzones/formattingUtils.ts
```diff
@@ -76,8 +77,12 @@ export const groupByLine = (data: SlowZone[]) =>
   }, {});
 
 export const getRoutes = (data: SlowZone[]) => {
```
Smilerk, line 79: I'm using this object as the "y axis". Each zone can figure out its `y:` by looking for the index of its `id` in this obj which has been pre-sorted

package.json
```diff
@@ -22,6 +22,7 @@
     "typescript": "^4.5.5"
   },
   "devDependencies": {
+    "@types/react-router-dom": "^5.3.3",
```
Smilerk, line 25: fixes build

mathcolo (approved):

PR #184, opened 2022-05-02
CSV download links

austinjpaul:
Addresses #9 

Apologies to @mathcolo who didn't want a library- There was extra messiness there that I'd rather just leave under the rug (this was complicated enough already)

image

These filenames are a bit obnoxious, but I can't think of many better ideas.

image

Oh! The icon is borrowed from iconoir (MIT license). Do I need to cite?

### DIFF
--- a/src/ChartSets.js
+++ b/src/ChartSets.js
@@ -1,4 +1,3 @@
-import React from 'react';
 import { AggregateByTimeSelectable } from './charts/SelectableCharts';
 import { SingleDayLine, AggregateByDate } from './charts/line';
 import { station_direction } from './stations';
--- a/src/ChartSets.js
+++ b/src/ChartSets.js
@@ -8,18 +8,21 @@ import { TODAY_SERVICE_DATE } from './constants';
 const dataFields = {
   traveltimes: {
     seriesName: "travel time",
+    fname: "traveltimes",
     xField: "dep_dt",
     yField: "travel_time_sec",
     benchmarkField: "benchmark_travel_time_sec"
   },
   headways: {
     seriesName: "headway",
+    fname: "headways",
     xField: "current_dep_dt",
     yField: "headway_time_sec",
     benchmarkField: "benchmark_headway_time_sec"
   },
   dwells: {
     seriesName: "dwell time",
+    fname: "dwells",
     xField: "arr_dt",
     yField: "dwell_time_sec",
     benchmarkField: null
@@ -52,6 +55,7 @@ const AggregateSet = (props) => {
         title={"Travel times"}
         data={props.traveltimes.by_date?.filter(x => x.peak === 'all') || []}
         seriesName={"Median travel time"}
+        fname={"traveltimes"}
         location={locationDescription}
         titleBothStops={true}
         isLoading={props.isLoadingTraveltimes}
@@ -62,6 +66,7 @@ const AggregateSet = (props) => {
         title={headwayTitle[props.bus_mode]}
         data={props.headways}
         seriesName={'Median headway'}
+        fname={"headways"}
         location={locationDescription}
         titleBothStops={false}
         isLoading={props.isLoadingHeadways}
@@ -73,6 +78,7 @@ const AggregateSet = (props) => {
           title={'Time spent at station (dwells)'}
           data={props.dwells}
           seriesName={'Median dwell time'}
+          fname={"dwells"}
           location={locationDescription}
           titleBothStops={false}
           isLoading={props.isLoadingDwells}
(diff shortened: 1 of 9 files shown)


src/charts/download.js
```diff
@@ -0,0 +1,50 @@
...
+  "southbound": "SB",
+  "eastbound": "EB",
+  "westbound": "WB"
+};
+
+function filename(datasetName, location, bothStops, startDate, endDate) {
+  // CharlesMGH-SB_dwells_20210315.csv
+  // CentralSquareCambridge-MelneaCassWashington_traveltimesByHour-weekday_20200101-20201231.csv
+  // BostonUniversityWest-EB_headways_20161226-20170328.csv
+  const fromStop = location.from.replace(/[^A-z]/g, "");
```
mathcolo, line 15: My brain is mush and I cannot figure out what this does

src/charts/download.js
```diff
@@ -0,0 +1,50 @@
...
+  /**
+  * props:
+  *  - data
+  *  - seriesName
+  *  - location
+  *  - bothStops
+  *  - date
+  */
+  return(
+    <div className="download-button" title="Download Data as CSV">
```
mathcolo, line 39: Recommend "Download Data" -> "Download data"

src/charts/download.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import { CSVLink } from 'react-csv';
+
+const directionAbbrs = {
+  "northbound": "NB",
```
mathcolo, line 5: I see the python in your soul

src/charts/download.js
```diff
@@ -0,0 +1,50 @@
...
+  // BostonUniversityWest-EB_headways_20161226-20170328.csv
+  const fromStop = location.from.replace(/[^A-z]/g, "");
+  const toStop = location.to.replace(/[^A-z]/g, "");
+  const dir = directionAbbrs[location.direction];
+  const where = `${fromStop}-${bothStops ? toStop : dir}`;
+
+  const what = datasetName;
+
+  const date1 = startDate.replaceAll("-", "");
+  const date2 = endDate ? `-${endDate.replaceAll("-", "")}` : "";
```
mathcolo, line 23: Is there any risk of endDate being undefined?

mathcolo (approved):

src/charts/download.js
austinjpaul: remove any non-letter characters from the stop name. Gets rid of slashes, parens, and spaces.

src/charts/download.js
austinjpaul: I agree, thanks

src/charts/download.js
austinjpaul: XD am i supposed to remove quotes here or something?

mathcolo: I mean TECHNICALLY you'd be using a `Map` ;)

src/charts/download.js
austinjpaul: yes, i think so. Does that count as falsey here?

mathcolo: Oh. yes! it's probably fine then.

src/charts/download.js
thiteixeira: You could use

```js
const directionAbbrs = {
  northbound: "NB",
  southbound: "SB",
  eastbound: "EB",
  westbound: "WB"
};
```

src/ChartSets.js
```diff
@@ -8,18 +8,21 @@ import { TODAY_SERVICE_DATE } from './constants';
 const dataFields = {
   traveltimes: {
     seriesName: "travel time",
+    fname: "traveltimes",
```
thiteixeira, line 11: Nit: With React 17 you don't need to import React at the top of the file to avoid:
`'React' is declared but its value is never read.`

src/OpenSource.js
```diff
@@ -1,134 +1,183 @@
 import React from 'react';
```
thiteixeira, line 1: Same here: this line can be removed

src/charts/download.js
```diff
@@ -0,0 +1,50 @@
...
+  *  - location
+  *  - bothStops
+  *  - date
+  */
+  return(
+    <div className="download-button" title="Download data as CSV">
+      <CSVLink
+        className="csv-link"
+        data={props.data}
+        filename={filename(props.datasetName, props.location, props.bothStops, props.startDate, props.endDate)}
```
thiteixeira, line 43: Wouldn't be cleaner to destruct these props?

```js
const DownloadButton = ({
  datasetName,
  data,
  location,
  bothStops,
  startDate,
  endDate,
}) => {
     ...
     <CSVLink
        className='csv-link'
        data={data}
        filename={filename(
          datasetName,
          location,
          bothStops,
          startDate,
          endDate
        )}
      ></CSVLink>

```

thiteixeira (approved):
LGTM

Sorry for the individual comments. I wasn't planning to review the whole PR at the beginning 🙂

src/charts/download.js
austinjpaul: good call. thanks for the feedback! :)

PR #185, opened 2022-05-03
add year to xLabel in aggregate charts

austinjpaul:
Adding years to label the X-axis of our aggregate chart (so that screenshots on twitter can get full context).
[anyone can review-- i'm just picking people arbitrarily]

image

### DIFF
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -31,6 +31,8 @@ const prettyDate = (dateString, with_dow) => {
                         options);
 }
 
+const dateYear = (dateString) => {return dateString.split("-")[0]}
+
 const departure_from_normal_string = (metric, benchmark) => {
   const ratio = metric / benchmark;
   if (!isFinite(ratio) || ratio <= 1.25) {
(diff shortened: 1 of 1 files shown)


src/charts/line.js
```diff
@@ -31,6 +31,8 @@ const prettyDate = (dateString, with_dow) => {
                         options);
 }
 
+const dateYear = (dateString) => {return dateString.split("-")[0]}
```
idreyn, line 34: May as well drop the brackets and the `return` keyword:

```ts
const dateYear = (dateString) => dateString.split("-")[0];
```

src/charts/line.js
```diff
@@ -317,7 +319,11 @@ const AggregateByDate = (props) => {
       xMin={new Date(`${props.startDate}T00:00:00`)}
       xMax={new Date(`${props.endDate}T00:00:00`)}
       fillColor={"rgba(191,200,214,0.5)"}
-      xLabel=""
+      xLabel={
+        dateYear(props.startDate) === dateYear(props.endDate) ?
```
idreyn, line 323: $0.02 — we might as well store these in variables instead of calling `dateYear` on the same value several times.

idreyn (approved):
LGTM, this will be really helpful!

src/charts/line.js
austinjpaul: that's what I wanted to do, but I can't figure out how to put anything more complicated than a simple expression in the `props={}` section. Am I being silly? Should I just pull the whole thing into a `function dateLabel(d1, d2)`?

PR #191, opened 2022-05-09
Generate a summary json for Thiago to consume + switch to json

mathcolo:
This PR:

- Switches the new train statistics files from CSV to JSON, which is nicer to work with and easier to consume
- Writes a summary json out to the public bucket so that Thiago can consume it, which will look like:

```
{
  "Red": {
    "new_vehicles": {
      "max": 0,
      "on_date": null
    }
  },
  "Orange": {
    "new_vehicles": {
      "max": 115,
      "on_date": "2022-05-07"
    }
  }
}
```

⚠️ Open question: Do we want to hardcode things like vehicle delivery counts in here as well?

### DIFF
--- a/server/.chalice/policy.json
+++ b/server/.chalice/policy.json
@@ -35,7 +35,8 @@
       "Effect": "Allow",
       "Resource": [
         "arn:aws:s3:::tm-mbta-performance/Alerts/*",
-        "arn:aws:s3:::tm-mbta-performance/NewTrains/*"
+        "arn:aws:s3:::tm-mbta-performance/statistics/*",
+        "arn:aws:s3:::dashboard.transitmatters.org/static/statistics/*"
       ]
     }
   ]
--- a/server/chalicelib/new_trains.py
+++ b/server/chalicelib/new_trains.py
@@ -1,4 +1,6 @@
+import json
 import sys
+import traceback
 from chalicelib import MbtaPerformanceAPI, s3
 from chalicelib.constants import EVENT_DEPARTURE
 from botocore.exceptions import ClientError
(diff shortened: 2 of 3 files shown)


austinjpaul:
Eek -- not a fan. 
Firstly, writing to any static files served by cloudfront will require a cache invalidation if you want to be served the latest data every time.
But beyond that, I have issues with the dashboard-hitchhiker growing in scope and writing a frontend file to be consumed by... the traintracker frontend (with CORS?)?
@idreyn perhaps you have thoughts on whether this data should be part of traintracker's initial payload or fetched by the client.

If this process is to live in the dashboard, I'd much rather it stay as a datamunging csv generator, and have the traintracker server read csv from s3. [[Well, json is fine. I just don't want to get dashboard's cloudfront involved in NTT world]]

idreyn:
> writing a frontend file to be consumed by... the traintracker frontend (with CORS?)?

For a lot of reasons it would make the most sense for this file to be periodically downloaded by the NTT backend and then served to its frontend via its own API. I certainly like the idea of automatically generating this info, but I'm curious to see what this looks like on the NTT side — how do we discover the URL of this resource?

mathcolo:
@austinjpaul @idreyn I'm not in love with this either, but I'm also not convinced it's that bad in the short term! Some points...

- slow zones (despite being in a separate repository) also publishes files to dashboard.transitmatters.org/static, so this change fits the existing pattern
- In a future where we've unified data ingestion, it doesn't seem crazy to me to continue to have some precomputed data available in a bucket. If you squint and see "data.transitmatters.org" instead of "dashboard.transitmatters.org", this starts to make more sense. The code in this patch can be moved to another location without much effort
- I don't think NTT periodically caching data from S3 makes sense; what problem is that solving? It would seem to me that we want NTT to do less over time, not more

nathan-weinberg:
Alt Text

PR #193, opened 2022-05-10
Add goatcounter to slow zones

mathcolo:
- Add goatcounter to slow zones page
- Factor goatcounter and some constants out

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -112,7 +112,6 @@ class App extends React.Component {
       progressBarKey: 0,
     };
 
-    goatcount(props.location.pathname);
 
     const url_config = new URLSearchParams(props.location.search).get("config");
     if (typeof url_config === "string") {
@@ -143,6 +142,7 @@ class App extends React.Component {
 
   componentDidMount() {
     this.download();
+    goatcount();
   }
 
   checkForErrors(config) {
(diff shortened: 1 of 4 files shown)


src/analytics.ts
```diff
@@ -0,0 +1,13 @@
+import { PRODUCTION } from './constants';
+
+declare global {
+  interface Window { goatcounter: any; }
+}
+
+export function goatcount(path=window.location.pathname + window.location.search) {
```
austinjpaul, line 7: whoa you can do this? this is breaking my poor python head

mathcolo: yep! they're evaluated at call time

src/App.js
```diff
@@ -120,11 +112,7 @@ class App extends React.Component {
       progressBarKey: 0,
     };
 
-    if (window.location.hostname === PRODUCTION) {
-      window.goatcounter?.count?.({
-        path: props.location.pathname,
-      });
-    }
+    goatcount(props.location.pathname);
```
austinjpaul, line 115: can this use the default param also? since window.location.search should be empty? or no...

austinjpaul (approved):
LGTM, thanks!!

src/App.js
mathcolo: Yep! adjusted. that also let me to a double counting bug, fixed by moving the calls around a bit

PR #196, opened 2022-06-10
april 2022 bus update (no more cambrige common loop for 71, 73)

austinjpaul:


mathcolo:
Does removing those stop entries prevent us from viewing historical data involving them? Or maybe we don't care?

austinjpaul:
Correct.
I figure we're not losing much since we have the Harvard bus ally that's
like right there already, but I could be convinced otherwise.

mathcolo (approved):
Gotcha!

PR #201, opened 2022-07-31
Updating the Alerts Components to Typescript

devinmatte:
# Motivation

- Getting my feet wet with the repo by picking something that's low effort
- Switching to typescript produces safer code

# Changes

- Updated all files in the `src/Alerts` directory to use Typescript
- Added related types that were needed


src/alerts/AlertBar.tsx
```diff
@@ -20,19 +21,18 @@ function chartTimeframe(start_date) {
...
+          "width": `${width}%`,
           "height": "100%",
           "float": "left",
           "position": "absolute",
-          "left": `${props.left}%`,
-          "borderLeft": props.border && "0.1px solid white",
-          "borderRight": props.border && "0.1px solid white",
+          "left": `${left}%`,
+          "borderLeft": border ? "0.1px solid white" : undefined,
+          "borderRight": border ? "0.1px solid white": undefined,
```
thiteixeira, line 35: Probably can DRY this a bit with

```js
// Before the return and of course, sample name :)
toBorderOrNotToBorder = border ? "0.1px solid white" : undefined

"borderLeft": toBorderOrNotToBorder,
"borderRight": toBorderOrNotToBorder,
```

src/alerts/AlertBar.tsx
```diff
@@ -55,17 +59,17 @@ export default function AlertBar(props) {
     }
 
     const [start, end] = chartTimeframe(today);
-    const duration = (end - start);
+    const duration = (end.getTime() - start.getTime());
```
thiteixeira, line 62: I don't think we need this change. As far as I can tell, `[start, end]` are two epochs, so duration is as simple as subtracting start from end.

src/alerts/LegendAlerts.tsx
```diff
@@ -9,4 +9,6 @@ export default function legend_alerts() {
       </p>
     </div>
   );
-}
\ No newline at end of file
+}
+
+export default legend_alerts;
```
thiteixeira, line 14: Nit: duplicate export.

Another thing topic of discussion is to retire default exports in favor or named exports. Here's a couple of articles:

https://www.bundleapps.io/blog/use-named-exports-over-default-exports-in-javascript

https://betterprogramming.pub/understanding-the-difference-between-named-and-default-exports-in-react-2d253ca9fc22

thiteixeira (commented):
Added a few comments, but great job. I didn't run the code so not approving at the moment

src/alerts/AlertBar.tsx
devinmatte: We'll have to mute a ts warning then
```
The right-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.ts(2363)
```

src/alerts/LegendAlerts.tsx
devinmatte: I'm in agreement of retiring default exports, but didn't want to impose my opinion on the repo. I'll happily replace usage of default exports

austinjpaul: Let's do it. I also find default exports confusing.

src/alerts/AlertFilter.ts
```diff
@@ -23,15 +25,12 @@ const findMatch = (alert) => {
   return known.some((exp) => text.match(exp))
 }
 
-const recognize = (alert) => {
+export const recognize = (alert: Alert) => {
```
austinjpaul, line 28: this looks like a partial clean-up, probably because it was half-refactored at another point also.
if `findMatch` just returns a boolean, then `recognize` is redundant? 
(Can you type annotate return types in tsx? I'm not very familiar yet)

src/alerts/AlertFilter.ts
```diff
@@ -23,15 +25,12 @@ const findMatch = (alert) => {
...
 
-const alertText = (alert) => {
-  return findMatch(alert)[0];
-};
-
-export {
-  recognize,
-  alertText,
+export const alertText = (alert: Alert) => {
+  if (findMatch(alert)) {
```
austinjpaul, line 33: I'm confused why we need to call `findMatch` here. Maybe it was legacy?

src/alerts/LegendAlerts.tsx
devinmatte: As part of https://github.com/transitmatters/t-performance-dash/pull/202 we could enable https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-default-export.md going forward to transition fully away from default exports

src/alerts/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface Alert {
+  text: string;
+  valid_from: string;
+  valid_to: string;
```
austinjpaul, line 4: Now that we're being explicit about this (thank you!!), would things be better if valid_from and valid_to were dates instead of strings? Or would that open up a mess of browser timezone localization issues?

I'm mostly just curious, not necessarily recommending a change here.

src/alerts/AlertFilter.ts
devinmatte: So it appears that `alertText` is unused in the code lol. `recognize` is used but only once.

src/alerts/AlertFilter.ts
devinmatte: `recognize` is used in `App.js`. And yes we can do return typing (although it's generally inferred and most IDEs should understand the return type. However if we want to be explicit we can do something like 
```typescript
export const recognize = (alert: Alert): boolean => {
```

Overall yes, recognize is entirely redundant and, I can replace it's usage with `findMatch`

src/alerts/types.ts
devinmatte: We could totally use Dates here, strings are also fine as long as we define it. A `Date` is automatically inferred as a `Date.valueOf()` when used as a string, so it's just a matter of whether we need a Date object or not downstream.
Date object could introduce localization issues, but unlikely (and also unlikely to matter given I imagine 95% of users are within US-East)

src/alerts/AlertFilter.ts
devinmatte: Since it's unused, should we remove it? Or is there an expectation of this being used in the future?

austinjpaul: Let's remove it. `alert.text` works just fine.

src/alerts/types.ts
austinjpaul: It looks like using Dates would simplify all the `new Date(alert.from_time).getTime()` calls, but the issue is they come in as text by loading json from https://dashboard-api2.transitmatters.org/alerts/2022-01-12?route=Red (e.g.) and would need to be created somewhere. I wouldn't worry about it for now, strings are ok.

austinjpaul (approved):
Looks good, and still works. Thanks for these changes! I'll get to the other PRs soon hopefully

mathcolo (approved):
Nice!

PR #202, opened 2022-07-31
Updating Eslint config with plugins

devinmatte:
# Motivation

- Noticed that the codebase had a inconsistent code style between components
    - Also noticed that https://github.com/transitmatters/new-train-tracker had a more detailed eslint file with a prettier config

# Changes
- Grabbed the .eslintrc.js from `new-train-tracker` to provide a more consistent codebase between the projects.
- Installed related eslint plugins
- Added prettier but stopped enforcement in eslint for now
    - That should be turned on at a later time to avoid a massive PR
- Fixed the few eslint warnings/errors that came up from new plugins

## Notes

I know from maintaining Open Source projects in the past, that big config changes to a repo to fit someone else's preferences can be annoying. I tried to ensure that the config was as close to the existing config in https://github.com/transitmatters/new-train-tracker as to not be imposing my opinions on the repo. That being said, I still think these config changes will be useful as they'll provide a cleaner codebase and help newcomers be able to better follow the style guide.


.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
devinmatte, line 1: This is copied from https://github.com/transitmatters/new-train-tracker/blob/master/.prettierrc.js

.eslintrc.js
```diff
@@ -1,3 +1,31 @@
 module.exports = {
```
devinmatte, line 1: This is copied from https://github.com/transitmatters/new-train-tracker/blob/master/.eslintrc.js

src/App.js
```diff
@@ -232,7 +232,7 @@ class App extends React.Component {
       update.traveltimes = [];
       update.dwells = [];
     }
-    this.setState((state, props) => (
+    this.setState((state) => (
```
devinmatte, line 235: `props` was unused

src/App.test.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
 
+// TODO: This test seems unused, jest or rtl aren't on the install list
+// eslint-disable-next-line no-undef
 it('renders without crashing', () => {
```
devinmatte, line 7: Seems like we're not using `jest` or `react-testing-library` in this repo so this test may be a holdover from `react-scripts`?

src/charts/Legend.js
```diff
@@ -2,9 +2,9 @@ function Legend() {
     return (
       <div className="legend">
         <p><span className="legend-dot" style={{"backgroundColor": "#64b96a"}}></span> On time</p>
-        <p><span className="legend-dot" style={{"backgroundColor": "#f5ed00"}}></span> >25% longer than benchmark</p>
-        <p><span className="legend-dot" style={{"backgroundColor": "#c33149"}}></span> >50% longer than benchmark</p>
-        <p><span className="legend-dot" style={{"backgroundColor": "#bb5cc1"}}></span> >100% longer than benchmark</p>
+        <p><span className="legend-dot" style={{"backgroundColor": "#f5ed00"}}></span> {">25% longer than benchmark"}</p>
+        <p><span className="legend-dot" style={{"backgroundColor": "#c33149"}}></span> {">50% longer than benchmark"}</p>
+        <p><span className="legend-dot" style={{"backgroundColor": "#bb5cc1"}}></span> {">100% longer than benchmark"}</p>
```
devinmatte, line 7: `>` gets a parse error in eslint checks now. It needs to be escaped as a string

src/inputs/date.js
```diff
@@ -45,23 +45,23 @@ class Flatpickr extends React.Component {
     this.handleNodeChange = this.handleNodeChange.bind(this);
   }
 
-  createFlatpickr = () => {
+  createFlatpickr() {
```
devinmatte, line 48: `eslint-react` wants to ensure class component functions are defined in this form. If we switch to functional components we can use arrow functions

devinmatte (commented):
Here's my explanations for each of the changes made

.prettierrc.js
```diff
@@ -0,0 +1,6 @@
...
+    trailingComma: "es5",
+    tabWidth: 4,
```
thiteixeira, line 3: I find `tabWidth: 2` to be more concise and would advocate for that to be the standard.

src/App.js
thiteixeira: I would change this to a functional component since we are using `"react": "^17.0.0"` here

src/charts/Legend.js
thiteixeira, line 7: Can we move the styles for the top of the file and give them a meaningful name? 

For instance, 
```javascript
statusGreen = {"backgroundColor": "#64b96a"}
...
```

src/inputs/date.js
thiteixeira: Probably better to make it a separate PR, but it would be good to switch this components to functional and change "as they get touched" in the future

thiteixeira (commented):
Took a quick look at this. Code looks good to me, but I didn't run the code, sorry.

src/App.js
devinmatte: I tend to agree, but I think that may be better served in a separate PR where the eslint changes are not merged with it. I may start that PR tonight and link it here.

src/inputs/date.js
devinmatte: I think I should do this in a separate PR as to not tie it to the eslint changes. I will start a PR for that

src/charts/Legend.js
devinmatte: Done in e114d3b

src/inputs/date.js
devinmatte: https://github.com/transitmatters/t-performance-dash/pull/203

src/App.js
austinjpaul: Doesn't the function signature need to stay `(state, props) => {}` for react? Or does js magically make it okay?
something like `(state, _) => {}` would be ok, although honestly i like the `props` there as an indicator should someone decide to use it later. 
I assume this all changes fwen we switch to functional components, but in the meantime, maybe we just silence the lint error?

src/App.test.js
austinjpaul: that sounds right to me, unless @idreyn or @mathcolo had long-term plans.

src/inputs/date.js
```diff
@@ -45,23 +45,23 @@ class Flatpickr extends React.Component {
...
     this.flatpickr = flatpickr(this.node,
       {
         // if we get here, we've already decided on flatpickr. Don't resort to mobile-native.
         disableMobile: true,
         date: this.props.value,
-        onChange: (sel, dateStr, inst) => this.props.onChange(dateStr),
+        onChange: (sel, dateStr) => this.props.onChange(dateStr),
```
austinjpaul, line 54: same question here: is react okay with a different function signature for its callbacks?

devinmatte: Yep should be fine with it. A leading unused param must be present, either named or as `_`. But a trailing param can be ignored as it's unused. You can place it as a trailing `_`, but I believe ESLint will call that out as bad behavior as well

src/App.js
devinmatte: See https://github.com/transitmatters/t-performance-dash/pull/202/files#r935971050. But yes, Javascript is fine with this. Leading unused params still need something present, hence the use of `_` often. However trailing unused can simply remain omitted.

We can silence the lint error, but that is a code smell imo. If we'd prefer to keep trailing unused params we can set the eslint rule to warn or info only, but I'd personally recommend against turning it off altogether

devinmatte: I can temporarily ignore it for this line also, if we want it on for the repo but off for just this line for now (we'd probably want to come back to it later though)

.prettierrc.js
devinmatte: Updated in 0d3c34d

src/inputs/date.js
devinmatte: Screen Shot 2022-08-02 at 4 03 11 PM
Yep trailing `_` is still a code smell

devinmatte: This specific line issue is made potentially irrelevant by https://github.com/transitmatters/t-performance-dash/pull/203 also

src/App.js
austinjpaul: Gotcha-- I'm in the wrong here. Don't let my unfamiliarity with js and react get in the way of best practices!

austinjpaul (approved):
This is great! Thank you so much for helping out here. 
I agree having a more consistent style across repos will certainly help development, and would love to continue to hear your ideas on how to improve.

src/App.test.js
mathcolo: indeed 👍 no long term plans.

src/charts/Legend.js
```diff
@@ -1,10 +1,15 @@
+const onTimeStyle = {"backgroundColor": "#64b96a"};
```
mathcolo, line 1: This is a bit superfluous, could it be an object, a `Map`, or something else instead?

mathcolo (approved):

src/charts/Legend.js
devinmatte: The map ended up feeling more verbose, so I moved to a css file, since at it's core, that's all that was needed

src/App.test.js
devinmatte: Okay I'm going to delete the file then

Smilerk (approved):

devinmatte:
Okay fixed everything that's been commented now, will leave Pr alone unless there are more comments

PR #205, opened 2022-08-03
add scheduled_headway to bus data files

austinjpaul:
image

I have re-generated all the bus data files to include the scheduled_headway column, and have passed that data through to the front end when it exists.

What I have not yet done is look closely at the data for accuracy and anomalies. These changes are live on beta, so I'd encourage reviewers to look around there to make sure things are sensible.

### DIFF
--- a/server/bus/bus2train.py
+++ b/server/bus/bus2train.py
@@ -76,12 +76,13 @@ def process_events(df):
     - Calculate event_type column with ARR and DEP entries
     """
     CSV_HEADER = ["service_date", "route_id", "trip_id", "direction_id", "stop_id",
-                  "stop_sequence", "vehicle_id", "vehicle_label", "event_type", "event_time"]
+                  "stop_sequence", "vehicle_id", "vehicle_label", "event_type", "event_time",
+                  "scheduled_headway"]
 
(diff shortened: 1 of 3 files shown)


mathcolo:
Found a weird aberration between 5 and 6pm here....
https://dashboard-beta.transitmatters.org/bus?config=1,1-1-72,1-1-79,2022-06-08,

Not expecting that? https://d2o8eokdkim9o8.cloudfront.net/sites/default/files/route_pdfs/2022-spring/001_3.pdf

austinjpaul:
Oof there turns out to be a lot of this unfortunately. 
https://dashboard-beta.transitmatters.org/bus?config=1,1-1-72,1-1-79,2022-06-14,
These headways look made up to make the dots greener.
Unfortunately I'm not sure there's an easy fix here. We could average these given headways in half hour blocks? But if there's lots of errors, we might have to try and process GTFS data as we load in the bus data.

mathcolo:
Dang, that's not good :( Sorry to put a golden spike in this.

Since averaging schedule information gives us an inexact result for something codified by the T, we should probably wait until we process GTFS.

PR #142, opened 2022-01-26
Stop treating lint warnings as errors in deploy step

mathcolo:

PR #207, opened 2022-08-07
Eslint update round 2

devinmatte:
Okay, so eslint caused a build issue despite working fine with `npm start`

Two major things are different with this PR compared to the last one:

- The eslint parser now handles typescript explicitly
- There's a new CI job to ensure the code builds

By adding these two new parsers/plugins it threw a few more errors/warnings. Ignored some, fixed others:

- No explicit `any`, ignored for now
- No more `@ts-ignore`, if you need to have ts shut up, use `ts-expect-error` and describe why in the comment
- No use of `Function`. It offers no real typechecking

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 16
-      - name: Lint frontend code with ESLint
+      - name: Fully build frontend code
         run: |
           curl -sSL https://install.python-poetry.org | python3 -
           npm ci
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,23 @@
+name: build
+
+on: pull_request
+
+jobs:
+  frontend:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v3
+        with:
+          python-version: 3.9
+      - name: Set up Node 16.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - name: Lint frontend code with ESLint
+        run: |
+          curl -sSL https://install.python-poetry.org | python3 -
+          npm ci
+          npm run build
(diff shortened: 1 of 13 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
+name: build
```
mathcolo, line 1: Nice! do we need `lint-frontend` if this is here?

devinmatte: Yes, linting and build are different. What was happening before is linting wasn't catching errors build caught, but also build only cares about a successful build. Lint will catch good practices and code style we want to enforce that otherwise builds fine

mathcolo: I know they're different, but we got lint errors on an `npm run build`. So if react-scripts's build step is not also effectively doing what `lint-frontend` is doing, what is it doing? https://github.com/transitmatters/t-performance-dash/runs/7715645996?check_suite_focus=true#step:6:148

devinmatte: So it appears that yes `react-scripts` does run eslint as part of the build, and will error out on any errors it finds. It also appears that it has some default parser. When the linter seems to only care about what it's told about. What I do find odd is that it was linting typescript files in `lint-frontend` but it wasn't in `build`. 
The errors we saw there were of the linter not knowing how to interpret typescript files during build, hence why we were getting undefined errors for type definitions. All around the fact that it wasn't caught by the linter beforehand is odd to me.

I still think we should have both. As build will only care about errors, and lint will show us warns, and it'll show them to us in a more visible way (see example)
183305728-70defce3-c2d8-4b1b-85e8-c6744ee6b23d

nathan-weinberg: Some context: https://github.com/transitmatters/t-performance-dash/pull/142

@devinmatte I'm fine with a new job; do you think we could add a badge for it in the REAMDE between `lint` and `deploy`?

devinmatte: Badge added

austinjpaul (approved):
Not an expert in any of this, but it seems fine to me.

mathcolo (approved):

PR #209, opened 2022-08-12
Adding eslint warn for prettier and formatting typescript

devinmatte:
# Motivation

Enforce Eslint rules on files in the repo to get us to a consistent code style

# Changes

- Turn on `prettier` to warn
- Enforce prettier style on Typescript files (leaving alone Javascript for now, those can updated as touched)
- Adds eslint warn for default exports

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@typescript-eslint/parser": "^5.32.0",
     "eslint": "^8.21.0",
     "eslint-config-prettier": "^8.5.0",
+    "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-react": "^7.30.1",
     "eslint-plugin-react-hooks": "^4.6.0",
--- a/src/alerts/AlertBar.tsx
+++ b/src/alerts/AlertBar.tsx
@@ -5,7 +5,6 @@ import 'tippy.js/dist/tippy.css'; // optional
 import { LegendAlerts } from './LegendAlerts';
 import { Alert } from './types';
 
-
 function chartTimeframe(start_date: string) {
   // Set alert-bar interval to be 5:30am today to 1am tomorrow.
   const today = `${start_date}T00:00:00`;
@@ -15,7 +14,7 @@ function chartTimeframe(start_date: string) {
 
   const high = new Date(today);
   high.setDate(high.getDate() + 1);
-  high.setHours(1,0);
+  high.setHours(1, 0);
 
   return [low, high];
 }
@@ -27,26 +26,25 @@ interface BoxSectionProps {
   border: boolean;
 }
 
-const BoxSection: React.FC<BoxSectionProps> = ({title, width, left, border}) => {
-  const toBorderOrNotToBorder = border ? "0.1px solid white" : undefined
+const BoxSection: React.FC<BoxSectionProps> = ({ title, width, left, border }) => {
+  const toBorderOrNotToBorder = border ? '0.1px solid white' : undefined;
   return (
     <Tippy content={title}>
-      <div className="incident-section" style={
-        {
-          "width": `${width}%`,
-          "height": "100%",
-          "float": "left",
-          "position": "absolute",
-          "left": `${left}%`,
-          "borderLeft": toBorderOrNotToBorder,
-          "borderRight": toBorderOrNotToBorder,
-        }
-      }>
-      </div>
+      <div
+        className="incident-section"
+        style={{
+          width: `${width}%`,
+          height: '100%',
+          float: 'left',
+          position: 'absolute',
+          left: `${left}%`,
+          borderLeft: toBorderOrNotToBorder,
+          borderRight: toBorderOrNotToBorder,
+        }}
+      ></div>
     </Tippy>
-
   );
-}
+};
 
 interface AlertBarProps {
   alerts: Alert[];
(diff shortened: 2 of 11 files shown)


src/constants.ts
```diff
@@ -1,601 +1,652 @@
...
 };
 
 export const stations = {
-	"Red": {
-		"type": "rapidtransit",
-		"direction": {
-			"0": "northbound",
-			"1": "southbound"
-		},
-		"stations": [
```
mathcolo, line 54: I know this is a JavaScript file, but this being valid JSON right now is handy for manipulation in other programs like `jq`. Perhaps we can exempt this section?

mathcolo (approved):
Nice! 👍

package.json
```diff
@@ -44,7 +44,7 @@
     "get-sz-totals": "wget -N -x -nH -P public https://dashboard.transitmatters.org/static/slowzones/delay_totals.json",
     "get-szs": "wget -N -x -nH -P public https://dashboard.transitmatters.org/static/slowzones/all_slow.json",
     "lint": "npm run lint-frontend && npm run lint-backend",
-    "lint-frontend": "eslint src --max-warnings=0",
+    "lint-frontend": "eslint src",
```
austinjpaul, line 47: Curious about justification here- This used to cause lint to fail in our CD so that we couldn't pass github checks with a warning and then fail at build time, which was more strict. Have the recent changes changed this behavior?

src/alerts/AlertBar.tsx
```diff
@@ -55,39 +53,42 @@ interface AlertBarProps {
...
       const alert_end = Math.min(end.getTime(), new Date(alert.valid_to).getTime());
-      const width = (alert_end - alert_start) / duration * 100;
+      const width = ((alert_end - alert_start) / duration) * 100;
 
-      const left_pos = (alert_start - start.getTime()) / (duration) * 100;
-      const tooltip = `${new Date(alert.valid_from).toLocaleTimeString('en-US')} - ${new Date(alert.valid_to).toLocaleTimeString('en-US')}\n${alert.text}`;
+      const left_pos = ((alert_start - start.getTime()) / duration) * 100;
+      const tooltip = `${new Date(alert.valid_from).toLocaleTimeString('en-US')} - ${new Date(
+        alert.valid_to
+      ).toLocaleTimeString('en-US')}\n${alert.text}`;
```
austinjpaul, line 79: this seems like a funny place to break this line, but i don't have any better suggestions...

src/alerts/AlertBar.tsx
```diff
@@ -55,39 +53,42 @@ interface AlertBarProps {
...
 
-      boxes.push(<BoxSection title={tooltip} border={true} width={width} left={left_pos} key={idx} />);
+      boxes.push(
+        <BoxSection title={tooltip} border={true} width={width} left={left_pos} key={idx} />
```
austinjpaul, line 82: This might be more readable with the props vertically stacked?

src/slowzones/formattingUtils.ts
```diff
@@ -236,21 +219,21 @@ export const generateXrangeOptions = (
...
-      return `<div><span style="font-size: 10px">${moment(
-        this.point.custom.startDate
-      ).utc().format("MMMM Do YYYY")} - ${moment(this.point.x2).utc().format(
-        "MMMM Do YYYY"
-      )}</span><br/> <span style="color:${this.point.color}">●</span> ${
+      return `<div><span style="font-size: 10px">${moment(this.point.custom.startDate)
+        .utc()
+        .format('MMMM Do YYYY')} - ${moment(this.point.x2)
+        .utc()
+        .format('MMMM Do YYYY')}</span><br/> <span style="color:${this.point.color}">●</span> ${
```
austinjpaul, line 226: eek- does the `-` on the same line have lower precedence than the `.` on multiple lines? I don't like this.

austinjpaul (approved):
Thanks for this work! Just a few minor things caught my eye.

package.json
devinmatte: This isn't totally necessary to do. `build` still fails for eslint warnings so we would still have a single failed CI job if there are warnings, but, it wouldn't explicitly fail lint. I'm good with or without this change

src/alerts/AlertBar.tsx
devinmatte: Yeah prettier has its opinions on how to break lines. We can change this a bit by setting a longer maximum line length, but it'll try to break somewhere. This is a fairly long line it's trying to break so I also am not sure where it would be better to break

src/slowzones/formattingUtils.ts
devinmatte: I think I can better break this up in a prettier compliant way

src/alerts/AlertBar.tsx
devinmatte: Prettier reverts it back to this single line

src/slowzones/formattingUtils.ts
devinmatte: Prettier is fighting with me. This is a pretty complex line that probably could be made a little more readable, but not sure I want to tackle that right now

package.json
austinjpaul: Gotcha - I forgot build was a precheck now.

src/slowzones/formattingUtils.ts
austinjpaul: Didn't realize the machine had such strong opinions. I'm happy to defer, but also curious how liberal we want to be with exemptions

src/alerts/AlertBar.tsx
austinjpaul: is it okay with something like the following?
```
      const tooltip = `${new Date(alert.valid_from).toLocaleTimeString('en-US')}` +
        ` - ${new Date(alert.valid_to).toLocaleTimeString('en-US')}\n${alert.text}`;
```

src/constants.ts
devinmatte: Moved to an actual json file and imported, to allow easier usage

src/alerts/AlertBar.tsx
devinmatte: Yes, fixed in latest commit

src/slowzones/formattingUtils.ts
devinmatte: We can make a few config changes https://prettier.io/docs/en/configuration.html

austinjpaul: gotcha. I just realized this is all one big format string, so breaking it up with addition is a possibility here also?
```
`<div><span style="font-size: 10px">`
  + `${moment(this.point.custom.startDate).utc().format('MMMM Do YYYY')} - `
  + `${moment(this.point.x2).utc().format('MMMM Do YYYY')}
...
``` etc. etc.

src/alerts/AlertBar.tsx
austinjpaul: eh, fine with me

src/slowzones/formattingUtils.ts
devinmatte: Done

PR #212, opened 2022-08-13
Moving chart lines to TS

devinmatte:
Chart lines moving to TS.

Moved a bunch of the files in charts to TS. Some were broken out to provide a cleaner structure, others were more simply swapped out.


src/charts/types.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export interface Location {
+  to: string;
+  from: string;
+  direction: Direction;
+  line: string;
+}
+
+type Direction = 'northbound' | 'southbound' | 'eastbound' | 'westbound' | 'inbound' | 'outbound';
```
thiteixeira, line 36: We could try to use `Object.keys(directionAbbrs)` to avoid writing this in multiple places

thiteixeira (approved):
Looks good. A couple of suggestions
I haven't tested the code

src/charts/types.ts
devinmatte: The only other place I saw it also had 2 letter abbreviations as keys, and then these for the values, so not sure how much that would help. And we couldn't use `Object.keys(directionAbbrs)` to define a TS type

PR #220, opened 2022-09-08
make slowzones line graphs always include zero

### LINKED ISSUE #216: slowzones: fix line graph ymin at 0
The line graphs would be less misleading if the y-axis always bottomed out at 0 despite the data selected.

image

mxmeinhold:
shifting y axis minimums can distort how line graphs are compared

since these graphs are always positive (negative time spent in slow zones sounds really cool actually), we can simply set the minimum y axis to 0, and graphs should be much more visually consistent (at least until we get negative slowzones)

This should resolve #216

Screenshot:
transitmatters labs time spent in slow zones graph. the y axis begins at zero, even though the data does not include zero

### DIFF
--- a/src/slowzones/formattingUtils.ts
+++ b/src/slowzones/formattingUtils.ts
@@ -332,6 +332,7 @@ export const generateLineOptions = (
         fontSize: textSize,
       },
     },
+    min: 0,
   },
   series: groupByLineDailyTotals(data, selectedLines),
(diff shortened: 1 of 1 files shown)


austinjpaul (approved):
thanks for addressing this!

mxmeinhold:
@austinjpaul Doesn't look like I do

PR #226, opened 2022-10-04
Slowzone frontend improvements

austinjpaul:
image
 ### Added:
- Vertical line on yesterday's date to help indicate which SZs are current vs fixed recently
- Caution icons to SZs created by the recent shutdown (by closing the range on the date comparison)
- Construction icons to the "shutdown six" slowzones that should have been fixed.

### Also:
- Fixed line-chart bug where data points were not aligned with tick marks due to timezones.
- Fixed gantt bug where min and max didn't actually limit the chart range.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -22,7 +22,7 @@ export const colorsForLine: Record<string, string> = {
 };
 
 export const getDateThreeMonthsAgo = () => {
-  return moment().subtract(3, 'months').startOf('day');
+  return moment.utc().subtract(3, 'months').startOf('day');
 };
 
 export const TODAY_SERVICE_DATE = () => {
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -46,36 +46,32 @@ export const busDateRange = {
 
 export const stations = stations_json;
 
-export const majorEvents: { Red: MbtaMajorEvent[]; Orange: MbtaMajorEvent[] } = {
-  Red: [
-    {
-      start: '2019-06-11T00:00:00Z',
-      end: '2019-09-21T00:00:00Z',
-      color: 'Red',
-      title: 'Red line derailment',
-      description:
-        'A red line train derailment at JFK/UMass destroyed important signal infrastructure that took months to repair.',
-      type: 'derailment',
-    },
-  ],
-  Orange: [
-    {
-      start: '2021-03-16T00:00:00Z',
-      end: '2021-04-16T00:00:00Z',
-      color: 'Orange',
-      title: 'Orange line derailment',
-      description:
-        'An orange line train derailment damaged a switch, requiring track replacement. A speed restriction followed in order to give the new track time to settle.',
-      type: 'derailment',
-    },
-    {
-      start: '2022-08-19T00:00:00Z',
-      end: '2022-09-19T00:00:00Z',
-      color: 'Orange',
-      title: 'Orange line shutdown',
-      description:
-        'The orange line was shut down for 30 days to complete deferred maintenance',
-      type: 'shutdown',
-    },
-  ],
+export const majorEvents: Record<string, MbtaMajorEvent> = {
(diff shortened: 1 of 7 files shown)


src/slowzones/SlowZones.tsx
```diff
@@ -163,7 +163,8 @@ export const SlowZones = () => {
       )}
       {chartView === 'xrange' && (
         <div className="event-footer">
-          <span className="event-footer-text">⚠️ = Affected by a derailment or shutdown</span>
+          <span className="event-footer-text">⚠️ = Caused by a derailment or shutdown</span>
```
devinmatte, line 166: I'm personally more a fan of `Affected` over `Caused` because that's far less definitive

austinjpaul:
Latest iteration: 
image

src/constants.ts
```diff
@@ -46,36 +46,32 @@ export const busDateRange = {
...
-    },
-  ],
+export const majorEvents: Record<string, MbtaMajorEvent> = {
+  RedDerailment: {
+    start: '2019-06-11T00:00:00Z',
+    end: '2019-09-21T00:00:00Z',
+    color: 'Red',
+    title: 'Red line derailment',
+    description:
+      'A red line train derailment at JFK/UMass destroyed important signal infrastructure that took months to repair.',
```
idreyn, line 56: $0.02: I think we should capitalize the names of transit lines (`Red Line`)

src/slowzones/formattingUtils.ts
```diff
@@ -9,21 +9,36 @@ const textSize = isMobile ? '11' : 14;
...
+const getFootnoteIcon = (start: Moment, end: Moment, color: string) => {
+  if (color === 'Blue') return '';
 
   if (color === 'Red') {
-    return (
-      majorEvents.Red.some(event => start.isBetween(moment(event.start), moment(event.end))) ||
-      majorEvents.Red.some(event => end.isBetween(moment(event.start), moment(event.end)))
-    );
+    const event = majorEvents.RedDerailment;
+    if (start.isBetween(moment(event.start), moment(event.end), undefined, '[]')) {
```
idreyn, line 23: I'm curious what these last two parameters are for.

src/slowzones/SlowZones.tsx
```diff
@@ -163,7 +163,9 @@ export const SlowZones = () => {
       )}
       {chartView === 'xrange' && (
         <div className="event-footer">
-          <span className="event-footer-text">⚠️ = Affected by a derailment or shutdown</span>
+          <span className="event-footer-text">{`${EMOJI.derailment} = Affected by a derailment`}</span>
```
idreyn, line 166: Maybe cleaner to express as:

```tsx
<span className="event-footer-text">{EMOJI.derailment} = Affected by a derailment</span>
```

without the extra layer of interpolation.

idreyn (approved):
Looks good to me from the perspective of functionality and good JS.

src/slowzones/formattingUtils.ts
devinmatte: `granularity` and `inclusivity` respectively. Not totally sure myself what difference `'[]'` makes to `inclusivity`

src/slowzones/SlowZones.tsx
devinmatte: I tend to prefer the string definition, as the formatting is more guaranteed

devinmatte:
Screen Shot 2022-10-07 at 5 31 34 PM

Updated with feedback from slack with Tim

src/slowzones/formattingUtils.ts
austinjpaul: Whether to include the end points as part of the interval. [ for closed, ( for open.
Undefined is just a place holder, so could be removed if keyword arguments are a thing? Idk

PR #227, opened 2022-10-10
Rankings comparison proof-of-concept

devinmatte:
Some python code that parses information from the MTA to allows us to see headways for each line and the system itself to compare against the mbta. Ideally we can get data like this for ~5 other systems to show where the MBTA ranks in several attributes (headways, ridership, etc.)


austinjpaul:
This is neat! Assuming we want to store archival MTA data in s3, `rankings.py` should probably go in the `chalicelib` folder of https://github.com/transitmatters/data-ingestion

PR #228, opened 2022-10-12
Donate button

idreyn:
As the Slow Zones tracker (and the Data Dashboard generally) receive an enormous influx of traffic, we should at least try to convert some of those visits into donations for TM. I've added a very quick link to our org-wide `/donate` page affixed to the top-right of all DD pages that looks like this:

image

It works well on desktop and on modern mobile viewports (>350px).

_Test plan:_
Visit any page and verify that the new button takes you to the `/donate` page, in a new tab.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -23,6 +23,7 @@
 
 <body>
   <section class="hero is-dark is-bold">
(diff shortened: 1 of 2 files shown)


friendchris (approved):
love it

PR #229, opened 2022-10-12
fix: Move the donate button above the hero logo in z-index

idreyn:
...whoops. The button was occluded by the header graphic on small screens.


friendchris (approved):
Love it x2

PR #208, opened 2022-08-11
Gtfs benchmarks

austinjpaul:
Add code to process GTFS data to create benchmark headways and traveltimes for bus routes.

Code is live on beta, and the benchmark data exists for all bus routes for June and July 2022.
Check it out there and make sure things look reasonable!
I still need to figure out how to disable it for 114-116-117 somehow, since those headways aren't for the combined route.

Issue #197, opened 2022-06-10
Bus headways issue


austinjpaul:
https://dashboard.transitmatters.org/bus?config=73,73-1-2064,73-1-12614,2022-03-31,

needs investigation

austinjpaul:
updated link because buses no longer go to cambridge common: https://dashboard.transitmatters.org/bus?config=73,73-1-2064,73-1-20761,2022-03-31,

PR #232, opened 2022-10-14
add sort to address #197

austinjpaul:
This commit is pending in #208 but with Park and Kenmore data in with #231 I'm pulling this ahead.


austinjpaul:
This can just wait for #208

PR #233, opened 2022-10-15
Migrating more files to Typescript

devinmatte:
We're really close to converting the whole app away from Javascript, this grabs some more easy wins


skaplan-dev (approved):

devinmatte:
Due to the incident yesterday, this PR will remain with no functionality changes or movement of files. Only prettier formatting and file changes from JS to TS as directly as possible. 
I will follow this up by doing a second rounds like this, until, all files are `.ts` files, whereafter, I will properly type them instead of using `any`.

src/App.js
```diff
@@ -352,47 +354,64 @@ class App extends React.Component {
...
-          className="date-selector"
-          defaultLabel="Choose a date..."
-        />
-        </>}
+    return (
+      <div className="main-column">
+        <div className="empty-state">
+          {error_message && <>{error_message}</>}
+          <div id="slowzone-container">
+            {' '}
```
austinjpaul, line 383: This {' '} business looks funny to me. I think I just don't understand what it does.

src/charts/Title.tsx
```diff
@@ -27,14 +28,13 @@ function font(size_px = 16) {
   return `bold ${size_px}px "Helvetica Neue", "Helvetica", "Arial", sans-serif`;
 }
 
-const calcLocationWidth = (words, ctx) => {
+const calcLocationWidth = (words: string[][], ctx: CanvasRenderingContext2D) => {
```
austinjpaul, line 31: Below, chart is type any. Can ctx be any too? Just in case chartsjs changes how they do things? Maybe this doesn't make any sense.

austinjpaul: Also, I know I wrote this but now I'm finding string[][] strange. Maybe we can have a subtype that is textPiece that is string[2]? I want a tuple where the first elem is a string to be printed and the second elem is a color. But maybe that's tricky to do. Probably this is fine and it just wants more comments.

src/ui/ProgressBar.js
```diff
@@ -7,6 +7,7 @@ function progressBarRate(date_start, date_end) {
     return null;
   }
   // Aggregation: fake rate based on how many days
+  // @ts-expect-error Typescript doesn't like subtracting a Date
```
austinjpaul, line 10: Good to know. We can fix this "later"

src/index.js
```diff
@@ -2,14 +2,14 @@ import React from 'react';
...
-import OpenSource from './OpenSource';
+import { OpenSource } from './OpenSource';
 import { BrowserRouter, Route, Switch } from 'react-router-dom';
 import { SlowZones } from './slowzones/SlowZones';
 
 ReactDOM.render(
   <BrowserRouter>
     <Switch>
-      <Route exact path="/opensource" >
+      <Route exact path="/opensource">
```
austinjpaul, line 12: Did eslint not want to single -quote everything here too?

src/charts/types.ts
```diff
@@ -0,0 +1,42 @@
...
+    dep_dt: Date;
+    arr_dt: Date;
+    current_dep_dt: Date;
+    travel_time_sec: number;
+    headway_time_sec: number;
+    dwell_time_sec: number;
+    benchmark_travel_time_sec: number;
+    benchmark_headway_time_sec: number;
+    threshold_flag_1?: string;
+  }
```
austinjpaul, line 13: I think we sometimes get other threshold flags also? We should be careful this won't break with an unexpected response. Perhaps that will require cleaning up our backend to return all standard fields (and only required fields) each time.

austinjpaul (commented):
This looks good, mostly thinking out loud. Would approve but I'm suddenly worried about our DataPoint type breaking with an unexpected response.

src/App.js
devinmatte: Basically prettier detected a space, and to ensure it isn't removed by newline reformatting, it adds it as a string `' '` injected into the resulting html

src/charts/Title.tsx
devinmatte: I was trying to avoid changing logic in this. As of today the code gets a `string[][]` at runtime, so I typed it accordingly. `ctx` is a `CanvasRenderingContext2D` today at runtime. `chart` is a Chart object, but, there's no included Typescript type. So I'd have to write my own, and that seemed like too much work for this PR.

The typing seems to pass today. If chartjs changes things, then Typescript will warn us with type errors, which is more than we'd get today. If we update chartjs, this gives us a better warning.

I think it's good to remember that Typescipt is a compile-time typechecker, it runs Javascript at runtime with no additional typechecks

src/ui/ProgressBar.js
devinmatte: Yeah. Didn't seem worth addressing now, so, adding the warning ahead of time

src/index.js
devinmatte: It will single quote it if we changed to `path={'/opensource'}>` without `{}` it uses `"`. It's prettier opinion

src/charts/types.ts
devinmatte: Types like this are compile time checks, not runtime. If the code never tries to access `. threshold_flag_2` explicitly, Typescript won't help us, and won't catch it. At runtime it runs as Javascript.

austinjpaul (approved):
All sounds great, thanks for the clarification!

PR #238, opened 2022-10-18
[v4 DEV] A working chart in ChartJs 3.9.1

devinmatte:
Screen Shot 2022-10-18 at 12 49 22 PM

### DIFF
--- a/pages/SingleDayLineChart.tsx
+++ b/pages/SingleDayLineChart.tsx
@@ -15,6 +15,7 @@ import { enUS } from 'date-fns/locale';
 import React from 'react';
 import { drawTitle } from './Title';
 import { DataPoints } from './types/dataPoints';
+import { colors } from './constants';
 
 ChartJS.register(
   CategoryScale,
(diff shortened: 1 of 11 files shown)


devinmatte:
This is Demo/Dev work, no need to fully review. There will be a full review at a later date

pages/SingleDayLineChart.tsx
```diff
@@ -0,0 +1,189 @@
...
+    if (point[benchmark_field] === null) {
+      return '#1c1c1c'; //grey
+    } else if (ratio <= 1.25) {
+      return '#64b96a'; //green
+    } else if (ratio <= 1.5) {
+      return '#f5ed00'; //yellow
+    } else if (ratio <= 2.0) {
+      return '#c33149'; //red
+    } else if (ratio > 2.0) {
+      return '#bb5cc1'; //purple
```
skaplan-dev, line 55: could we have these as constants somewhere?

skaplan-dev (approved):

austinjpaul:
No more fill for the benchmark?

pages/SingleDayLineChart.tsx
devinmatte: Yeah we could. I just copy pasted from the current `src`. I'll move these to a better place

devinmatte:
It doesn't seem to be taking @austinjpaul. Still trying to figure out what the new attributes needed to make that work again

PR #239, opened 2022-10-19
[V4 Dev] Add routing and slow zones component

skaplan-dev:
- Adds Next.Js routing
- Re-organizes some code
- Adds SlowZone component


components/Layout.tsx
```diff
@@ -0,0 +1,10 @@
+import Navbar from './Navbar';
+
+export default function Layout({ children }) {
```
devinmatte, line 3: eslint should be throwing an error on this, we should avoid default exports when possible. There was a discussion about this as an enforced code style a little while ago

devinmatte (approved):

components/Layout.tsx
skaplan-dev: not seeing an eslit error on my machine :( but I'll update

PR #241, opened 2022-10-19
[v4 DEV] Showing alerts and using node 18

devinmatte:
Screen Shot 2022-10-19 at 5 02 59 PM

- Shows alerts in the v4 dashboard

Issues:
- I can't get tooltips to display. Using them the same as in v3, and I think next.js is getting in the way


styles/globals.css
```diff
@@ -51,21 +66,112 @@ a {
 }
 
 @media screen and (max-width: 770px) {
-.chart-extras {
-  margin-left: 44px;
-  display: flex;
-  justify-content: space-between;
+  .alerts {
```
devinmatte, line 69: @skaplan-dev I think I need to look further into how stylesheets are supposed to work in next.js, because this can't possibly be the correct way to include styles, but the react way to handle style imports throws errors.

Also we're doing `.css` files, so nesting isn't an option? It looks like we need a plugin for nested css with postcss?

skaplan-dev: This is the only thing I can find https://nextjs.org/docs/basic-features/built-in-css-support

skaplan-dev (approved):

PR #244, opened 2022-10-23
fix date selector cut-off

### LINKED ISSUE #217: slowzones: date selector range cut-off on small screens
notice top right: 
image

danielreynoldsma:
this affects both the regular data dashboard and the slowzones tracker

this fixes #217 

attached are clips of both

data-dashboard-sizing
slowzones-sizing


danielreynoldsma:
to clarify, it adds an intermediate stage between phone-screen width and desktop-screen width

skaplan-dev (approved):

austinjpaul:
I want to make sure the chart legends and toggles look okay on the intermediate size before this goes in.

danielreynoldsma:
that would be good to make sure, I sort of haphazardly used the code from the other two stages

austinjpaul:
image
A few things caught my eye:
1) Can we close the space between the toggle and the line selector?
2) Can we add some space here between the left and right components? Maybe the end date picker could then be aligned with the selector above it and the little x can float in the region off to the right? No need if that's too complicated though.
3-4) This should at the very least be aligned with the chart like for mobile. You might just want to borrow the mobile code entirely for these components.

I'm no expert on css, but maybe we should add a section (<1300px) for the shared chart-extras/legend stuff, and then reserve the (770px-1300px) section for the tablet-specific config styles and the (<720px) section for the mobile-specific config styles.

danielreynoldsma:
image
here is the fix for 3/4

as for 1/2, i'm not quite sure how to change it, but even if i did im not sure it'd look much better.
here's how the same layout looks in single-date mode:

image

not sure it's a huge deal either way. let me know!

danielreynoldsma:
@austinjpaul what do you think? forgot to mention you

austinjpaul (approved):
Great! Yeah, this was all very minor stuff- very much an improvement over what we had. 
Thanks again for getting this in, and feel free to merge whenever.

PR #250, opened 2022-10-26
Use Next.JS 13 app folder structure

skaplan-dev:
- removes the `pages` folder in exchange for the new next.js `app` folder


app/layout.tsx
```diff
@@ -0,0 +1,28 @@
+'use client';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import React from 'react';
+import { Layout } from '../components/Layout';
+import '../styles/dashboard.css';
+import '../styles/globals.css';
+const queryClient = new QueryClient();
+
+export default function RootLayout({
```
devinmatte, line 9: We should avoid default exports in favor of explicit exports
```suggestion
export function RootLayout({
```

app/layout.tsx
```diff
@@ -0,0 +1,28 @@
...
+  // Layouts must accept a children prop.
+  // This will be populated with nested layouts or pages
+  children,
+}: {
+  children: React.ReactNode;
+}) {
```
devinmatte, line 15:
```suggestion
export const RootLayout = ({
  // Layouts must accept a children prop.
  // This will be populated with nested layouts or pages
  children,
}: {
  children: React.ReactNode;
}) => {
```

You can also get cleaner React Typescript typing by using the `React.FC<{
  children: React.ReactNode;
}>` typing

```suggestion
export const RootLayout: React.FC< {
  children: React.ReactNode;
}> = ({
  // Layouts must accept a children prop.
  // This will be populated with nested layouts or pages
  children,
}) => {
```

app/page.tsx
```diff
@@ -1,14 +1,15 @@
-/* eslint-disable import/no-default-export */
-import React from 'react';
-import { SingleDayLineChart } from '../components/dashboard/charts/SingleDayLineChart';
+'use client';
+
+import { useState } from 'react';
```
devinmatte, line 3: We should keep React in scope for Typescript
```suggestion
import React, { useState } from 'react';
```

app/slowzones/error.tsx
```diff
@@ -0,0 +1,17 @@
+'use client';
+
+import { useEffect } from 'react';
+
+export default function Error({ error, reset }: { error: Error; reset: () => void }) {
```
devinmatte, line 5: We should avoid default exports: https://github.com/transitmatters/t-performance-dash/pull/201#discussion_r934063189
We had discussed retiring them, and eslint should be yelling at you (do you have eslint on in your editor?)

components/Navbar.tsx
```diff
@@ -25,12 +26,12 @@ export const Navbar = () => {
                   </div>
                   <div className="flex flex-1 items-center justify-center sm:items-stretch sm:justify-start">
                     <div className="flex flex-shrink-0 items-center">
-                      <img
+                      <Image
```
devinmatte, line 30: Does this display now? I was running into an issue with it in next 12

devinmatte (approved):

components/Navbar.tsx
skaplan-dev: ye

PR #253, opened 2022-10-28
Temporarily fix cors issue

skaplan-dev:


devinmatte:
Can you include fetchDelayTotals in this PR, the file isn't committed

skaplan-dev:
> Can you include fetchDelayTotals in this PR, the file isn't committed



> 

yeah just pushed it

PR #246, opened 2022-10-23
Add healthcheck to GitHub Actions

nathan-weinberg:
This is a quick and dirty CI implementation using the `/healthcheck` endpoint introduced in #88 

Very open to suggestions here - for example right now the job is running on PRs and merges to `main`, but we could also have it be an independent job that fires on a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) instead.

PR #254, opened 2022-10-29
Add MbtaPerformanceAPI test to healthcheck

mathcolo:
- Try to grab data from realtime.mbta.com in the healthcheck
- Also print the failure reason if it exists


nathan-weinberg (commented):
@mathcolo The new test looks good - for the handling of the functions/responses as what are your thoughts of the format I have in https://github.com/transitmatters/t-performance-dash/pull/246?

austinjpaul:
I really like @nathan-weinberg's improvements to the checking functions. Can we adopt that portion of #246?
The check here is a good baseline. I would also want to record the response time and give a warning if it's above a certain length (say, 10s), but that might be overkill for right now.

mathcolo:
Yep! @nathan-weinberg would you like to pull in what I have here into #246 ? Or the other way around?

nathan-weinberg:
@mathcolo @austinjpaul I've updated https://github.com/transitmatters/t-performance-dash/pull/246 with y'alls feedback - maybe we could merge that one if it looks good then add this test afterword?

nathan-weinberg:
I can't seem to deploy to beta as I was saying on Slack, so if someone else can deploy the PR I can verify it's working as expected.

nathan-weinberg:
Discussed internally, changes will be incorporate into #246

Issue #237, opened 2022-10-18
Slowzone links on beta go to main dashboard


austinjpaul:
This disguised a bug that we would have otherwise caught before deploying.
We should probably fix it so links stay on the beta server.

PR #256, opened 2022-11-08
Slowzone links fix

PatrickCleary:
Fix for Issue #237 

Changed the links on the slowzones page to link to the current URL rather than always linking to production.


austinjpaul (approved):
Looks good to me, thanks!

Issue #225, opened 2022-09-28
Headway cap


austinjpaul:
e.g. https://dashboard-beta.transitmatters.org/bus?config=15,15-1-322,15-1-64000,2022-07-13
Some bus schedules lead to actual (and scheduled) headways that are many hours long. Per Chris Friend ,we should set any headways and expected headways > 120 minutes to null.

PR #257, opened 2022-11-08
Cap headways at 2 hours

PatrickCleary:
For issue #225  

Set expected and actual headway values to null when expected headway is greater than 2 hours.


austinjpaul:
I need to think a bit more about this-- I'm debating if this change should be made in the backend instead.

austinjpaul:
This fix will go in the backend (part of #208)

Issue #121, opened 2021-12-16
Redundant Date selection


austinjpaul:
If the user selects the same date that is already selected, the data reloads.
Same when they close an empty end-date picker.
We should somehow prevent updateConfiguration from doing anything if the configs haven't actually changed (or just not call it in the first place).

PR #259, opened 2022-11-09
Prevent update on selection of same date

PatrickCleary:
Issue #121: The Flatpickr object was firing onChange for date selection even when the currently selected date was chosen.

This is because Flatpickr always fires onChange on selection of a date ([doc]( https://flatpickr.js.org/events/)). Our `DatePicker` object has a separate element for use on mobile `RegularDateInput` which does not have this issue. So I added logic only on the Flatpickr. Not sure if it would be more clear to do just do it on both or leave a comment explaining this in the code. lmk.


austinjpaul (approved):
👍

Issue #126, opened 2021-12-16
Back button memory utilization


austinjpaul:
Because we are saving the entire state in the history stack, that includes all the data (i think?). Perhaps we just want to save the configuration and the error_message state.
Alternatively, we could use the saved data to prevent a re-download when using browser back/fwd navigation.

PR #261, opened 2022-11-10
reduce history memory

PatrickCleary:
Issue #126 

**Issue:**
React router history was storing all of the current app state in the browser. This included all data points in any visualization shown.

**Fix:**
Changed history to only save configuration and error message.

**Other Changes:**
- I was having a hard time understanding the code which was dealing with history and updating the URL query parameters. If people don't think the changes I made are improvements, I'm happy to remove them.
- One change was to remove the trailing comma at the end of the links when there is no `end_date`. Like this one: `/rapidtransit?config=Red,70061,70067,2022-11-09,` will now be: `/rapidtransit?config=Red,70061,70067,2022-11-09`
- Another change was to not add any parameters to the URL Query when the form is not complete. I figured there is no need to have a URL such as this one: `/rapidtransit?config=Red,,,,` which just links to an empty dashboard with the red line selected.
- Removed `update.error` field, as it was not being used anywhere.

I think it might be good to transition away from using an array of values as a single query parameter. If we need to add another optional field like `end_date` it will be easier to do so without adjusting the ordering of all the fields. I can make a ticket for this if there is agreement.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -47,6 +47,11 @@ const formIsComplete = (config) => {
     return line && from && to && date_start
 }
 
+const formIsEmpty = (config) => {
+    const { line, from, to, date_start, date_end } = config;
+    return !line && !from && !to && !date_start && !date_end
+}
+
 // TODO: Replace query array with individual fields. 
 const getQueryStrings = (config) => {
   const { line, from, to, date_start, date_end } = config;
@@ -58,6 +63,10 @@ const getQueryStrings = (config) => {
 const urlFromConfig = (config) => {
   const path = config.bus_mode ? BUS_PATH : RAPIDTRANSIT_PATH;
   const query = getQueryStrings(config);
+  // If nothing is entered, return without config query.
+  if(!query.some(piece => piece)) {
+    return path;
+  }
 
   // Remove empty fields from array.
   const queryString = query.join(',');
@@ -120,7 +129,8 @@ class App extends React.Component {
 
     const url_config = new URLSearchParams(props.location.search).get('config');
     if (typeof url_config === 'string') {
-      this.state.configuration = stateFromURL(props.location.pathname, url_config);
+      const config = stateFromURL(props.location.pathname, url_config);
+      this.state.configuration = config
       this.state.error_message = this.checkForErrors(this.state.configuration);
     }
 
@@ -133,11 +143,11 @@ class App extends React.Component {
     }
 
     // Handle back/forward buttons
-    window.onpopstate = (e) => {
-      this.setState(e.state?.state, () => {
+    window.addEventListener("popstate", (e) => {
+      this.setState(e.state.state, () => {
         this.download();
       });
-    };
+    });
 
     this.download = this.download.bind(this);
     this.updateConfiguration = this.updateConfiguration.bind(this);
@@ -146,6 +156,9 @@ class App extends React.Component {
   }
 
   componentDidMount() {
+    // Save configuration to history on load. This allows the user to go back to the dashboard with no fields filled.
+    // We also need to get the proper current URL to replace for the case when someone navigates to the page via a pre-configured URL.
+    this.props.history.replace(`${this.props.location.pathname}${this.props.location.search}`, {configuration: this.state.configuration});
     this.download();
   }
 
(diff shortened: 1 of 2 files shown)


austinjpaul:
Ooh, glad to see this issue getting some attention! Hopefully I can take a fuller look tomorrow.
Regarding future changes to the URL structure: it will be important to keep backwards compatibility, as we probably don't want to break links to our pages that people have been posting in forums, on twitter, etc.

PatrickCleary:
I didn’t make any changes to how the URLs are parsed, so it shouldn’t affect any existing ones. I only changed how URLS are generated. Let me know if you have any questions though, this PR is a bit confusing I think.

PatrickCleary:
Just found an issue on this. Current behavior when switching between lines will clear the url when switching lines. i.e. if you are at `https://dashboard.transitmatters.org/rapidtransit?config=Red,70061,70065,2022-11-10,` and switch to orange line you will have the URL: `https://dashboard.transitmatters.org/rapidtransit?config=Orange,,,2022-11-10,` with this change you are at: `https://dashboard-beta.transitmatters.org/rapidtransit?config=Blue,70061,70065,2022-11-09` Which is not a valid URL since 70061 and 70065 are not stops on the blue line.

austinjpaul:
I think I'm still seeing commas and partial urls. Behavior seems pretty indistinguishable from the current. 
There is a bug (pre-existing) where the back button doesn't update the station selectors to remove the dest from the src field, e.g., so you can end up in a situation like : 
image

PatrickCleary:
I'm not able to reproduce the empty destination field bug. How did you get that behavior?

Also, I ended up removing the change to cut off the commas and partial URLs. I think the partial URLs actually have a use-case (i.e. sending someone a link to the orange line dashboard on a certain date without stops input). I thought that wasn't really necessary but now idk.

And if you are using those partial links, you can't remove the commas because the code to parse the URL relies on the positioning of the commas. We could fix this by converting each field into a separate query parameter. That would also make it easier if we ever needed to add additional fields, especially if other fields are optional (like date_end). I can make a new ticket for that.

I also think it would be good to use something like redux to manage the state of the form. But I believe they're adding redux in v4 so might be worth it to hold off for now.

austinjpaul (approved):
This clean-up is great! Behavior looks identical to today's site.
This should be good to merge, though i'll leave a couple of optional notes about pre-existing non-ideal behavior that you can address or leave for a future PR:
- Ideally if there are no pieces in the config, we should return just the base path (`/bus' instead of '/bus?config=,,,,,').
- Ideally the empty state above would also be pushed to history so people first landing on the page and filling it out can use the back button to return to the landing page.
- The bug I mentioned above. To repro, put in a full configuration. Then change the configuration (select a new date, e.g.). Then press the back button. Then, in the From Station field, you will be able to select the already-selected to-station.

Again, these are optional.
Like you said, while Redux might be nice, I think a change of that scale better left for v4.

PatrickCleary:
It was surprisingly confusing to get the correct behavior, but hopefully it works now. I think I've fixed the 3 things you mentioned @austinjpaul, I'll leave some more comments in the file diff about what I did.

src/StationConfiguration.js
```diff
@@ -89,12 +89,12 @@ export class StationConfiguration extends React.Component {
     }
     if (type === "from") {
       const toStation = this.getVal("to");
-      return options_station_ui(this.props.current.line).filter(entry => entry.value !== toStation);
+      return options_station_ui(this.props.current.line).filter(entry => entry.value?.stop_name !== toStation?.stop_name);
```
PatrickCleary, line 92: I think this was just an issue of object comparison. For some reason after going back, the two objects are no longer the same object, when previously they were. So I switched it to use stop_name, which I _think_ is always a unique field? Would be good to get that confirmed by someone who knows more.

Could also use Object.is(x, y) but that feels excessive.

src/App.js
```diff
@@ -138,6 +156,9 @@ class App extends React.Component {
   }
 
   componentDidMount() {
+    // Save configuration to history on load. This allows the user to go back to the dashboard with no fields filled.
+    // We also need to get the proper current URL to replace for the case when someone navigates to the page via a pre-configured URL.
+    this.props.history.replace(`${this.props.location.pathname}${this.props.location.search}`, {configuration: this.state.configuration});
```
PatrickCleary, line 161: This is the only way I was able to get the functionlity of:
1. open dashboard main page
2. Fill out form to load data
3. Press back and go back to empty dashboard.

It feels a little hacky but I spent a while on this and it's the only thing that worked. Open to suggestions

PatrickCleary: I just realized another solution may be to edit the onPopState function to get the configuration from the URL. Might be a cleaner solution.

austinjpaul: I'm happy with either approach. Thanks for the very clear comments too!

austinjpaul (approved):
Amazing! Glad to have this all working properly at last :)
LGTM

PR #266, opened 2022-11-13
Force poetry to use Python 3.9

nathan-weinberg:
Basically, this forces Poetry to use Python 3.9 for the virtualenv, even if a newer version exists on the system

Currently running on Beta: https://dashboard-beta.transitmatters.org/

### DIFF
--- a/server/poetry.lock
+++ b/server/poetry.lock
@@ -51,7 +51,7 @@ crt = ["botocore[crt] (>=1.21.0,<2.0a0)"]
 
 [[package]]
 name = "botocore"
-version = "1.23.53"
+version = "1.23.54"
 description = "Low-level, data-driven core of boto 3."
 category = "main"
 optional = false
@@ -67,11 +67,11 @@ crt = ["awscrt (==0.12.5)"]
 
 [[package]]
 name = "certifi"
-version = "2021.10.8"
+version = "2022.9.24"
 description = "Python package for providing Mozilla's CA Bundle."
 category = "main"
 optional = false
-python-versions = "*"
+python-versions = ">=3.6"
 
 [[package]]
 name = "chalice"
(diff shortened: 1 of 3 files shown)


server/poetry.lock
```diff
@@ -353,208 +352,39 @@ python-versions = "*"
...
-    {file = "wcwidth-0.2.5-py2.py3-none-any.whl", hash = "sha256:beb4802a9cebb9144e99086eff703a642a13d6a0052920003a230f3294bbe784"},
-    {file = "wcwidth-0.2.5.tar.gz", hash = "sha256:c4d647b99872929fdb7bdcaa4fbe7f01413ed3d98077df798530e5b04f116c83"},
-]
+ansicon = []
+attrs = []
+blessed = []
+boto3 = []
+botocore = []
+certifi = []
+chalice = []
```
mathcolo, line 365: Do we know why all of these are empty now?

nathan-weinberg: I was wondering that myself - I did a local `npm install` to regenerate my Poetry env and this is what happened :man_shrugging:

mathcolo: can you try deleting server/poetry.lock and then npm install again just to see what happens?

nathan-weinberg: Yes for sure!

nathan-weinberg: @mathcolo No change

mathcolo: alright!

mathcolo (approved):

server/poetry.lock
nathan-weinberg: if this breaks prod for some reason, we can def revert it lmao

austinjpaul: This looks to be a bug in poetry. Upgrade to poetry 1.1.15 and it should keep the hashes

austinjpaul:
Can we revert and actually fix this?

nathan-weinberg:
@austinjpaul See #268

PR #269, opened 2022-11-15
V4 aggregation charts

PatrickCleary:
WIP V4 Aggregation Charts. @skaplan-dev 

I manually got the data and put it into the data/headways_agg folder for this to work. There's probably a better way to do that. I can send the data to anyone who wants to review though.

* Add DateInput
* Add Aggregate Line charts (appear once start and end date are populated).


app/page.tsx
```diff
@@ -82,41 +91,148 @@ export default function Home() {
...
-        <div className={'charts main-column'}>
-          <SingleDayLineChart
-            chartId={'dwells'}
-            title={'Time spent at station (dwells)'}
-            data={dwellsData}
-            metricField={'dwell_time_sec'}
-            pointField={'arr_dt'}
-          />
-        </div>
+        {dateSelection?.endDate ?
```
PatrickCleary, line 101: Might be good to split this into two components. `SingleDayLineChartPage` and `AggregateLineChartPage` or something

components/dashboard/charts/AggregateLineChart.tsx
```diff
@@ -0,0 +1,175 @@
...
+  return (
+    <div className='chart'>
+      <div className="chart-container">
+        <div>
+          <Line
+            id={chartId}
+            height={250}
+            data={{
+              labels,
+              datasets: [
```
PatrickCleary, line 77: Maybe we want to manage chart options outside of this component? Might make it easier to share configuration between Single and Aggregate charts.

components/inputs/DateInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+                        onClick={() => setDateSelection({ ...dateSelection, range: true })}>
+                    Range...
+                </button>
+                //x button
+            }
+
+
+        </div>
+
+        // <div className="option option-date">
```
PatrickCleary, line 41: Forgot about this. This whole component needs to be cleaned up and it doesn't have all the functionality of the original date input, so I need to add that.

constants/colors.ts
```diff
@@ -0,0 +1,20 @@
+export const COLORS  = {
```
PatrickCleary, line 1: I made this so we can share colors between tailwind and non-tailwind.
Might be better ways to do this, I didn't do much research.

src/charts/types.ts
```diff
@@ -1,29 +1,29 @@
...
     direction: number;
-    dep_dt: Date;
-    arr_dt: Date;
-    current_dep_dt: Date;
-    travel_time_sec: number;
-    headway_time_sec: number;
-    dwell_time_sec: number;
-    benchmark_travel_time_sec: number;
-    benchmark_headway_time_sec: number;
+    dep_dt?: string;
```
PatrickCleary, line 4: I ended up changing a lot of these from dates to strings to get the ts compiler to stop complaining. Probably not ideal, I will look into this again.

types/dataPoints.ts
```diff
@@ -25,9 +24,7 @@ export interface HeadwayPoint {
   threshold_flag_3?: string;
 }
 
-export interface DwellPoint {
-  route_id: string;
-  direction: number;
```
skaplan-dev, line 30: Were these properties not used anywhere?

src/charts/types.ts
```diff
@@ -35,8 +35,25 @@ export interface DataPoint {
...
-  export type YField = 'travel_time_sec' | 'headway_time_sec' | 'dwell_time_sec';
+  export enum MetricFieldKeys {
+    travelTimeSec = 'travel_time_sec',
+    headWayTimeSec = 'headway_time_sec',
+    dwellTimeSec = 'dwell_time_sec',
+  }
+  export enum BenchmarkFieldKeys {
+    benchmarkTravelTimeSec = 'benchmark_travel_time_sec',
+    benchmarkHeadwayTimeSec = 'benchmark_headway_time_sec',
+  }
```
skaplan-dev, line 54: @devinmatte did we decide on enums vs union types vs typeOf const objects?

app/page.tsx
skaplan-dev: agreed!

skaplan-dev (approved):
This looks great! :)

types/dataPoints.ts
PatrickCleary: I moved those fields to a general `DataPoint` which is extended by each of the three types of points

src/charts/types.ts
PatrickCleary: Happy to go either way. Personally, I like the enums bc it makes it easy to find all the references, and less mystery strings in the code.

devinmatte: This is a habit I've picked up from work. Enums are probably better for this project. Enums take longer for the typescript compiler to compile, but union types are not only quicker but are smaller in the resulting bundle. So we have eslint rules that enforce them at work, so I've developed the habit. But for a project as small as ours enums are probably preferred as they are easier to work with despite the performance hit

devinmatte (approved):

devinmatte:
Frontend should be able to successfully pass the `lint` job in the v4 branch, so might want to address those

PatrickCleary:
I just fixed the warnings, but the errors are from

`4:29  Error: Unable to resolve path to module '../../../data/headways.json'.  import/no-unresolved`

and that folder  isn't being checked in to git.

How are you getting around that? I saw Seth's most recent PR didn't fail on that but I'm not sure why.

PatrickCleary:
Not sure what's going on there, but my next PR is getting rid of those files anyway so I'm gonna merge before these get too backed up.

PR #270, opened 2022-11-16
proxy for next.js

PatrickCleary:
Set up proxy to backend with next.js

https://nextjs.org/docs/api-reference/next.config.js/rewrites


skaplan-dev (commented):
Will this work once we push to prod?

PR #272, opened 2022-11-19
V4 single day fetch

PatrickCleary:
Added queries to get data for single day charts. Haven't used these technologies before, so feedback is more than welcome.

TanStack has default behavior to refetch all queries on page focus. This meant if you switched tabs, clicked on dev tools, etc. then the data will be fetched when you return to the site. I disabled this on the queries I added. Not sure if people agree with that. Might be worth discussing.

I also added React Queries Dev tools. Shows up as a little flower icon on the page when in development. It didn't end up helping me much with the bug I was working on but it seems useful. I'm fine with keeping or removing it.

One really cool thing with react queries I didn't know about was the queryKeys. Lets you define what causes the query to refetch. So I set it up that headways and dwells don't rely on toStopIds changing, since the values won't change, as they only rely on fromStopIds.

I added a rewrite in next.config.js, which takes the place of the `proxy` we had before switching to next.js. It's included based on process.env.DEV_NODE === 'development'. Which is when the command `next dev` is used. Pretty sure that's what we want but if someone knows better lmk.

### DIFF
--- a/components/dashboard/charts/SingleDayPage.tsx
+++ b/components/dashboard/charts/SingleDayPage.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
-import { useQuery } from '@tanstack/react-query';
 
 import { BenchmarkFieldKeys, MetricFieldKeys, PointFieldKeys } from '../../../src/charts/types';
 import { stopIdsForStations } from '../../../utils/stations';
-import { fetchSingleDayData } from '../../../api/datadashboard';
+import { useQuerySingleDayData } from '../../../api/datadashboard';
 import { Station } from '../../../types/stations';
+import { QueryNameKeys, SingleDayAPIKeys } from '../../../types/api';
 import { DateOption } from '../../../types/inputs';
 import { SingleDayLineChart } from './SingleDayLineChart';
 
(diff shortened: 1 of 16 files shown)


components/dashboard/charts/SingleDayPage.tsx
```diff
@@ -1,25 +1,66 @@
...
+
+export const SingleDayPage: React.FC<SingleDayPageProps> = ({ configuration }) => {
+  const { fromStation, toStation, dateSelection } = configuration;
+  const { fromStopIds, toStopIds } = stopIdsForStations(fromStation, toStation);
+  const date = dateSelection?.startDate;
+  const queryReady = !!(fromStopIds && date);
+
+  //TODO: deal with errors
+  const headwaysRequest = useQuery({
+    refetchOnWindowFocus: false,
```
devinmatte, line 27: We should set up some default settings. 
```
    keepPreviousData: true,
    refetchOnWindowFocus: false,
    retry: 0,
    staleTime: 10000 // 10 seconds
```

devinmatte (commented):
Great job so far, can't test but have offered some rational defaults for useQuery from my experience with it. You had the right idea with turning off `refetchOnWindowFocus`

components/dashboard/charts/SingleDayPage.tsx
PatrickCleary: Refactoring this a bit. Seems like we can either pass defaults to each `useQuery` or we can wrap the app in a `queryClient` which holds these defaults along with some other configuration. However, to pass it globally we need to override the app component [(doc)](https://nextjs.org/docs/advanced-features/custom-app). It doesn't seem like there's any downside to doing that but I might be missing something since I don't have experience with Next.

The third option would be to wrap each page with the QueryClientProvider. Any thoughts?

PatrickCleary: Nevermind, just realized this is in the rootlayout component

PatrickCleary:
Added the query client defaults, I ended up not using keepPreviousData since it was causing some odd behavior on the charts. (more on that below). 
I also did some kinda funky stuff to get the `SingleDayAPIOptions` type to work as I wanted. Open to suggestions on that.

weird chart behavior: When I change the date, the old data stays at first but since the `date` is changed, the x axis on the graph changes. For example if I start with 11/19/22 then switch to 11/01/22 the X axis will be from 11/01/22 6:00 AM to 11/20/22 1:00 AM. All the existing data squishes up to the right side until the new data loads. I'm also concerned it might not be the best user interface since if the new data fails to fetch there wouldn't be any indication that it has failed, and the old data would still be there. We could probably solve that by watching the query and displaying some sort of error. But at that point I'm wondering if we're gaining anything from keeping the data.

devinmatte:
Screen Shot 2022-11-20 at 11 34 30 AM
Not sure if this is a result of this PR or the last, but looks like the legend is now underneath the chart instead of on the side. I think some css was lost. Doesn't need to be addressed here, just noting it

devinmatte (approved):

PR #275, opened 2022-11-22
Add fetch for aggregate charts

PatrickCleary:
Adding fetch for aggregate API to use in aggregate charts.

I think something that is causing issues is that the home page is loading `SingleDayPage` before any inputs are made. So many props of `SingleDayPage` are null (date, fromStation, etc.).  This propagates to all subcomponents which then also have date, fromStation, etc. set to null. I think some sort of default landing page (as we've discussed) before these inputs are made will help that situation. 

Things I could use feedback on:
1. I ended up making three separate paths for fetching dashboard data. Previously we had one which had logic to separate out the types.
The reasons I did that:
- Easier to read
- Issues with types. Single day, Aggregate (headways & dwells), and aggregate (traveltimes) return data in 3 distinct formats. This made it difficult to assign types to the functions and caused issues with unexpected fields when using the data as a prop.

Downsides: more lines of code, more paths, others (?)

2. The type I created in `types/api.ts` and used in the aggregate page and the dashboard API. I wanted a map from API parameters to values. Ended up having to make this type: `export type AggregateAPIOptions = { [key in AggregateAPIKeys]?: string[] | string };` which is a little funky, but does the job. Open to suggestions if anyone knows how to do this better.

3. Fixed the issue @devinmatte was probably having. In `next.config.js` I had used rewrites to send API calls to port `http://localhost:5000`. For some reason with more up to date versions of Node this does not work so I changed it to proxy to `http://127.0.0.1:5000`. I assume this is backwards compatible if anyone is on older versions of node but I didn't check. Something to keep an eye on.


devinmatte:
@PatrickCleary We might want to enforce Node 18 if there are issues related to node versions. We should simply be on the latest LTS. We don't have legacy users or anything so there's no reason to allow older versions

components/dashboard/charts/AggregatePage.tsx
```diff
@@ -20,14 +25,51 @@ TODOS:
...
+      [AggregateAPIKeys.end_date]: dateSelection?.endDate || '',
+    },
+    QueryNameKeys.dwells,
+    queryIsReady
+  );
+
   return (
     <div>
       <div className={'charts main-column'}>
         <AggregateLineChart
```
skaplan-dev, line 67: could we add a loading spinner here for each chart instead of doing these null & undefined checks?

components/dashboard/charts/AggregatePage.tsx
```diff
@@ -20,14 +25,51 @@ TODOS:
...
+  );
+  const dwellsRequest = useQueryAggregateData(
+    {
+      [AggregateAPIKeys.stop]: fromStopIds || [],
+      [AggregateAPIKeys.start_date]: dateSelection?.startDate || '',
+      [AggregateAPIKeys.end_date]: dateSelection?.endDate || '',
+    },
+    QueryNameKeys.dwells,
+    queryIsReady
+  );
```
skaplan-dev, line 62: this seems like a lot of duplicate code. Could we have a function that returns an object with all 3 requests?

api/datadashboard.ts
```diff
@@ -25,10 +25,68 @@ export const useQuerySingleDayData = (
...
+export const fetchTraveltimes = (
+  name: string,
+  options: AggregateAPIOptions
+): Promise<TravelTimesRequest> => {
+  const url = new URL(`${APP_DATA_BASE_PATH}/aggregate/traveltimes2`, window.location.origin);
+  Object.entries(options).forEach(([key, value]) => {
+    if (Array.isArray(value)) {
+      value?.forEach((subvalue) => url.searchParams.append(key, subvalue));
+    } else {
+      url.searchParams.append(key, value?.toString() || '');
```
skaplan-dev, line 88: whats the case that we have a key and no value here?

skaplan-dev (changes requested):
Few comments / questions!

components/dashboard/charts/AggregatePage.tsx
PatrickCleary: I ended up changing it so these components are not rendered until all the user input is complete.

If you are talking about the `data || []` checks in the chart, I agree we need some loading symbol but I think it will be a bit more complicated. For example if the fetch failed vs. if the data is just empty.
For general loading, there's the react-query isLoading property which I'm planning to use for a little loading symbol.
Idk if that's what you're referring to. Lmk.

components/dashboard/charts/AggregatePage.tsx
PatrickCleary: Yes, good point. I refactored this a bunch.

api/datadashboard.ts
PatrickCleary: good catch. This got fixed by preventing the charts from showing up until user input was finished as well.

PatrickCleary:
Just updated. A couple things I could use input on:

- In `api/datadashboard.ts` I wanted to import `UseQueryResult`. For some reason ESLint was giving me an error saying that it is not exported from 'react-query' even though it compiled fine.  Ended up using `ReactQuery.UseQueryResult`

- At some point, I did an `npm install` and a bunch of `swc` packages were added. Seems like that's expected though because they're for essentially every architecture and all flagged as optional. I checked my `node_modules` and only darwin-x64 was actually downloaded which is what I use. Seems correct to me. 

- workflows are failing because I setup .npmrc to require node 18 as suggested by @devinmatte and the workflows are ran on 16. Not sure what we wanna do there.

- I wanted to make startDate and endDate required in `AggregatePage` and `SingleDayPage`. I did with some type guards in a new component `ChartPages`. Felt a little funky, so please let me know if that doesn't look great.

devinmatte (approved):
@PatrickCleary This is g2g imo

PR #278, opened 2022-12-11
Add GLX-Medford stations

mathcolo:


friendchris (approved):
Looks good. Can we enable temporarily to see if they publish data tomorrow?

PR #280, opened 2022-12-17
Ol cleanup

austinjpaul:
This will remove the orange line shutdown-related emoji, and change the line graph description for the backend changes.

### DIFF
--- a/src/slowzones/formattingUtils.ts
+++ b/src/slowzones/formattingUtils.ts
@@ -11,9 +11,7 @@ const textSize = isMobile ? '11' : '14';
 const DAY_MS = 1000 * 60 * 60 * 24;
 
 export const EMOJI = {
-  'derailment': ` 🚨`,
-  'construction': ` 🚧`,
-  'shutdown': ` ⚠️`,
+  'derailment': ` 🚨`
 }
 
 const getFootnoteIcon = (start: Moment, end: Moment, color: string) => {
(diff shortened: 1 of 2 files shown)


devinmatte (commented):
~So I feel like we should remove the gridlines, but still keep the indicators on the lines like we kept the derailments. Otherwise that 30 day drop and spike looks odd to a new viewer~

Ignore me, that's exactly what you did

PatrickCleary (approved):

src/slowzones/formattingUtils.ts
```diff
@@ -11,9 +11,7 @@ const textSize = isMobile ? '11' : '14';
 const DAY_MS = 1000 * 60 * 60 * 24;
 
 export const EMOJI = {
-  'derailment': ` 🚨`,
-  'construction': ` 🚧`,
-  'shutdown': ` ⚠️`,
+  'derailment': ` 🚨`
```
mathcolo, line 14: does this need to be a template literal?

mathcolo (approved):

src/slowzones/formattingUtils.ts
austinjpaul: I do think we will come back and add more here later. (MBTA prove me wrong...)

mathcolo: nooo I mean backticks `` vs regular string ''

austinjpaul: Ohhhhh haha sorry. Probably not, that's my bad

PR #299, opened 2023-01-09
V4/design updates

PatrickCleary:
- Added Homescreen Widget Title Component
- Updated header to match Figma design
- Updated SVG imports to use Babel
- Added bottom nav bar
- Added line selector component

This includes the first changes for a new concept for the navigational hierarchy.

The hierarchy of pages is:
1) <lineName>
2) <sectionName> Commute/Service/Trains
3) <statisticName> {
Commute: [general, travel times, headways, dwells, etc.]
Service: [general, service, ridership, etc.]
Trains: [New Train tracker]
}

So you can be on the page `redLine/Commute/TravelTimes` or `orangeLine/Service/General`


PatrickCleary:
Frontend lint is failing on this now:
```7:64  Error: Unable to resolve path to module '../../app/data/tailwind'.  import/no-unresolved ```

Not sure why, it doesn't fail locally.

devinmatte:
I think the data dir is .gitignored. So nothing under it is committed

PatrickCleary:
Ah, good catch. No idea why that's even in there lol

devinmatte (approved):

PR #300, opened 2023-01-16
Ignore TS Errors on build temporarily

PatrickCleary:
Temporary work-around to get the build to work while sorting out typescript issues. As per: https://transitmatters.slack.com/archives/GSJ6F35DW/p1673839728818259 @mathcolo 

I set it up to use a separate command to prevent us from accidentally deploying a build to prod with this workaround.


mathcolo (approved):
Yesssssssss

PR #305, opened 2023-01-17
Add Section Selector

PatrickCleary:
Add section selector.

### DIFF
--- a/components/Dropdowns/LineSelector.tsx
+++ b/components/Dropdowns/LineSelector.tsx
@@ -2,10 +2,9 @@ import { Popover, Transition } from '@headlessui/react';
 
 import React, { Fragment } from 'react';
 import { LINES, LINE_OBJECTS } from '../../constants/lines';
-import SelectedLineIndicator from '../../public/Icons/Components/SelectedLineIndicator.svg';
 
 import { classNames } from '../utils/tailwind';
-import { lineSelectionButtonConfig, lineSelectionConfig } from './LineSelectorStyle';
+import { buttonConfig, lineSelectionButtonConfig, lineSelectionConfig } from './LineSelectorStyle';
 
 const LineSelectionItem = ({ lineName, selectedLine }) => {
   const isSelected = lineName === selectedLine;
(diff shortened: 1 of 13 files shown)


components/Dropdowns/LineSelector.tsx
```diff
@@ -34,34 +33,33 @@ const LineSelectionItem = ({ lineName, selectedLine }) => {
 
 export const LineSelector = ({ selectedLine }) => {
   return (
-    <Popover className="relative">
+    <Popover className="relative flex h-full items-center">
       {({ open }) => (
         <>
-          <Popover.Overlay className="fixed inset-0 bg-black opacity-30" />
+          <Popover.Overlay className="mb-safe fixed inset-0 bottom-11 bg-black opacity-30" />
```
PatrickCleary, line 39: mb-safe is needed for chrome on mobile. When the bottom navigation collapses, it overlaps with the home slide-up thingy on iphone.

components/general/StatSelectionSlider.css
```diff
@@ -0,0 +1,147 @@
+.SSI-Invisible {
```
PatrickCleary, line 1: Lots of animation stuff that is kinda messy, but I think I'll rework this component entirely later on.

styles/globals.css
```diff
@@ -39,3 +42,12 @@ a {
   width: 92%;
   max-width: 1200px;
 }
+
+.pb-safe {
```
PatrickCleary, line 46: This prevents the site from overlapping with the iphone home bar.

skaplan-dev:
Could you seperate out the statistics changes and nav changes?

styles/globals.css
```diff
@@ -4,11 +4,14 @@
 
 :root {
   --transitmatters-red: #d31a2b;
-  --red-line: #da291c;
+  --red-line: #d13434;
```
skaplan-dev, line 7: did we have the colors wrong here?

PatrickCleary:
The red I have been using in the designs is the same red as the NTT, which
is different from the regular red. Now that you point this out though, I
probably shouldn’t have changed this value if it will affect the old
dashboards. I’ll fix this.

skaplan-dev (approved):
feel free to merge once we figure out the color stuff

PR #316, opened 2023-01-24
Travel Times Widget and Skeleton

devinmatte:
Screenshot 2023-01-24 at 6 54 25 PM


devinmatte:
Not sure what's going on with the red on the graph, doesn't happen on the standalone page, so likely so css on the homepage I'm missing

components/widgets/TravelTimesWidget.tsx
```diff
@@ -0,0 +1,96 @@
+'use client';
+import React from 'react';
+import classNames from 'classnames';
+import ArrowDownNegative from '../../public/Icons/ArrowDownNegative.svg';
+import { SingleDayLineChart } from '../dashboard/charts/SingleDayLineChart';
+import { BenchmarkFieldKeys, MetricFieldKeys, PointFieldKeys } from '../../src/charts/types';
+import { SingleDayAPIKeys } from '../../types/api';
```
skaplan-dev, line 7: sorry I didn't catch this earlier but could we avoid the term `Keys` for these types? 

we should avoid using the variables `type` in its name.

skaplan-dev (approved):
🔥 just one comment re: naming some types

components/widgets/TravelTimesWidget.tsx
devinmatte: I'm curious what a better name would be. We use `Keys` in a lot of variables, since it's a Key in a key:value pair. `Key` makes the most sense to me here

skaplan-dev: Maybe "MetaData"?

devinmatte: I went for `Params`, but the more I think about it the more I think `Keys` fits better

PR #317, opened 2023-01-25
[v4 dev] Adding a basic headways widget

devinmatte:
Added a basic headways widget with real data

Updated travel times to show a real number (real-ish)

### DIFF
--- a/utils/date.ts
+++ b/utils/date.ts
@@ -1,0 +1,7 @@
+export const getCurrentDate = () => {
+  const iso_date = new Date();
+  const offset = iso_date.getTimezoneOffset();
+  const local_date = new Date(iso_date.valueOf() - offset * 60 * 1000);
+  const maxDate = local_date.toISOString().split('T')[0];
+  return maxDate;
+};
--- a/utils/date.ts
+++ b/utils/date.ts
@@ -1,7 +1,7 @@
 export const getCurrentDate = () => {
-  const iso_date = new Date();
-  const offset = iso_date.getTimezoneOffset();
-  const local_date = new Date(iso_date.valueOf() - offset * 60 * 1000);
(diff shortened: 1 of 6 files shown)


PatrickCleary:
Missed the last PR, but I was gonna suggest we keep the home screen widgets simple. I was thinking about just a single data point like median for headways and travel times for the day. Then if the user clicks in, they see the graph.

 maybe on desktop we could expand those widgets to include the graphs.

components/widgets/TravelTimesWidget.tsx
```diff
@@ -47,6 +49,23 @@ export const TravelTimesWidget: React.FC = () => {
     false
   );
 
+  const averageTravelTime = React.useMemo(() => {
```
PatrickCleary, line 52: I don't think we need to answer this on this PR, but I've been wondering if we should use a separate API route for homescreen widgets. As we add more widgets, it may become slow if we have to do lots of aggregations on the frontend.

utils/date.ts
```diff
@@ -0,0 +1,7 @@
+export const getCurrentDate = () => {
+  const iso_date = new Date();
+  const offset = iso_date.getTimezoneOffset();
+  const local_date = new Date(iso_date.valueOf() - offset * 60 * 1000);
```
PatrickCleary, line 4: Nitpick, but I think we should stick to camel case

PatrickCleary (approved):
Looks good to me. Thanks for doing this, it's coming along nicely.

skaplan-dev (approved):

utils/date.ts
devinmatte: Yeah can do, I copied this from existing v3 code and didn't reformat it

PR #318, opened 2023-01-25
DataPage Navigation

PatrickCleary:
Screen Shot 2023-01-25 at 8 16 29 PM
Screen Shot 2023-01-25 at 8 20 45 PM

First half of navigation. Allows us to navigate between lines and data pages. Not sure what we want on the root URL so I just put a super simple line selector for now.

Still need to update the main nav bar to include slow zones, NTT, etc.

Next rewrites don't play nicely with dynamic routing. Added a temporary fix to append `api/` to API requests. Will need to come up with a more permanent solution to fix this in dev.

Removed section selector.

Simplified data page selector. Have not added any styling to it so it's pretty plain.

Removed store since we can use router to track current page. Can probably remove Zustand, but we may use it in the future so I left it for now.


components/utils/router.tsx
```diff
@@ -0,0 +1,7 @@
+export const getPage = (path: string) => {
+  const pathItems = path.split('/');
+  return {
+    line: pathItems[1],
+    datapage: pathItems[2],
+  };
+};
```
skaplan-dev, line 7: you could add `useRouter` here and then you wouldn't have to call the hook in every component

```javascript
export const getPage = () => {
    const { asPath } = useRouter()
    const pathItems = asPath.split('/')
    return {
        line: pathItems[1],
        datapage: pathItems[2],
    };
};
```
I also think the name `getPage` is a little ambiguous since that's not _really_ what it does.
Maybe `getDelimitatedRoute`? Up for ideas.

i also think the return of this function could be typed. Right now all we know as a user of the hook is that it's returns an object with two properties that are strings.  We know the `line` can only be one of `LINE_OBJECTS`'s keys, and we know the list of possible `datapages`. This will help narrow it down

components/widgets/DataPageHeader.tsx
```diff
@@ -1,19 +1,36 @@
...
+import { getPage } from '../utils/router';
 import { classNames } from '../utils/tailwind';
 
 type DataPageHeaderProps = {
-  title: string;
   dateString: string;
-  line: string;
   children?: ReactNode;
 };
+const headerStyle = {
```
skaplan-dev, line 12: could we remove these styles into a styles file like we have with some other `button` styles?

skaplan-dev (approved):
looking pretty good! just a few code cleanup comments.
Could you add some images to your PR description of UI changes?

I'm also curious if you think we should change the line url to `Red / Orange / Blue` instead of `RL / GL / OL`?

skaplan-dev (approved):
lookin pretty good to me! just a few code optimizations / cleanup comments

could you add some images of any UI changes to the PR description?

components/Dropdowns/LineSelector.tsx
```diff
@@ -49,22 +52,24 @@ export const LineSelector = () => {
                     value={metadata}
                   >
                     {({ selected }) => (
-                      <div
-                        className={classNames(
-                          selected ? 'font-semibold' : 'font-normal',
-                          'flex flex-row gap-2 truncate'
-                        )}
-                      >
+                      <Link href={`/${metadata.key}`}>
```
skaplan-dev, line 55: could we put this link above `Listbox.Option`? Currently only the `text` is a link and not the whole button

skaplan-dev: I also think it would be cool to have the `selected` line keep its darker background

skaplan-dev: Something I noticed testing is that if you are in the slow zone drill down and nav to another line you get thrown back into the index route. I think it makes sense for users to be sent to the `slowzone` drill down of the line they route to.

components/utils/router.tsx
PatrickCleary: I did `useDelimitatedRoute` because I think we need to convert it to a custom hook to use useRouter().

PatrickCleary:
Screen Shot 2023-01-25 at 8 16 57 PM
Screen Shot 2023-01-25 at 8 16 57 PM
Screen Shot 2023-01-25 at 8 16 57 PM

PatrickCleary:
Forgot your first comment about orange vs OL for the routes before I pushed. I think that's a good suggestion though, happy to do it with the full color.

I can change in next PR if that's alright.

components/DashboardLayout.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { WidgetPage } from './widgets/Widget';
+
+export const DashboardLayout = ({ children }) => {
+  return (
+    <div className={classNames('flex w-full flex-col items-center')}>
+      <DataPageHeader dateString="Today (TBD)" />
+      <WidgetPage>{children}</WidgetPage>
+      <SecondaryNavBar />
+      <BottomNavBar />
+    </div>
```
skaplan-dev, line 15: We might want to change this as we think about the desktop layout. But its ok to keep for now 👍

PR #319, opened 2023-01-28
Add api/ to backend app.py routes

PatrickCleary:
We have set up [nextJS rewrites](https://nextjs.org/docs/api-reference/next.config.js/rewrites) to hit port 5000 on API requests.

Right now, our API has no prefix in the path, so any path can be an API request. When next JS sees a URL it doesn't recognize i.e. a [dynamic route](https://nextjs.org/docs/routing/dynamic-routes) or a page that should be a 404, it rewrites these pages to the API, which fails.

This change adds `api/` to all backend API routes to prevent that from happening, and changes the next config to send rewrites to `api/`. In the previous commit, we changed the frontend URLs for the new V4 dashboard.

These are 3 other possible solutions I considered but didn't use for various reasons:
1. Dynamically append `api` to URL in DEV
    * Would need to track env variables in the frontend. Feels yucky.
2. Add all APIs as a rewrite config individually.
    * Would need a master list of all APIS. Also yucky
3. Append “API/” to all URLS and rewrite on dev and prod.
    * We have to use `next export` to host the site statically. Next export does not support rewrites: https://nextjs.org/docs/advanced-features/static-html-export#unsupported-features

I wanted to avoid changing the backend but I  imagine adding api/ to the api routes is not a concern. If it is, I'm open to suggestions or pursuing one of the options I mentioned above. 

Also, if we are okay with this change, I will add another commit to update the v3 dashboard APIs.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -5,7 +5,7 @@ let rewrites = [];
 if (process.env.NODE_ENV === 'development') {
   rewrites.push({
     source: '/api/:path*',
-    destination: 'http://127.0.0.1:5000/:path*',
+    destination: 'http://127.0.0.1:5000/api/:path*',
   });
 }
 
--- a/server/app.py
+++ b/server/app.py
@@ -34,7 +34,7 @@ def mutlidict_to_dict(mutlidict):
     return res_dict
 
 
-@app.route("/healthcheck", cors=cors_config)
+@app.route("/api/healthcheck", cors=cors_config)
 def healthcheck():
     # These functions must return True or False :-)
     checks = {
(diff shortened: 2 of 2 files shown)


skaplan-dev (approved):

devinmatte (commented):
I think in theory this is good, just need to make sure this isn't a breaking change for any external api usage (i doubt it) and we should definitely port it to v3 so it's a seamless transition

austinjpaul:
> I think in theory this is good, just need to make sure this isn't a breaking change for any external api usage (i doubt it) and we should definitely port it to v3 so it's a seamless transition

A good plan. External API usage includes our slow-zone calculations, so we'll have to update that when we make the change.

mathcolo (approved):

devinmatte (approved):
We can move forward for v4, just have to work with @austinjpaul to migrate v3 before we launch v4 altogether

PatrickCleary:
Ok. Sounds like we are all good with this. I'll add a commit to add it to all v3 API calls.

We probably want to make sure we don't forget to update the slow zone API calls when we are ready to deploy though. I can make a draft PR to update those. I'll also create a tracker doc for things we need to prepare for the eventual V4 launch, as I imagine this won't be the last situation like this.

PatrickCleary:
Now that I think about it... Would it be easier to just update this now, before releasing v4?

PatrickCleary:
Going to merge this so we can get a demo working. Made this doc for tracking changes necessary for v4 release:
https://docs.google.com/document/d/1xpbvzYzNamlXLZ19S4EwuLN7kK7ExezdqmPCqqk1AA0/edit?usp=sharing

PR #323, opened 2023-01-31
Add Line Selector to navbar

PatrickCleary:
* Add line selector to navbar. Was able to reuse existing line selector for this.

* Add a new line selector for the sidebar and mobile slide-in menu.

* A few style changes to the existing line selector

Screen Shot 2023-01-31 at 10 26 32 AM
Screen Shot 2023-01-31 at 10 26 41 AM
Screen Shot 2023-01-31 at 10 26 53 AM


components/Dropdowns/LineSelector.tsx
```diff
@@ -1,85 +1,125 @@
...
-export const LineSelector = () => {
+interface LineSelectorProps {
+  value?: NavBarItem;
+}
+
+// LineSelector used by the top navbar in desktop mode, and bottom navbar in mobile mode.
+export const LineSelector: React.FC<LineSelectorProps> = ({ value }) => {
   const route = useDelimitatedRoute();
+  // `value` is only set if this line selector is opened from the nav bar. Which is only in desktop mode.
+  const isMobile = value == null;
```
skaplan-dev, line 26: Can we use some `isMobile` hook rather than this? https://github.com/haldarmahesh/use-mobile-detect-hook/blob/master/src/index.ts

components/Dropdowns/LineSelector.tsx
```diff
@@ -1,85 +1,125 @@
...
+
+  // Don't render until we have the line.
+  if (!route.line) {
+    return <div></div>;
+  }
```
skaplan-dev, line 31: why do we need this? When `useDelimitatedRoute` resolves it should re-render and auto highlight the selected line right?

components/utils/router.tsx
```diff
@@ -20,3 +24,12 @@ export const useDelimitatedRoute = (): { line: Line; datapage: DataPage } => {
     datapage: pathItems[2] as DataPage,
   };
 };
+
+// If a datapage is selected, stay on that datapage. If the current line is selected, go to overview.
+export const getLineSelectionItemHref = (metadata: LineMetadata, route: Route): string => {
```
skaplan-dev, line 29: 🔥

components/widgets/TravelTimesWidget.tsx
```diff
@@ -81,7 +81,7 @@ export const TravelTimesWidget: React.FC = () => {
 
   return (
     <>
-      <HomescreenWidgetTitle title="Travel Times" href={`/${LINE_OBJECTS[route.line].path}/traveltimes`} />
+      <HomescreenWidgetTitle title="Travel Times" href={`/${LINE_OBJECTS[route.line]?.path}/traveltimes`} />
```
skaplan-dev, line 84: is there any point when we're on this page that we don't know the `route`?

skaplan-dev (approved):
Wasn't able to test it locally but looks good! 

just a few questions

components/Dropdowns/LineSelector.tsx
PatrickCleary: I added it to prevent this error which gets thrown in the console:
`component is changing from uncontrolled to controlled. This may be caused by the value changing from undefined to a defined value, which should not happen.`

My understanding is the Listbox component uses the `value` prop as state and so it can't be null on render. Maybe there is a better way to do this?

components/widgets/TravelTimesWidget.tsx
PatrickCleary: Once in a while it throws this error on load
Screen Shot 2023-02-01 at 11 45 48 AM

components/Dropdowns/LineSelector.tsx
skaplan-dev: hmm interesting. I'd say we do
```javascript
if(!route.line) return null
```

components/widgets/TravelTimesWidget.tsx
PatrickCleary: Oh wait. I think this was actually an issue with the URL. It keeps autocorrecting to Green instead of green.

Although we probably want to catch that and convert to lowercase, right?

components/Dropdowns/LineSelector.tsx
PatrickCleary: I could also do a div matching the size to prevent shifting after it loads?

PatrickCleary: Or pass a default lineMetadata with blank values

components/Dropdowns/LineSelector.tsx
PatrickCleary: So, I think what we want is for this to be a flag that matches the tailwind `sm` breakpoint. Since that's when we'll switch from bottom nav to top. I think tracking isMobile might not always align with that.

I _think_ the best option may be to track screen width and pass that value around as global state. Unless you know of something else?

PatrickCleary: This is the best thing I've found. What do you think?

https://stackoverflow.com/a/71098593

components/Dropdowns/LineSelector.tsx
skaplan-dev: i like the matching div size

components/Dropdowns/LineSelector.tsx
skaplan-dev: That looks good to me!

components/widgets/TravelTimesWidget.tsx
skaplan-dev: ya

skaplan-dev (approved):

components/widgets/TravelTimesWidget.tsx
PatrickCleary: Added an issue in the v4 project for this.

components/Dropdowns/LineSelector.tsx
PatrickCleary: I think I'm gonna split this into `LineSelectorMobile` and `LineSelectorNavBar`. It's a little messy with all the `isMobile` switches.

Also since the buttons depend on screen size, it doesn't work well with SSR. It will load either the mobile or navbar button and then switch once viewport size is known.

components/Dropdowns/LineSelector.tsx
PatrickCleary: I had actually added it, but I don't think it's necessary anymore. I actually have it in my other PR as well. Gonna remove it from this one though

PR #325, opened 2023-01-31
[V4 dev] Updating to new Datepicker

devinmatte:
Decided to find and use a real datepicker that will suffice for our needs

Screenshot 2023-01-31 at 3 48 59 PM

On mobile, we show the native datepicker just like we do on v3

Screenshot 2023-01-31 at 4 48 57 PM


constants/lines.ts
```diff
@@ -1,7 +1,7 @@
 import { COLORS } from './colors';
 
 export type Line = 'RL' | 'OL' | 'GL' | 'BL' | 'BUS';
-export type LinePath = 'red' | 'orange' | 'green' | 'bus';
+export type LinePath = 'red' | 'orange' | 'green' | 'blue' | 'bus';
```
PatrickCleary, line 4: Poor blue line, no one ever remembers it exists...

components/general/SecondaryNavBar.tsx
```diff
@@ -15,14 +15,47 @@ const visualizationOptions = [
...
+                asSingle={!range}
+                useRange={range}
+                showShortcuts={true}
+                containerClassName={'w-auto'}
+                inputClassName={'h-8'}
+              />
+            );
+          }}
+        </Device>
+        <Button text={range ? '🅧' : 'Range'} onClick={() => setRange(!range)} />
```
PatrickCleary, line 58: Is it possible to make this part of the date picker component? It looks like a separate filter at first glance

PatrickCleary (approved):
Looks great, just left a couple small comments.

components/general/SecondaryNavBar.tsx
```diff
@@ -15,14 +15,47 @@ const visualizationOptions = [
...
   return (
     <div className="pb-safe fixed bottom-11 z-20 w-full border border-gray-300 bg-white px-2">
       <div className="flex h-11 w-full flex-row items-center gap-x-2 bg-white">
-        <Dropdown
-          options={dateOptions}
-          setSelectedValue={() => null}
-          selectedValue={dateOptions[0]}
-        />
+        <Device>
+          {({ isMobile }) => {
```
PatrickCleary, line 31: Is this flag triggered on an iPad or tablet?

We probably want to use native datepicker in that case too right?

components/general/SecondaryNavBar.tsx
```diff
@@ -15,14 +15,47 @@ const visualizationOptions = [
...
+                  setDateSelection={({ startDate, endDate, range }) => {
+                    setRange(range);
+                    setDates({ startDate, endDate });
+                  }}
+                />
+              );
+            }
+            return (
+              <Datepicker
+                primaryColor={linePath !== 'bus' ? linePath : 'yellow'}
```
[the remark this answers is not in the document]
devinmatte: Doesn't seem like it from docs, we may be able to do with classnames, but at least not here using this prop

components/general/SecondaryNavBar.tsx
devinmatte: iPad is often treated like a smaller screen laptop by websites, and I think in this case as well. I could see it either way tbh

components/general/SecondaryNavBar.tsx
devinmatte: I can't add it to the library one, so we'd end up with two instances of this button, so this felt cleaner

PR #327, opened 2023-02-02
sidebar navigation

PatrickCleary:
* Add separate layout for desktop which includes sidebar navigation.
* Add  hook useBreakpoint to detect tailwind breakpoints.
* Disable SSR on _app.tsx, since it fully depends on viewport size. I _think_ SSR would still work in sub-components (?) Could use some weigh-in on this.
* I don't think we had been getting the full benefits of Headless UI accessibility. Using a `<Link>` component inside meant a user could not properly navigate using the `tab` key. I am going to update existing Headless components to match this new one. Also will probably change NavBar to use Headless <Tab> component.
* I added the `SecondaryNavBar` to desktop layout temporarily so we don't lose fuctionality. It looks terrible. Will fix soon.

Sidebar nav
Screen Shot 2023-02-02 at 4 23 32 PM

Sidebar nav with Travel Times focused.
Screen Shot 2023-02-02 at 4 24 05 PM

Secondary nav bar floating. Will fix soon.
Screen Shot 2023-02-02 at 4 26 14 PM


components/DashboardLayoutDesktop.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { SecondaryNavBar } from './general/SecondaryNavBar';
+import { SideNavBar } from './SideNavBar';
+import { DataPageHeader } from './widgets/DataPageHeader';
+import { WidgetPage } from './widgets/Widget';
+
+export const DashboardLayoutDesktop = ({ children }) => {
+  return (
+    <div className="flex h-full flex-row">
+      <SideNavBar />
+      <div className="flex h-full w-full flex-col items-center pl-40">
```
skaplan-dev, line 11: We should be able to have one `DashboardLayout` be smart enough to handle mobile / etc. You can add some classes on `SideNavBar` that will dynamically show/hide https://tailwindcss.com/docs/visibility . 

We can also use that new `useBreakpoint` endpoint any dynamically set the classes on the main `div`

skaplan-dev:
I think the navbar could be a little bit wider. It feels a little cramped at the moment. I like how roomy the tailwind components ones are https://tailwindui.com/components/application-ui/application-shells/sidebar

skaplan-dev (approved):

PR #330, opened 2023-02-03
Cleaning up UX and using real station data

devinmatte:
Screenshot 2023-02-03 at 4 25 31 PM

In an effort to clean things up, I am now using real station data to fetch the graphs, and, ensuring things are displayed properly, and cleaning up code/json

Screenshot 2023-02-03 at 4 34 15 PM


pages/[line]/traveltimes.tsx
```diff
@@ -3,48 +3,35 @@
...
-    station: 'place-dwnxg',
-    order: 9,
-    stops: {
-      '0': ['70078'],
-      '1': ['70077'],
-    },
-  };
+  const route = useDelimitatedRoute();
+
+  const stations = optionsStation(route.line);
```
Comment (skaplan-dev) [inline pages/[line]/traveltimes.tsx:17]: could this be a hook and then be
``` javascript
const {toStation, fromStation} = useStations(route.line)
```

skaplan-dev (commented):
LGTM except getting👇 at `[line]/slowzones`
Screen Shot 2023-02-03 at 4 54 55 PM

love that our types are getting tigher 
are you planning on adding functionality for the date picker in another PR?

PatrickCleary:
I think the chart and legend don't dynamically shrink when resizing the screen.
Screen Shot 2023-02-03 at 5 12 58 PM

pages/[line]/traveltimes.tsx
Comment (devinmatte) [reply pages/[line]/traveltimes.tsx:17]: It could, but right now I'm just populating dummy stations. We should probably start populating the url with real stations

devinmatte:
Shrinking this PR via https://github.com/transitmatters/t-performance-dash/pull/333, then will break up one more time

devinmatte:
Waiting on https://github.com/transitmatters/t-performance-dash/pull/340

modules/headways/HeadwaysWidget.tsx
```diff
@@ -47,12 +33,57 @@ export const HeadwaysWidget: React.FC = () => {
...
+  const longestHeadway = useMemo(() => {
+    if (headways && headways.data && headways.data.length >= 1) {
+      const allHeadways = headways?.data
+        .map((trip) => trip.headway_time_sec)
+        .filter((headway) => headway !== undefined) as number[];
+      return Math.max(...allHeadways);
+    } else {
+      return 0;
+    }
+  }, [headways]);
```
PatrickCleary, line 65: How do you feel about separating these from the component?

devinmatte:
Okay this PR is finally ready to test and I'm going to leave it alone

modules/traveltimes/TravelTimesWidget.tsx
```diff
@@ -70,33 +54,52 @@ export const TravelTimesWidget: React.FC = () => {
...
+              data={traveltimes.data ?? []}
+              date={startDate}
+              metricField={MetricFieldKeys.travelTimeSec}
+              pointField={PointFieldKeys.depDt}
+              benchmarkField={BenchmarkFieldKeys.benchmarkTravelTimeSec}
+              isLoading={isLoading}
+              bothStops={true}
+              location={location}
+              fname={'traveltimes'}
+              showLegend={!isMobile}
```
PatrickCleary, line 99: Maybe this should be decided by the `useBreakpoint` hook?

I think that's what we should use for screen size and use the `<Device>` component for stuff like choosing a date picker.

That way we can be certain what the screen will look like at each tailwind breakpoint.

As it is, if you shrink the screen to mobile size on desktop, the legend stays.

PatrickCleary (approved):
Thanks for improving the types, that's gonna make our lives a lot easier.

Left a couple small comments. Also still getting this behavior: https://github.com/transitmatters/t-performance-dash/pull/330#issuecomment-1416475682

modules/headways/HeadwaysWidget.tsx
devinmatte: Probably should, I can do that

modules/traveltimes/TravelTimesWidget.tsx
devinmatte: Ah I see. Yeah I suppose this could just be hidden for small screens

PR #333, opened 2023-02-04
Adding a dwells widget

devinmatte:
Adding a dwells widget to the home page

Shrinks https://github.com/transitmatters/t-performance-dash/pull/330 a little as a result

Screenshot 2023-02-04 at 4 28 28 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "concurrently": "^7.6.0",
     "date-fns": "^2.29.3",
     "dayjs": "^1.11.7",
+    "lodash": "^4.17.21",
     "mini-css-extract-plugin": "^2.7.2",
     "next": "13.1.6",
     "react": "18.2.0",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "concurrently": "^7.6.0",
     "date-fns": "^2.29.3",
     "dayjs": "^1.11.7",
+    "lodash": "^4.17.21",
```
skaplan-dev, line 19: Is it worth bringing this in just for `capitalize`?

skaplan-dev (approved):

package.json
devinmatte: There's been a few things I've wanted lodash for recently and figured it was worth adding now:
- `sample` selects a random value from an array
- `isEmpty` for checking easily if an object is empty
- `isEqual` for comparing two objects

If we want to avoid adding lodash that's fine, but, it is a pretty useful library to have around

PR #335, opened 2023-02-05
V4/redesign widgets

PatrickCleary:
Screen Shot 2023-02-05 at 2 17 54 PM


common/components/widgets/internal/BasicWidgetDataLayout.tsx
```diff
@@ -1,33 +1,62 @@
...
-  value?: string;
+  value: number;
   analysis: string;
-  units?: string;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  Icon: any;
+  delta?: number;
+  unit: WidgetUnit;
+  // If sentiment is true -> increase in stat = negative sentiment.
+  sentiment?: boolean;
```
devinmatte, line 12: I think this would be better served with a descriptive toggle

```
type Sentiment = 'positiveIncrease' | 'negativeIncrease';
```

Something like this, that way it's clear what each stands for

modules/headways/HeadwaysWidget.tsx
```diff
@@ -82,25 +82,17 @@ export const HeadwaysWidget: React.FC = () => {
...
             title="Average Headway"
-            value={
-              headways.data
-                ? secondsToMinutes(averageHeadway(headways.data)).toString()
-                : 'Loading...'
-            }
-            units="min"
-            analysis="+1.0 since last week"
-            Icon={<ArrowDownNegative className="h-3 w-auto" alt="Your Company" />}
+            value={headways.data ? averageHeadway(headways.data) : 0} // todo: send null and loads
```
devinmatte, line 85: Yeah I think we need a loading state as part of this component going forward. I want to avoid default values when possible especially for data that could load slowly

devinmatte (approved):
Looks good to me, glad to finally be able to start showing some delta data

devinmatte (commented):
Screenshot 2023-02-06 at 9 50 30 PM
Found a bug, doesn't handle hour+

common/types/basicWidgets.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type WidgetType = {
+  format: (value: number) => string;
+  description?: string;
+};
+export type WidgetFormat = Record<WidgetUnit, WidgetType>;
+
+export const WIDGET_UNITS: WidgetFormat = {
+  time: {
+    description: 'min.',
```
devinmatte, line 15: This should be able to be `min` | `sec` or `hour` depending on how big the data is.

I'm thinking:
- Sub 2 minutes, show in seconds
- Sub 1 hour, show in minutes + seconds
- Above 1 hour, show hours + minutes

PatrickCleary: We can assume the data will be passed to the widget in seconds right?

PatrickCleary: Also, wb sub 90 or 99 seconds show in seconds?

devinmatte: Yep data will be passed in seconds, and under 99 seconds sounds good for showing as seconds

common/components/graphics/LoadingSpinner.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
PatrickCleary, line 1: Found this spinner online. Don't love it, but figured we can just use it for now.

common/components/widgets/internal/BasicWidgetDataLayout.tsx
```diff
@@ -1,35 +1,64 @@
...
 export const BasicWidgetDataLayout: React.FC<BasicWidgetDataLayoutProps> = ({
   title,
-  value,
   analysis,
-  units,
-  Icon,
+  widgetValue,
+  sentimentDirection = 'negativeOnIncrease',
 }) => {
+  const getDelta = () => {
```
PatrickCleary, line 21: I acknowledge this function is gross

PatrickCleary:
Screen Shot 2023-02-07 at 11 54 34 AM
Screen Shot 2023-02-07 at 11 54 47 AM

common/components/graphics/LoadingSpinner.tsx
devinmatte: Was there a license or attribution we need to add?

public/Icons/ArrowUpPositive.svg
```diff
@@ -1,5 +0,0 @@
-<svg width="100%" height="100%" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
-<circle cx="8" cy="8" r="8" fill="#D9D9D9"/>
-<circle cx="8" cy="8" r="7.5" fill="#00FF00" fill-opacity="0.21" stroke="#00930F"/>
-<path d="M4.64645 10.6464C4.45118 10.8417 4.45118 11.1583 4.64645 11.3536C4.84171 11.5488 5.15829 11.5488 5.35355 11.3536L4.64645 10.6464ZM11.5 5C11.5 4.72386 11.2761 4.5 11 4.5L6.5 4.5C6.22386 4.5 6 4.72386 6 5C6 5.27614 6.22386 5.5 6.5 5.5L10.5 5.5L10.5 9.5C10.5 9.77614 10.7239 10 11 10C11.2761 10 11.5 9.77614 11.5 9.5L11.5 5ZM5.35355 11.3536L11.3536 5.35355L10.6464 4.64645L4.64645 10.6464L5.35355 11.3536Z" fill="#00930F"/>
-</svg>
```
devinmatte, line 5: On the topic of icons, do we need to license to attribute any of the other icons we're using? Could we instead use something like fontawesome: https://fontawesome.com/search?q=bus&o=r&m=free https://fontawesome.com/search?q=train&o=r&m=free
Which offers a free license tier that would suffice for us and has plenty of icons we'd need

devinmatte (approved):
Looks great, two questions that do not block this merging

PR #344, opened 2023-02-08
[v4] Real date selection

devinmatte:
Uses the real date from the selector and the url bar. Allows deep linking to a specific date

### DIFF
--- a/common/api/datadashboard.ts
+++ b/common/api/datadashboard.ts
@@ -9,13 +9,15 @@ import type {
 } from '../types/api';
 import { AggregateAPIParams, QueryNameKeys, SingleDayAPIParams } from '../types/api';
 import { APP_DATA_BASE_PATH } from '../../common/utils/constants';
+import { getCurrentDate } from '../utils/date';
 
 // Fetch data for all single day charts.
 export const fetchSingleDayData = (
   name: QueryNameKeys,
   options: PartialSingleDayAPIOptions
 ): Promise<SingleDayDataPoint[]> => {
-  const url = new URL(`${APP_DATA_BASE_PATH}/api/${name}/${options.date}`, window.location.origin);
+  const date = options.date ?? getCurrentDate();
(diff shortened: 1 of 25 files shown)


common/api/datadashboard.ts
```diff
@@ -108,6 +108,7 @@ export const useCustomQueries: UseQueriesOverload = (
             ? fetchAggregateData(name, queries[name].params)
             : fetchSingleDayData(name, queries[name].params),
         enabled,
+        staleTime: 30000, // Don't refetch within 30 seconds
```
devinmatte, line 111: This prevents spam to the API. We can likely bump this up to 1 minute with minimal to no user impact, but I decided to start out cautious

common/utils/date.ts
```diff
@@ -25,3 +25,11 @@ export const getCurrentDate = (): string => {
   const maxDate = localDate.toISOString().split('T')[0];
   return maxDate;
 };
+
+export const getOffsetDate = (date: string): string => {
+  const isoDate = new Date(date);
+  const offset = isoDate.getTimezoneOffset();
+  const localDate = new Date(isoDate.valueOf() + offset * 60 * 1000);
+  const offsetDate = localDate.toISOString().split('T')[0];
+  return offsetDate;
```
skaplan-dev, line 34: we should be able to use `dayjs` utils here

modules/traveltimes/TravelTimesWidget.tsx
```diff
@@ -31,14 +31,12 @@ export const TravelTimesWidget: React.FC = () => {
       [SingleDayAPIParams.fromStop]: fromStopIds || '',
       [SingleDayAPIParams.toStop]: toStopIds || '',
       [SingleDayAPIParams.stop]: fromStopIds || '',
-      [SingleDayAPIParams.date]: startDate,
+      [SingleDayAPIParams.date]: startDate ?? getCurrentDate(),
```
skaplan-dev, line 34: could we set this default date in the `hook` itself? we're repeating it a lot

skaplan-dev (approved):
just few nits

PR #345, opened 2023-02-08
Add Slow Zone data

skaplan-dev:
## This PR
- Adds in some real calculations for the SZ widgets
Screen Shot 2023-02-07 at 9 03 53 PM


devinmatte (approved):
Screenshot 2023-02-08 at 2 36 22 PM
I'm only able to get the slow zones to display 0. Outside of that it seems good

PR #353, opened 2023-02-10
Move prod backend from dashboard-api2.transitmatters.org to dashboard-api.labs.transitmatters.org

mathcolo:
Now that DNS for *.labs.transitmatters.org is hooked up, we can make subdomains on that as easy as pie! As a result, let's move the production Data Dashboard backend's hostname to be at dashboard-api.labs.transitmatters.org, which has the following benefits:

1. One fewer environment variable; no more TM_BACKEND_CERT_ARN!
2. One fewer SSL cert in our account (they're free, but still).


server/frontend-cfn.json
```diff
@@ -24,8 +24,8 @@
     },
     "TMBackendHostname": {
       "Type": "String",
-      "Default": "dashboard-api.labs.transitmatters.org",
```
PatrickCleary, line 27: Why was this already the labs subdomain?

austinjpaul:
We need to sync up with the slowzones repo so we don't break things again.

server/frontend-cfn.json
mathcolo: The default value wasn't actually being used—it was really more for spiritual purposes

PR #357, opened 2023-02-11
Style changes

PatrickCleary:
Fix bug with seconds units on widgets

UI Changes:
1. two column grid at screen size xl
2. homescreen widgets are line-colored
3. Change chartjs labels to be the grey we use for widgets
4. Move widget titles inside widgets and make line-colored

Screen Shot 2023-02-14 at 12 13 33 PM


common/components/charts/SingleDayLineChart.tsx
```diff
@@ -99,9 +101,17 @@ export const SingleDayLineChart: React.FC<SingleDayLineProps> = ({
...
+                    ? LINE_COLORS[line]
+                    : pointColors(data, metricField, benchmarkField),
+                pointBorderWidth: homescreen ? 0 : undefined,
                 pointHoverRadius: 3,
-                pointHoverBackgroundColor: pointColors(data, metricField, benchmarkField),
+                pointHoverBackgroundColor:
+                  homescreen && line
+                    ? LINE_COLORS[line]
+                    : pointColors(data, metricField, benchmarkField),
                 pointRadius: 3,
```
devinmatte, line 115: Love the color changes, I wonder if it wouldn't also be nice to get rid of the points on the homepage like we do for slowzones. Since they're not color coded the red/orange dots feel noisy

common/components/inputs/DateSelector.tsx
```diff
@@ -45,7 +45,9 @@ export const DateSelector: React.FC<DateSelectorProps> = ({ range }) => {
   return (
     <Datepicker
       primaryColor={linePath !== 'bus' ? linePath : 'yellow'}
-      value={dates}
+      value={
+        dates ?? { startDate: dayjs().format('YYYY-MM-DD'), endDate: dayjs().format('YYYY-MM-DD') }
```
devinmatte, line 49: Is this meant to default to today? Doesn't seem to work

modules/ridership/RidershipWidget.tsx
```diff
@@ -28,47 +28,45 @@ export const RidershipWidget: React.FC = () => {
...
+            new PercentageWidgetValue(
+              ridershipPercentage[ridershipPercentage.length - 1],
+              ridershipPercentage[ridershipPercentage.length - 1] -
+                ridershipPercentage[ridershipPercentage.length - 31]
+            )
+          }
+          analysis={`from last month.`}
+          sentimentDirection={'positiveOnIncrease'}
+        />
+        <BasicWidgetDataLayout
```
devinmatte, line 57: These should probably be above like we do for the other widgets

devinmatte (approved):

common/components/charts/SingleDayLineChart.tsx
PatrickCleary: This made me think... does it even make sene to have them connected? There isn't really a meaning behind the slope of two points. Any relationship in the date is clear from the change in location of groups of dots over time.

The line between dots almost gives someone the impression that it is one continuous datapoint which is changing values when in reality it's a bunch of separate trips.

Screen Shot 2023-02-14 at 2 37 15 PM

common/components/inputs/DateSelector.tsx
PatrickCleary: Ah,  yea I was trying to do that and then gave up lol. Musta forgot to revert this piece.

modules/ridership/RidershipWidget.tsx
PatrickCleary: whoopsy

common/components/charts/SingleDayLineChart.tsx
PatrickCleary: On Dwells - I think it actually makes it easier to see the outliers

Screen Shot 2023-02-14 at 2 43 09 PM

PatrickCleary: And it makes it easier for the user to hover the point they want - since we can use a hover radius rather than horizontal location

devinmatte: I dislike it without the lines much more than just keeping the dots as a scatter. I'd much rather have it how it was before than this way. If we want to highlight outliers I think there's a better way we can do that

devinmatte: I do see your point about whether or not they're connected, but I think it still looks better with the lines

PatrickCleary: Fair enough. fwiw, I think headways in particular would benefit from being represented differently. When I look at this graph I see this up and down motion. But I think the key insight from this graph is that the headways are super inconsistent. Headways might lend themselves to a different type of visualization altogether. Maybe a histogram with 30 sec increments on x-axis.

Screen Shot 2023-02-14 at 4 06 53 PM

PatrickCleary: Not suggesting we remove this - just another visualization.

devinmatte: Now that we have a dedicated headways page, this is a great time to play with different visulizations. Maybe @austinjpaul has some he's thought about showing before. As for the main view, either the one you started with, or your most recent screenshot style are my preferences

austinjpaul: I prefer the dots with no lines. Or, maybe faint lines? We do need to highlight that these are data points. The faint lines help show order (esp on bimodal distributed headways). The solid line above gives equal weight to the in-between stuff and we don't want that

PatrickCleary:
Screen Shot 2023-02-16 at 10 08 25 AM

skaplan-dev (approved):

PR #358, opened 2023-02-12
Ridership and service levels

devinmatte:
Screenshot 2023-02-12 at 3 46 20 PM

Adding ridership and service levels

We can start to make inferences about this data without comparing it to 2020 levels, and still also compare to 2020 levels


PatrickCleary:
Looks great. Haven't looked at code yet but one tiny nit:
Axis units get cut off at screen width > 1328px
Screen Shot 2023-02-13 at 1 27 29 PM

PatrickCleary:
Also, not a blocker on merging, but we might need more context on the widgets to explain what "Service Levels" is since the data isn't on a daily basis (I think).

Might require some redesign of the widgets. I've been working on adjusting some of the homepage stuff so I can think about this.

devinmatte:
Yeah I think we might need to add some copy to the views that explain terms and why certain data points matter. Also the date selection with this data is rough so not sure the best way to make it work

PatrickCleary (approved):

PR #362, opened 2023-02-14
Add 10th and 90th percentiles to backend

austinjpaul:
I realize this change would ideally wait for v4, but I want to start populating an aggregate database and it would be very useful to have these numbers in it as i load it up.


friendchris (approved):
lgtm

PR #363, opened 2023-02-15
[v4] Got basic aggregates working

devinmatte:
Screenshot 2023-02-14 at 10 32 58 PM

Got the bare bones basics for aggregates working

did some work to get endDate to be treated right (also some work to prevent `line` from being unset for a second at first load)


PatrickCleary (approved):
A nitpick  with the datepicker that I just noticed: If the user is in "single day" mode then the last _x_ days selector won't choose a range, it only chooses the first date in that range. The reverse is also true - if you're looking at aggregates and select "today" or "yesterday" it makes a one day range.

Assuming this isn't related to these changes I'm fine to just make a new issue in GH for this, doesn't need to be part of this imo.

PR #364, opened 2023-02-16
V4/headways histogram

PatrickCleary:
Histogram!

(And I moved some types around)

Screen Shot 2023-02-16 at 4 42 58 PM


modules/headways/charts/HeadwaysHistogram.tsx
```diff
@@ -0,0 +1,173 @@
...
+    let max = 0.5;
+    return (
+      headways.data?.reduce(
+        (datapointMap: Record<string, number>, datapoint: HeadwayPoint) => {
+          const bucketString = (
+            Math.floor(datapoint[MetricFieldKeys.headwayTimeSec] / 60) + 0.5
+          ).toString();
+
+          // If the bucket doesn't exist, create all buckets leading up to it and set them to 0.
+          while (datapointMap[bucketString] == null) {
```
skaplan-dev, line 42: I think a more clearer way of doing this would be 

``` javascript
// get all formatted values from the data
const values  = headsays.data.map((headway => getString(headway))

// Get max and min numbers
const max = Math.max(...values)
const min = Math.min(...values)

// Create an obj with keys between min + max with value zero

// Find occurrences in headways.data and store in obj created above
```

common/types/charts.ts
```diff
@@ -59,3 +62,58 @@ export enum BenchmarkFieldKeys {
...
+  metricField: MetricField;
+  date: string | undefined;
+  benchmarkField?: BenchmarkField;
+}
+
+export interface HeadwayHistogramProps {
+  title: string;
+  chartId: string;
+  data: SingleDayDataPoint[];
+  date: string | undefined;
```
skaplan-dev, line 104: you should be able to make this just optional instead of doing ` | undefined ` no?

modules/headways/charts/HeadwaysHistogram.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const ref = useRef();
+
+  // dataObject is a mapping from headway bucket -> number of trains.
+  // All keys are increased by 0.5. This is a workaround to get chartjs to display the tick labels in between the bars.
+  const dataObject: Record<string, number> = useMemo(() => {
+    let max = 0.5;
+    return (
+      headways.data?.reduce(
+        (datapointMap: Record<string, number>, datapoint: HeadwayPoint) => {
```
skaplan-dev, line 36: we shouldn't have to type the `datapoint` here right? that might mean the `headways` type isn't tight enough. 

you may have to set the type on the query?

doing `.filter / map / reduce / etc` on a typed array should type the individual values properly automatically

skaplan-dev (commented):
Lookin good so far! Few comments on code cleanliness / cleanup

modules/headways/charts/HeadwaysHistogram.tsx
PatrickCleary: I think the issue is all of our single day queries return the same type, `SingleDayDataPoint` which can be a dwell, headway, or travel time.

And that is because I wrote this code when we had the original setup which was identical to v3, where the single day queries are all on the same page. I think we'll probably want to redo that. No reason to get traveltimes if someone is on the dwell page.

I can make a new issue for this. I think this will solve both issues at once.

modules/headways/charts/HeadwaysHistogram.tsx
PatrickCleary: Yea....

In hindsight, idk what I was thinking lol. Lost in the sauce.

common/types/charts.ts
devinmatte: So if you do `?` typescript won't make you even set a variable to the prop. If you do `| undefined` you'll need to pass a variable to the prop, but, the variable value can be undefined. So there's value in both approaches.

In this case, it yelling at you if you don't even attempt to give it a date object is probably good

modules/headways/charts/HeadwaysHistogram.tsx
devinmatte: Yeah the direct v3 translation has led to some loose typings. We should definitely clean this up to allow 3 different types, 3 different API calls.

This may be a bigger project so I think we should make this an issue for longer term tracking

modules/headways/charts/HeadwaysHistogram.tsx
PatrickCleary: I'd like to keep it so it always starts at 0 though if that's cool with you. It's typically only like 3-4 extra columns and it gives you a better sense of how bad some lines are.

modules/headways/charts/HeadwaysHistogram.tsx
```diff
@@ -0,0 +1,167 @@
...
+    linePath,
+    lineShort,
+    query: { startDate },
+  } = useDelimitatedRoute();
+
+  const ref = useRef();
+
+  // dataObject is a mapping from headway bucket -> number of trains.
+  // All keys are increased by 0.5. This is a workaround to get chartjs to display the tick labels in between the bars.
+  const dataObject: Record<string, number> = useMemo(() => {
```
skaplan-dev, line 32: waaay cleaner 🔥

skaplan-dev (approved):

modules/headways/charts/HeadwaysHistogram.tsx
PatrickCleary: ✅ https://github.com/orgs/transitmatters/projects/5?pane=issue&itemId=20789116

PR #367, opened 2023-02-17
V4/fix links

PatrickCleary:
Our links were not working in v4. This was caused by the static export not generating files for our dynamic routes like `/[line]/`

There was also an issue caused by the cloudfront distribution. We needed the "website" URL `.s3-website-us-east-1.amazonaws.com` for our S3 bucket rather than just `s3`.

Last change was adding `trailingSlash:true` to the next config. My understanding is that allows the server to know a url ending in `/green` should become '/green/' which is a folder containing an `index.html' file.

Would've liked to apply `getStaticPaths` and `getStaticProps` universally to all pages inside the `[line]` folder, but don't think it's possible. I also didn't include bus routes for dwells and slow zones.

@mathcolo Pretty sure there was no good way of getting `.s3-website-us-east-1.amazonaws.com` from cloudformation other than directly appending it. Closest thing is `WebsiteURL` but that includes `https`. I also don't know if we want/need a more general setup for other regions. 

Tested by running `deploy.sh` locally. Successfully created the correct Cloudfront distribution.

### DIFF
--- a/common/types/lines.ts
+++ b/common/types/lines.ts
@@ -1,14 +1,3 @@
-import type { TimeUnit } from 'chart.js';
-import type React from 'react';
-import type {
-  AggregateDataPoint,
-  BenchmarkField,
-  MetricField,
-  PointField,
-  SingleDayDataPoint,
-  Location,
-} from './charts';
-
 export type Line = 'RL' | 'OL' | 'GL' | 'BL' | 'BUS';
 export type LineShort = 'Red' | 'Orange' | 'Green' | 'Blue' | 'Bus';
 export type LinePath = 'red' | 'orange' | 'green' | 'blue' | 'bus';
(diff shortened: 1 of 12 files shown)


devinmatte (approved):

common/types/lines.ts
```diff
@@ -54,3 +54,17 @@ export interface SingleDayLineProps extends LineProps {
   date: string | undefined;
   benchmarkField?: BenchmarkField;
 }
+
+export const ALL_LINE_PATHS = ['red', 'orange', 'green', 'blue'].map((line) => {
```
skaplan-dev, line 58: can we use an existing `const` here for the line names? I swear we have like 6 different versions 🤣

skaplan-dev (approved):

common/types/lines.ts
PatrickCleary: Yea I see what you mean. We have the LinePath type, but it's just a type and includes `bus` which we don't want here. I don't think there's any other const which fills this need. I agree there seems to be a lot of random line objects all over the place, and we switch between indexing on keys (`RL`) and paths (`red`). Could use some clean up, but might be better as a new PR.

PR #373, opened 2023-02-21
update bus data for December 2022

austinjpaul:
At long last!
There's a simple change to the ingestion script as well because the input data format has changed yet again -.-


server/bus/bus2train.py
```diff
@@ -58,9 +58,9 @@ def load_data(input_csv, routes):
     df.stop_id = df.stop_id.astype(str)
 
     # Convert dates
-    df.scheduled = pd.to_datetime(df.scheduled)
+    df.scheduled = pd.to_datetime(df.scheduled).dt.tz_localize(None)
```
mathcolo, line 61: This results in local time, right? Wouldn't `America/New_York` be preferable to that, as we don't want OS timezone to mess us up?

austinjpaul: Not quite - this removes the timezone information in order to create timezone-naive datetimes. All the past data inputs had datetime naive inputs ('1900-01-01T10:30:00') already but the latest csv has ('1900-01-01T10:30:00Z'), which is being interpreted as GMT+0. Definitely don't want that to be converted to local time, so we just ditch the timezone.

friendchris (approved):
thanks for adjusting again

PR #376, opened 2023-03-03
[v4] Adding a basic station selector to travel times

devinmatte:
Screenshot 2023-03-03 at 6 36 15 PM

Adds a basic station selector using similar logic to v3.

Adds new metadata about accessibility and bike storage (because it felt like we should call out the stations that are/aren't accessible and don't provide multimodal support)

Screenshot 2023-03-23 at 12 33 09 PM

This calls out Green line stops very clearly as not being accessible


scripts/update_stations.py
PatrickCleary, line 1: Would like to get input from @austinjpaul on this:

I think we probably want one source of truth on stations. So we might want this to update the stations.json file directly?

And idk where the stations.json file gets generated or if it is our own manual creation, but if a script exists to generate it, we probably want this to be part of that. If it's made manually, then maybe we should put it in a repo somewhere along with this script.

PatrickCleary (commented):
Looks good. Do you mind adding a `swap` button like v3 has? Also v3 removes the currently selected `from` stop from the `to` dropdown and vice versa. I was thinking we could keep it there but `disable` it with the listBox disable option. What do you think?

scripts/update_stations.py
PatrickCleary: Just talked to Austin, he said we should just replace the stations.json file with this new one. And we can keep the script in here to run manually when things change?

Does the script work if it's updating the current file rather than creating a new one?

devinmatte:
So I use the same station listing logic as v3 and I personally like that the selected station is simply not available. It's also true that if I select a to go from a station on the E branch it won't list anything on the C branch in the to field, and I think that's good. Maybe showing the selected one disabled would be okay, but I think it's fine as is

I think we could do a lot more UX wise for the selector that I'm not design minded enough to quite figure out

devinmatte:
As for the swap button, I'll get to that today, should be easy enough

PatrickCleary:
I think the logic on the old dashboard is a little different.

For example if Malden Center is selected for `from` then it doesn't show up in the `to` dropdown.

In this change, if Malden center is selected for `from` then it still shows up in `to`, but not in `from`

devinmatte:
Hmmm that's a bug. I literally copy pasted code from v3 so I'm not sure how that happened 😕

scripts/update_stations.py
devinmatte: Yes, we ran the script on the existing file, and it simply added the new fields to the existing station entries

PatrickCleary: Ah, I was looking at the V3 stations.json and thought you were updating that one, nevermind.

devinmatte:
Screenshot 2023-03-25 at 3 30 25 PM
Bug fixed, swpa button added, and cleaned up my UI a bit (still feel free to do more later)

scripts/update_stations.py
devinmatte: We are? It's the same file as was in the repo since v3, just, adding new `accessible` and `enclosed_bike_parking` and adding the GLX entries we had never pulled into the v4 branch before from v3

PatrickCleary (approved):

PR #378, opened 2023-03-06
Default date selector to today

PatrickCleary:
Set default date to be current date.

Three notes:
1) I didn't put the date in the URL. So if someone shares the link it will always be today. Makes sense to me but idk. Maybe if the user is on a data page then we should add the date? Just not on overview?

2) I think the whole router situation needs some work. Lots of type weirdness. I started getting into it then undid my changes because it was getting gross.

3) Small UI change: date format YYYY-MM-DD, changed to M/D/YY. Still need to update on mobile.

Screen Shot 2023-03-06 at 6 05 56 PM

### DIFF
--- a/common/components/inputs/DateSelector.tsx
+++ b/common/components/inputs/DateSelector.tsx
@@ -15,24 +15,26 @@ export const DateSelector: React.FC<DateSelectorProps> = ({ range }) => {
   const updateQueryParams = useUpdateQuery({ range });
   const maxDate = getCurrentDate();
 
-  const dates = useMemo(
-    () => ({
-      startDate: startDate ?? null,
-      endDate: endDate ?? startDate ?? null,
-    }),
-    [startDate, endDate]
-  );
+  const dates = useMemo(() => {
+    return startDate
+      ? {
+          startDate: startDate,
+          endDate: endDate ?? startDate,
+        }
+      : null;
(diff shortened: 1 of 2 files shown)


common/components/inputs/DateSelector.tsx
```diff
@@ -15,24 +15,26 @@ export const DateSelector: React.FC<DateSelectorProps> = ({ range }) => {
...
       primaryColor={linePath !== 'bus' ? linePath : 'yellow'}
       value={dates}
+      placeholder="Date"
       onChange={(evt) => updateQueryParams(evt)}
       maxDate={maxDate}
       asSingle={!range}
       useRange={range}
       showShortcuts={true}
-      displayFormat={'YYYY-MM-DD'}
+      displayFormat={'M/D/YY'}
```
devinmatte, line 37: If we're going to this, I personally would prefer MM/DD/YYYY since it's a more standard format in single digit months and dates

PatrickCleary: What about this?
Most human-readable imo
Screen Shot 2023-03-07 at 6 28 28 PM

devinmatte: Yeah since it's just a display format

skaplan-dev (approved):

devinmatte (approved):

PR #379, opened 2023-03-08
Add API to v3 and make simple widget to display shuttling info

PatrickCleary:
UI needs some work but wanted to set it up so we can start seeing alerts and get a feel for how this will work

Two times of alerts:
- Current (Will be active today)
- Upcoming (Will be active anytime after tomorrow at 4:00 AM)

One alert can be both current and upcoming - if it is active now but also will span a few days.
Current shows times of day and upcoming shows dates.

Due to inconsistencies with the way MBTA uploads this info, sometimes there will be a little funkiness. May have some kinks to work out but I think best way to do that is get this in and that check up on it as we experience alerts and make sure it is working as expected.

Overall:
Screen Shot 2023-03-10 at 5 35 34 PM

Opened upcoming
Screen Shot 2023-03-10 at 5 39 36 PM

Click to expand and see header
Screen Shot 2023-03-10 at 5 41 50 PM


common/api/datadashboard.ts
```diff
@@ -49,12 +53,12 @@ export const fetchAggregateData = (
       url.searchParams.append(key, value.toString());
     }
   });
-  return fetch(url.toString())
-    .then((resp) => resp.json())
-    .then((resp) => {
-      // traveltimes API returns data under two fields: by_date and by_time. This formats the other two APIs to be the same shape.
-      return name === QueryNameKeys.traveltimes ? resp : { by_date: resp };
-    });
+  const response = await fetch(url.toString());
```
skaplan-dev, line 56: couldnt we do a `.catch` here instead of making the func async?

skaplan-dev, line 56: couldn't we do a `.catch()` here and avoid making this async?

modules/commute/alerts/Alerts.tsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+const AlertBox: React.FC<AlertBoxProps> = ({ alerts, lineShort, line, type }) => {
+  const relevantAlerts: FormattedAlert[] = [];
+  alerts.forEach((alert) => {
+    const relevantTimes = alert.active_period.filter((period) => period[type] === true);
+    if (relevantTimes.length > 0) {
+      relevantAlerts.push({ ...alert, relevantTimes: relevantTimes });
+    }
+  });
```
skaplan-dev, line 42: could we `useMemo` this? also might be cleaner with a `.reduce` or `.map` instead of the for loop

modules/commute/alerts/styles/AlertStyles.ts
```diff
@@ -0,0 +1,17 @@
+export const lightLineBorder = {
+  RL: `bg-mbta-red border border-mbta-lightRed`,
+  OL: `bg-mbta-orange border border-mbta-lightOrange`,
+  GL: `bg-mbta-green border border-mbta-lightGreen`,
+  BL: `bg-mbta-blue border border-mbta-lightBlue`,
+  BUS: `bg-mbta-bus border border-mbta-lightBus`,
+  DEFAULT: `bg-gray-400 border border-white`,
+};
+
+export const lineBackground = {
```
skaplan-dev, line 10: do we have these somewhere? i feel like we have a bunch of versions of these exact styles

skaplan-dev (commented):
few comments!

we'll probably have to attribute those icons you added no?

common/api/datadashboard.ts
PatrickCleary: What is the reason to keep fetchAggregateData sync? It would function the same, right?

Also the reason I didn't use `.catch()` is `fetch()` doesn't throw on 400/500 HTTP errors (but could still do what you suggest with a `.then()` of course).

modules/commute/alerts/Alerts.tsx
PatrickCleary: will do

modules/commute/alerts/styles/AlertStyles.ts
PatrickCleary: I'll add a general styles file

PatrickCleary:
wrt attribution for icons - I have that in a to do list before launch. Don't love the ones I used so might just wait until we're sure which library we want to use.

PatrickCleary:
Just refactored a bunch. Still needs work probably.

skaplan-dev (approved):
LGTM!

PR #383, opened 2023-03-11
Adding datadog rum

devinmatte:
As we get more and more traffic, might be nice to have a more detailed look at usage data

https://app.datadoghq.com/rum/application/1c18c845-9d82-40ba-b4bb-066ca95612be/overview/browser

Someone with admin perms, should add the two new secrets to github for deployment

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+REACT_APP_VERSION=$npm_package_version
+REACT_APP_NAME=$npm_package_name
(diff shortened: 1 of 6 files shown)


nathan-weinberg:
@devinmatte I can add those secrets if you can point out to me where I can find the values - assume these need to be org-wide secrets and not just for this repo?

nathan-weinberg (approved):
Secrets are added :+1:

PR #384, opened 2023-03-14
V4/sidebar updates

PatrickCleary:
1. Fix keyboard nav for sidebar (doesn't work with `Disclosure` so I changed to `Listbox`)
2. Add bus navigation - but the queries still don't work.
3. Unfortunately no good way to keep the dropdown open when the group is selected. There may be a work around if we think it is important.
4. Some UI changes. Not perfect, open to suggestions but didn't want to spend too much time.
5. Added alerts for buses
6. Added modal for alerts

Screen Shot 2023-03-13 at 7 55 32 PM
Screen Shot 2023-03-13 at 7 55 40 PM

Screen Shot 2023-03-16 at 2 40 01 PM

Screen Shot 2023-03-16 at 3 13 08 PM


common/utils/router.tsx
```diff
@@ -21,18 +22,26 @@ export const useDelimitatedRoute = (): Route => {
...
     datapage: (pathItems[2] as DataPage) || 'overview', //TODO: Remove as
+    tab: tab,
     query: router.isReady
       ? {
           startDate: Array.isArray(startDate)
             ? startDate[0]
             : startDate ?? dayjs().format('YYYY-MM-DD'),
           endDate: Array.isArray(endDate) ? endDate[0] : endDate,
-          busLine: Array.isArray(busLine) ? busLine[0] : busLine ?? '22', // TODO: Remove default bus
+          busLine: (Array.isArray(busLine) ? busLine[0] : busLine) as BusRoute,
```
skaplan-dev, line 44: if we type `router.query` can we avoid this cast?

common/utils/router.tsx
```diff
@@ -76,22 +85,37 @@ export const useUpdateQuery = ({ range }: { range: boolean }) => {
...
-    : '';
-  let href = `/${metadata.path}`;
-  if (metadata.key !== line && datapage) {
-    if (datapage !== 'overview') {
-      href += `/${datapage}`;
-    }
-    if (queryParams !== '') {
-      href += `?${queryParams}`;
-    }
+        Object.entries(query).filter(([key, value]) => value !== undefined && key !== 'busLine')
```
skaplan-dev, line 98: What is the case that a query param would be undefined?

modules/navigation/desktop/SideNavBar.tsx
```diff
@@ -1,33 +1,38 @@
...
+        { name: 'Blue', path: 'blue', key: 'BL' },
+        { name: 'Green', path: 'green', key: 'GL' },
       ],
     },
+    {
+      name: 'Bus',
+      current: false,
+      icon: faBus,
+      key: 'Routes',
+      children: getBusLines().map((busRoute) => {
```
skaplan-dev, line 32: I wonder if I like having all of the routes in the side nav. I think a dropdown (consistent with our stop picker) on desktop and a bottom nav selector on mobile might be cleaner. But I think we should talk about it with the group!

skaplan-dev (commented):
overall looks good. few comments and questions! 
will approve once those are addressed :)

modules/navigation/desktop/SideNavBar.tsx
PatrickCleary: I'm fine doing it either way. Are you cool with keeping it for now? Gonna work on bus queries next so I need a way to switch between them.

PR #395, opened 2023-03-24
Adding labeler action

devinmatte:
Adding an action to automatically label all PRs with useful labels based on what was changed

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,11 +1,10 @@
 name: Labeler
+
 on: [pull_request]
 
 jobs:
   label:
-
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/labeler@v4
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/label.yml
```diff
@@ -0,0 +1,13 @@
+name: Labeler
+on: [pull_request]
+
+jobs:
+  label:
+
+    runs-on: ubuntu-latest
+
```
nathan-weinberg, line 8: Going to be annoying and ask you to remove the empty lines here but otherwise I say we merge and try it out. V4 is going to be sort of a proving-ground for org-wide changes to come later.

nathan-weinberg (approved):
LGTM - @devinmatte can you make a note of this PR in our GitHub doc? I think we want to try a lot of the new stuff here in the Data Dashboard and then if we like the workflow I'll likely mimick these configurations in our other app repos

PR #396, opened 2023-03-24
[v4] Removing v3 code

devinmatte:
If needed, reference existing main branch

I figured we don't need to keep dead code in the branch anymore


PatrickCleary:
When we launch, if (god forbid) things go horribly wrong - do we want to be able to swap back to v3 quickly? I'm guessing that would be a matter of pointing the domain to an old release, in which case we're good to remove this code. Right?

PatrickCleary (approved):

devinmatte:
Yeah I think the day before we merge v4 into main, we should tag a release as the last v3 version, and have it available at like an `old.dashboard` style URL for a few months

PR #399, opened 2023-03-25
server changes to add speed API

PatrickCleary:
Add /api/speed route

Small fix to my old alerts code

Autoformatter redid some json files. Can undo if we don't want that.


skaplan-dev (approved):

server/chalicelib/dynamo.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def query_speed_tables(params):
+    config = AGG_TO_CONFIG_MAP[params["agg"]]
+    table = dynamodb.Table(config["table_name"])
+
+    start_datetime = datetime.strptime(params["start_date"], DATE_FORMAT_BACKEND)
+    end_datetime = datetime.strptime(params["end_date"], DATE_FORMAT_BACKEND)
+    if start_datetime + timedelta(days=config["delta"]) < end_datetime:
+        return {"Error": "Invalid Query - too many items requested."}
```
austinjpaul, line 23: I know we don't do this currently, but I would love for input validation to be done in app.py, where we can raise a 4xx status code (Chalice has some builtins to help here I think) instead of a 5xx when the user input is wrong.

server/chalicelib/dynamo.py
```diff
@@ -0,0 +1,39 @@
...
+            '#date': 'date'
+        },
+        'ExpressionAttributeValues': {
+            ':pk': params["line"],
+            ':start_date': params["start_date"],
+            ':end_date': params["end_date"]
+        }
+    }
+
+    response = table.query(**query_params)
```
austinjpaul, line 38: I'm partial to replacing this with the (IMO) more readable form:
```
from boto3.dynamodb.conditions import Key
...
response = table.query(KeyConditionExpression=Key("line").eq(line) & Key("date").between(start_date, end_date))
```

server/chalicelib/dynamo.py
```diff
@@ -0,0 +1,39 @@
+import boto3
+from datetime import datetime, timedelta
+
+# Create a DynamoDB resource
+dynamodb = boto3.resource('dynamodb')
+
+DATE_FORMAT_BACKEND = "%Y-%m-%d"
+
+AGG_TO_CONFIG_MAP = {
```
austinjpaul, line 9: A comment explaining the delta here wouldn't be amiss.

server/chalicelib/dynamo.py
```diff
@@ -0,0 +1,39 @@
...
+    "daily": {"table_name": "DailySpeed", "delta": 150},
+    "weekly": {"table_name": "WeeklySpeed", "delta": 7 * 150},
+    "monthly": {"table_name": "MonthlySpeed", "delta": 30 * 150},
+}
+
+
+def query_speed_tables(params):
```
austinjpaul, line 16: In conjunction with validating params in app.py (see below), this function should take distinct arguments (line, start_date, end_date). 
(Right now, user can specify secret params that get passed straight through to `table.query()`. not good!)

server/chalicelib/dynamo.py
austinjpaul: "here", well this dict should maybe move inside get_speed() in app.py with the rest of validation.

austinjpaul (changes requested):
We need to be a little more careful with the params users can pass to us, but otherwise looks great!

server/chalicelib/dynamo.py
austinjpaul: oh, and `table` i suppose. Someday if we add a bunch more tables we might want to separate out calls to each, but this is good for now.

austinjpaul (approved):
looks great, thanks!!

PR #402, opened 2023-03-25
[v4] Adding better pickers on desktop and mobile

### LINKED ISSUE #369: Date Picker fix
If the user is in "single day" mode then the last _x_ days selector won't choose a range, it only chooses the first date in that range. The reverse is also true - if you're looking at aggregates and select "today" or "yesterday" it makes a one day range.

devinmatte:
Cleaning up the pickers and fixing some bugs with the datepicker, and mobile views

Screenshot 2023-03-25 at 5 45 37 PM
Screenshot 2023-03-25 at 5 45 59 PM

Fixes #369

### DIFF
--- a/common/components/inputs/Button.tsx
+++ b/common/components/inputs/Button.tsx
@@ -9,17 +9,19 @@ interface ButtonProps
     HTMLButtonElement
   > {
   content: string | React.ReactNode;
+  isFullWidth?: boolean;
 }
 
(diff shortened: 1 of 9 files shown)


devinmatte:
Given #403 some of the changes in here are temporary, but, I still think they're worth getting in in the meantime

skaplan-dev (approved):

PR #405, opened 2023-03-27
Aggregate data for Headways and Dwells

devinmatte:
Aggregate data for the two remaining pages


devinmatte:
Accidentally pushed fixes to the accessibility data to this branch, but figured it's small enough to not break out into a new PR unless someone disagrees

PR #407, opened 2023-03-27
Adding OSS licenses page

devinmatte:
Found a tool https://github.com/zumwald/oss-attribution-generator that generates a json blob of all licenses in our package.json that we use

Taking that json, we can display a list of all licenses of used software. This only applies to package.json, if we need to list python side libraries we may need to manually add those

Screenshot 2023-03-26 at 10 28 10 PM

Issue, need to display a page without the header


PatrickCleary (approved):
Looks good to me. Just pushed a small change to float the footer to the bottom of the page since I already had it open, is that cool with you?

PR #409, opened 2023-03-27
Cleaning up the Bus

devinmatte:
Better bus viewing

- Hiding unused data pages for buses
- Fixing broken station selectors
- Fixing broken ridership stats
- Adding bus data notice


common/components/notices/BusDataNotice.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { useDelimitatedRoute } from '../../utils/router';
+
+export const BusDataNotice: React.FC = () => {
+  const { line, linePath } = useDelimitatedRoute();
+
+  if (line === 'BUS' || linePath === 'bus') {
```
PatrickCleary, line 7: These will always be the same right?

PatrickCleary (approved):

common/components/notices/BusDataNotice.tsx
devinmatte: Sometimes `line` is undefined so I figured I'd add redundancy. Not strictly necessary though

PR #410, opened 2023-03-28
V4/date selector

PatrickCleary:
Add a new date selector. It still could use some work, but it works well and I don't think the effort/reward is worthwhile at this point.

It doesn't fit in great with the header, but that will be another PR.

Also: 
* our Router had a bug where `line` got added to queryParams. Fixed that.

Screen Shot 2023-03-28 at 6 57 08 PM

Screen Shot 2023-03-28 at 6 57 20 PM
Screen Shot 2023-03-28 at 6 57 49 PM


devinmatte (approved):
Overall this is good. I do think we need to spend some more time on the UX to fix things like the button colors inside the calendar picker based on the line, and on making things fully polished (especially for buses) but we can do that as follow ups

PR #411, opened 2023-03-29
[v4] Cleanup datepicker further

devinmatte:
Screenshot 2023-03-28 at 8 20 03 PM
Screenshot 2023-03-28 at 8 20 16 PM

### DIFF
--- a/common/components/inputs/DateSelection/DatePickers.tsx
+++ b/common/components/inputs/DateSelection/DatePickers.tsx
@@ -7,6 +7,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import Flatpickr from 'react-flatpickr';
 import 'flatpickr/dist/themes/light.css';
 import { useDelimitatedRoute, useUpdateQuery } from '../../../utils/router';
+import { buttonHighlightConfig } from '../styles/inputStyle';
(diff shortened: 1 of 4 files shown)


common/components/inputs/DateSelection/DatePickers.tsx
```diff
@@ -90,11 +95,15 @@ export const DatePickers: React.FC<DatePickerProps> = ({ config, setConfig }) =>
             </div>
 
             <Flatpickr
-              className="pointer w-32 border-0 py-0"
+              //TODO: Change calendar to line color
+              className={classNames(
+                'w-32 cursor-pointer border-none py-0 focus:ring-opacity-0',
+                line && buttonHighlightConfig[line]
```
PatrickCleary, line 101: Does this do anything if we don't have a ring?

PatrickCleary (approved):

common/components/inputs/DateSelection/DatePickers.tsx
devinmatte: Nope, fixing now

devinmatte: Actually yes, it still shows a slight highlight on the left side, and when set to this color it isn't noticable

PR #413, opened 2023-03-29
V4/sz dates

PatrickCleary:
Slow zones data page needed some work. Still a work in progress but wanted to get this in before the PR became huge.

The new widget isn't perfect, but I think it's better than what we had. If we don't like it, I'd prefer to have no widgets for now for the sake of launching sooner. 

I decided to use deltas rather than averages or maximums for these widgets because I think it makes more sense in the case of slow zones. Average slow zone over a period is not very meaningful information

Changes:
* New type of widget for showing only a delta. I think this is what we should use for slow zone range selections. Doesn't make sense to try to come up with a relevant comparison for any arbitrary date selection.
* Update slow zones data page to be affected by date selection.
* Convert everything to use UTC to adhere to slow zones data format.
* Add prompt to SZ Page for user to add range when on SZ page and single day is selected.
* Changed the Widget setup to work better with 2 different widget types (simple delta + the original)
* Removed some useMemos. I think we might be overusing it anyway. It can hurt performance if it's not used carefully. I think it would be better to wait until we see performance issues then apply it judiciously.

Screen Shot 2023-03-29 at 5 05 02 PM
Screen Shot 2023-03-29 at 5 06 06 PM


common/components/widgets/internal/SimpleDeltaWidget.tsx
```diff
@@ -0,0 +1,67 @@
...
+            className={classNames(
+              'mt-1 flex flex-row items-center rounded-full',
+              widgetValue.delta ? bgColor : 'bg-gray-100',
+              widgetValue.delta ? textColor : 'text-rb-800'
+            )}
+          >
+            <p
+              className={classNames('px-4 text-3xl', widgetValue.delta ? textColor : 'text-rb-800')}
+            >
+              {deltaValue}
```
skaplan-dev, line 44: could we make this smaller? the bubble is lookin LARGE

common/utils/slowZoneUtils.ts
```diff
@@ -76,49 +76,65 @@ export const getRoutes = (data: SlowZone[], direction: Direction) => {
...
-        (sz) =>
-          sz.color === lineShort && dayjs.tz(sz.end, est).isSame(dayjs().subtract(1, 'day'), 'day')
-      ).length,
-    [allSlow, lineShort]
-  );
+  const numZonesOnStartDate =
+    allSlow?.filter((sz) => {
+      return dayjs.utc(sz.start).isSameOrBefore(startDateUTC);
+    }).length ?? null;
+  const numZonesOnEndDate =
```
skaplan-dev, line 104: these names seem long to me. could we use just `start` and `end`? The name of the function should provide enough context for the variables.

common/utils/slowZoneUtils.ts
```diff
@@ -76,49 +76,65 @@ export const getRoutes = (data: SlowZone[], direction: Direction) => {
...
+    allSlow?.filter((sz) => {
+      const zoneEnd = dayjs.utc(sz.end);
+      if (endDateUTC.isSame(TODAY_UTC, 'day')) {
+        // Our latest SZ data is always 1 day behind. So use yesterday's data.
+        return zoneEnd.isSameOrAfter(endDateUTC.subtract(1, 'day'));
+      }
+      return zoneEnd.isSameOrAfter(endDateUTC);
+    }).length ?? null;
```
skaplan-dev, line 112: could we do `0` here instead of `null`? then you wouldn't have to do the `undefined` checks below

common/utils/slowZoneUtils.ts
```diff
@@ -76,49 +76,65 @@ export const getRoutes = (data: SlowZone[], direction: Direction) => {
...
+  lineShort,
+  totals,
+  allSlow,
+  startDateUTC,
+  endDateUTC,
+}: {
+  lineShort?: string;
+  totals?: DayDelayTotals[];
+  allSlow?: SlowZoneResponse[];
+  startDateUTC?: dayjs.Dayjs;
```
skaplan-dev, line 129: same question about optional types. This makes the function really messy with all of the undefined checks

modules/slowzones/SlowZonesDetails.tsx
```diff
@@ -1,58 +1,71 @@
...
-  );
+  const startDateUTC = dayjs.utc(startDate);
+  const endDateUTC = endDate ? dayjs.utc(endDate) : TODAY_UTC;
 
-  const { current, lastWeek, totalDelay, totalDelayLasteek } = useSlowZoneCalculations({
+  const delayTotalsData = delayTotals.data ?? [];
+  const filteredDelayTotals = delayTotalsData.filter((t) => {
+    const date = dayjs.utc(t.date);
+    return date.isSameOrAfter(startDateUTC) && date.isSameOrBefore(endDateUTC);
+  });
```
skaplan-dev, line 31: I memoed these because we're passing over hundreds of data points, which is an overall expensive calculation.

modules/slowzones/SlowZonesDetails.tsx
```diff
@@ -1,58 +1,71 @@
...
-  const formattedTotals = useMemo(
-    () =>
-      delayTotals.data && delayTotals.data.filter((t) => new Date(t.date) > new Date(2022, 0, 1)),
-    [delayTotals.data]
-  );
+  const startDateUTC = dayjs.utc(startDate);
+  const endDateUTC = endDate ? dayjs.utc(endDate) : TODAY_UTC;
 
-  const { current, lastWeek, totalDelay, totalDelayLasteek } = useSlowZoneCalculations({
+  const delayTotalsData = delayTotals.data ?? [];
```
skaplan-dev, line 27: if we don't have `allSlow` or `delayTotals` shouldn't we just return an error page? No need to filter over data we don't have

common/utils/slowZoneUtils.ts
```diff
@@ -76,49 +76,65 @@ export const getRoutes = (data: SlowZone[], direction: Direction) => {
   return [...new Set(routes)];
 };
 
-export const useSlowZoneCalculations = ({
+export const getSlowZoneDelayDelta = ({
+  totals,
   lineShort,
+}: {
+  lineShort?: string;
+  totals?: DayDelayTotals[];
```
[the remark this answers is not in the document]
PatrickCleary: All/most the values are undefined on first load, and if we want to memoize this function we can't optionally call it. So I think I can either do a check that these are all defined before calling this, or change it back to a hook and have the two internal functions be memoized and check for undefined inside those.

What do you think? Or is there another option?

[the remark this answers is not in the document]
PatrickCleary: Ah okay, I think I was going wrong by trying to have the data functions and loading component live at the same level. 
The loading component needs to exist where data may be undefined, and data formatting functions need to exist where data is defined.

So I added a wrapper in between to do the data formatting

devinmatte:
> Add prompt to SZ Page for user to add range when on SZ page and single day is selected.

Are we not able to show _something_ when there's only a single date selected? Maybe counts of current slowzones, or even a different style view that's only focused on $current_date

common/components/inputs/DateSelection/DateConstants.ts
devinmatte, line 1: I feel like this file should be moved to `common/constants/dates.ts` maybe?

modules/slowzones/charts/LineSegments.tsx
```diff
@@ -105,7 +119,13 @@ export const LineSegments: React.FC<{ data: SlowZoneResponse[]; line: LineShort
                 const data = context[0].raw;
                 return (
                   // @ts-expect-error data type is unknown
-                  new Date(data.x[0]).toDateString() + ' - ' + new Date(data.x[1]).toDateString()
+                  `${dayjs.utc(data.x[0]).format('MMM D, YYYY')} - ${
+                    // @ts-expect-error data type is unknown
```
devinmatte, line 123: Why is `data` unknown? That feels dangerous

PatrickCleary: Good call, I was able to find a better field to use for this on the `context` argument.

PatrickCleary:
Also, I am working on a bit of a rework of the navigation that I think will fix the issue with what to display for SZs on a single date @devinmatte

common/components/widgets/internal/SimpleDeltaWidget.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <p className="text-base text-design-subtitleGrey">{widgetValue.getUnits()}</p>
+        </div>
+      </div>
+    );
+  };
+
+  return (
+    <>
+      <div className={classNames('relative flex  flex-1 bg-white')}>
+        {widgetValue.value === undefined && <LoadingSpinner />}
```
skaplan-dev, line 56: you could do `!widgetValue` here

skaplan-dev (approved):

common/components/widgets/internal/SimpleDeltaWidget.tsx
PatrickCleary: Sometimes it is 0. That got me a couple times

PR #429, opened 2023-04-01
V4/trip explorer

PatrickCleary:
Add back the "classic" 3 stacked graphs. It is currently not in the navigation menu, you can find it at `/<lineName>/trips`
I started calling it the "Trip Explorer" but I'm open to suggestions
- The `customQueries` code is not new, I removed it a week or so ago and just added it back (with some small modifications)
- I did no design work, that will come next. Trying to keep PRs smaller.
Screen Shot 2023-04-01 at 12 48 39 PM


devinmatte (approved):
Looks good as a starter, the navigation is highlighting the ridership tab, but otherwise, looks good to me

Screenshot 2023-04-01 at 3 23 16 PM

Issue #113, opened 2021-12-08
Add TM watermark to figures


austinjpaul:

devinmatte:
We can use https://www.npmjs.com/package/chartjs-plugin-watermark to easily add the transitmatters logo somewhere in the page

PR #433, opened 2023-04-01
Watermarking charts

devinmatte:
Watermark our charts #113 

Screenshot 2023-05-01 at 2 39 24 PM
Screenshot 2023-05-01 at 2 39 32 PM

## Discussions

- What corner should it be in? Top/Bottom/Left/Right?


devinmatte:
Requires https://github.com/AlbinoDrought/chartjs-plugin-watermark/pull/9 to work

devinmatte:
Screenshot 2023-05-01 at 2 39 24 PM
Screenshot 2023-05-01 at 2 39 32 PM

PatrickCleary:
I like it. I think the location and opacity are good. My only suggestion might be to make the whole thing one color, so remove the red from "Matters". But I think either way it's good

PatrickCleary (approved):

PR #437, opened 2023-04-04
Slow Zones map visualization

idreyn:
This PR adds:

- A generalized `LineMap` component that uses a rewritten version of the NTT line rendering engine to draw strip maps
- A `SlowZonesMap` component that uses `LineMap` to show slow zones
- Storybook, for helping us debug/test React components.

There's still work I want to do on this: tooltips, layout edge cases, etc. But I think it's in a good state to share!

image

_Test plan:_
- Visit the Slow Zones tab for each of Red, Orange, Blue and make sure it renders
- Run `npm run storybook` and make sure that works too.


PatrickCleary:
Looks fantastic. Couple small things I noticed:
* I think the labels are flipped compared to the highlights
* It seems like the hardest part of getting this right will be placing the labels. I actually think we could do without the labels and just do tooltips? I think the highlights do a good job of telling the story.
*  Do you think this will work with a range of dates? One idea might be to show deltas for each SZ segment. Right now that would just be added or removed SZs, but I think we could eventually have daily avgs for SZs

.eslintrc.js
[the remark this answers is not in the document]
devinmatte: https://github.com/transitmatters/t-performance-dash/pull/440 should make this less likely to happen in the future

devinmatte:
@idreyn Is this expected to show live data yet/work for the blue line?
Screenshot 2023-04-04 at 7 40 12 PM
This also could be us in v4 still using non-live data and not your fault, just checking

package.json
```diff
@@ -35,7 +36,17 @@
     "react-responsive": "^9.0.2"
   },
   "devDependencies": {
+    "@babel/core": "^7.21.4",
+    "@storybook/addon-actions": "^6.5.16",
+    "@storybook/addon-essentials": "^6.5.16",
+    "@storybook/addon-interactions": "^6.5.16",
+    "@storybook/addon-links": "^6.5.16",
+    "@storybook/builder-webpack5": "^6.5.16",
```
devinmatte, line 44: Looks like 7.0.2 released yesterday https://github.com/storybookjs/storybook/releases/tag/v7.0.2. 
Might be worth upgrading before launching v4, so we're not as worried about breaking something in prod by upgrading. This can be translated to an issue, doesn't need to be done now

modules/slowzones/types.ts
```diff
@@ -0,0 +1,7 @@
+const SLOW_ZONES_LINES = ['Red', 'Blue', 'Orange'] as const;
+
+export type SlowZonesLineName = (typeof SLOW_ZONES_LINES)[number];
+
+export const isSlowZonesLine = (name: unknown): name is SlowZonesLineName =>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
devinmatte, line 6: We don't need to disable `@typescript-eslint/no-explicit-any` as it's just a warning. We should leave the warning to remind us to fix this in the future

modules/slowzones/map/SlowZonesMap.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+    duration: 20,
+  },
+];
+
+export const Primary = () => {
+  return (
+    <>
+      <SlowZonesMap
+        lineName="Red"
+        slowZones={slowZonesResponses as SlowZoneResponse[]}
```
devinmatte, line 42: We shouldn't need to `as` translate the types here. `slowZonesResponses` is already typed properly above

modules/slowzones/map/SlowSegmentLabel.tsx
```diff
@@ -0,0 +1,83 @@
...
+        color={color}
+        isHorizontal={isHorizontal}
+        size={2}
+      />
+      {delayString}
+    </div>
+  );
+};
+
+const SlowSegmentLabel = (props: SlowSegmentLabelProps) => {
```
devinmatte, line 60: In modern react, if you should use the `React.FC` or `React.FunctionComponent` type, with the props interface passed into that typing.
```suggestion
const SlowSegmentLabel: React.FC<SlowSegmentLabelProps> = (props) => {
```

We also should use object spreading for the props and avoid `props.` when possible.

```
const SlowSegmentLabel: React.FC<SlowSegmentLabelProps> = ({ slowZonesByDirection, line, isHorizontal }) => {
```

idreyn:
@devinmatte I don't know exactly how this works in production but if you `npm run get-sz-data` it'll pull down the latest slow zones, including the new BL ones.

modules/slowzones/map/SlowSegmentLabel.tsx
```diff
@@ -0,0 +1,83 @@
...
+};
+
+type SlowZoneLabelProps = {
+  direction: SlowZoneDirection;
+  slowZone: SlowZoneResponse;
+  color: string;
+  isHorizontal: boolean;
+};
+
+const SlowZoneLabel = (props: SlowZoneLabelProps) => {
```
devinmatte, line 26: Same thing as I stated below
```suggestion
const SlowZoneLabel: React.FC<SlowZoneLabelProps> = (props) => {
```

modules/slowzones/map/DirectionIndicator.tsx
```diff
@@ -0,0 +1,31 @@
...
+import styles from './DirectionIndicator.module.css';
+
+type Props = {
+  direction: SlowZoneDirection;
+  color: string;
+  isHorizontal: boolean;
+  size?: number;
+};
+
+const DirectionIndicator = (props: Props) => {
```
devinmatte, line 14: Same thing as I stated elsewhere
```suggestion
const DirectionIndicator: React.FC<Props> = (props) => {
```

modules/slowzones/map/DirectionIndicator.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import type { SlowZoneDirection } from './segment';
+
+import styles from './DirectionIndicator.module.css';
+
+type Props = {
```
devinmatte, line 7: This should be an interface, and be more specifically named
```suggestion
interface DirectionIndicatorProps {
```

devinmatte (commented):
Overall looks awesome! I have a few comments around styling and modern react/ts before we get this in, but love the UX so far

devinmatte:
I added the same packages installs to https://github.com/transitmatters/t-performance-dash/pull/441 to help shrink the line changes to a more realistic value

modules/slowzones/map/SlowZonesTooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+        </div>
+        <div className={styles.body}>
+          {DIRECTIONS.map((direction) => renderSlowZoneForDirection(direction))}
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default SlowZonesTooltip;
```
devinmatte, line 92: We should avoid default export when possible

modules/slowzones/types.ts
```diff
@@ -0,0 +1,6 @@
+const SLOW_ZONES_LINES = ['Red', 'Blue', 'Orange'] as const;
+
+export type SlowZonesLineName = (typeof SLOW_ZONES_LINES)[number];
+
+export const isSlowZonesLine = (name: unknown): name is SlowZonesLineName =>
+  SLOW_ZONES_LINES.includes(name as any);
```
devinmatte, line 6: Do we need an `as any` here? Is `name` really `unknown`? Is it not a string?

package-lock.json
devinmatte, line 1: You seem to be downgrading a few packages here. Maybe worth resetting against `dashboard-v4` and then `npm install`ing to make sure

modules/slowzones/SlowZonesDetails.tsx
```diff
@@ -56,6 +59,22 @@ export default function SlowZonesDetails() {
...
+              lineName={lineShort}
+              direction="horizontal-on-desktop"
+            />
+          ) : (
+            <div className="relative flex h-full">
+              <ChartPlaceHolder query={delayTotals} />
+            </div>
+          )}
+        </div>
+      </div>
```
devinmatte, line 77: I think the map should probably be shown on the single day view as well. We can follow up on this though, not a blocker for merging by any means

devinmatte (commented):
Looking great, few more code style comments and thoughts.

All good from a testing perspective

modules/slowzones/types.ts
```diff
@@ -0,0 +1,6 @@
+const SLOW_ZONES_LINES = ['Red', 'Blue', 'Orange'] as const;
+
+export type SlowZonesLineName = (typeof SLOW_ZONES_LINES)[number];
+
+export const isSlowZonesLine = (name: unknown): name is SlowZonesLineName =>
```
devinmatte, line 5: This seems like it's a util, not a type

common/components/widgets/internal/BasicWidgetDataLayout.tsx
```diff
@@ -4,21 +4,32 @@ import type { WidgetValueInterface } from '../../../types/basicWidgets';
 import { LoadingSpinner } from '../../graphics/LoadingSpinner';
 import { Delta } from './Delta';
 
+export type LayoutKind = 'total-and-delta' | 'delta-and-percent-change';
 export type SentimentDirection = 'positiveOnIncrease' | 'negativeOnIncrease';
 
 export type BasicWidgetDataLayoutProps = {
-  title: string;
-  analysis: string;
+  title: React.ReactNode;
```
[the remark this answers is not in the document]
idreyn: `React.ReactNode` includes `string` (and `boolean`, `null`, and anything else that can be returned by a React component)

modules/slowzones/map/SlowZonesMap.module.css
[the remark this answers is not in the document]
idreyn: No, good catch!

package-lock.json
idreyn: Got it. I think since you moved the deps for this PR into `dashboard-v4` I can safely check out `package.json` and `package-lock.json` from that branch.

modules/slowzones/types.ts
idreyn: True true. It's a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) and I always stress about where to put these but I find that the best place is usually next to the types that they guard for.

In this case we're no longer using it so imma just delete it.

modules/slowzones/map/SlowZonesTooltip.module.css
PatrickCleary, line 1: We've mostly been using Tailwind classNames for styling. I also don't think we use size 14 font anywhere else - I think tailwind `text-sm` (12px/.75rem) or `text-base` (16px/1rem) might fit in better. Those tailwind values also adjust with system font sizing I believe.

If you want, I can open a PR on this to convert this file to tailwind? Lmk

PatrickCleary: Can also wait until after it's merged if you're pretty much finished

PatrickCleary (approved):

modules/slowzones/map/SlowZonesTooltip.module.css
idreyn: I was able to remove that `font-size` and it still looks good! (I think it was only being used by the title)

I don't know if you'll be able to convert the entire thing to Tailwind (e.g. the tooltip triangle via CSS border hack) but I support using it to the extent possible. I have been struggling a bit with Tailwind in this repo because I don't have a good sense of which classes are getting the most use across the codebase, and so the classes I choose feel arbitrary.

Once v4 is out it might be fun(?) to go in and create a style guide or a styles.js for this.

PR #444, opened 2023-04-05
MBTA speed restrictions data in SZT

idreyn:
This PR pulls in speed restrictions data from the MBTA open data portal, and shows it in the SZT. The URL for the data is hard-coded to February for now, but they've told us they're going to start publishing regular updates as CSVs soon.

The design is pretty basic, but all the data is getting piped in and there's plenty we can do with this. One thought is that we might want to show this data in the "Locations" bar chart view, too.

_Screenshot:_

image

_Test plan:_

- Run `npm run get-sz-data` and verify that it now creates a `public/static/slowzones/speed_restrictions.json`
- Open the SZT and verify that you can see the new data in the map tooltips.


server/scripts/update_stations.py
```diff
@@ -1,5 +1,4 @@
 import json
```
idreyn, line 1: @devinmatte I moved the top-level Python `scripts/` directory to `server/scripts` so that Poetry can manage it.

PatrickCleary:
This is gonna be so sick.
So the MBTA SZ data can have multiple entries per section right? And you are aggregating them for the `oldest` `slowest` metrics right?

1 bug I found is the orange and blue lines are getting RL values drawn on for some reason
Screen Shot 2023-04-07 at 12 58 46 PM

Other thoughts I had that shouldn't prevent this from being merged:
1) would be nice to have some sort of animation for the tooltips like the chartJS ones. Feels a little jarring as is.
2) I think the `over baseline` delta here shouldn't have the +/- symbol. 
Screen Shot 2023-04-07 at 1 01 46 PM
3) When there is only a SB/NB value, should it be center-aligned or aligned to the side it is going? For example in this screenshot, it is only SB and I would suggest it aligns on the right of the box.
Screen Shot 2023-04-07 at 1 03 08 PM

idreyn:
> This is gonna be so sick. So the MBTA SZ data can have multiple entries per section right? And you are aggregating them for the `oldest` `slowest` metrics right?

Yeah, I'm picking the most egregious value for each one, and they might all refer to different speed restriction zones on the unidirectional track segment between the two stations. I would feel kinda bad about this if the T didn't do the same thing on [their own map](https://www.mbta.com/performance-metrics/speed-restrictions).

> 1 bug I found is the orange and blue lines are getting RL values drawn on for some reason Screen Shot 2023-04-07 at 12 58 46 PM

@PatrickCleary  Bizarre, good catch, but I cannot reproduce it locally. Could it be a Next.js HMR thing happening as you change between lines? I just pushed up a change that adds a `key={lineShort}` to the `SlowZonesMap` instance which should force it to remount between line pages. Does that make this go away for you? 

> 1. would be nice to have some sort of animation for the tooltips like the chartJS ones. Feels a little jarring as is.

I'll think about it, but I kinda like how snappy this feels. (One weird animation thing we could do is set `snapToSegment: false` in our `<LineMap>` instance so that the tooltip always follows the cursor).

> I think the `over baseline` delta here shouldn't have the +/- symbol.

It is redundant here. But I don't know if I want to add a special case for this inside the `Widget`-related components. It seems like in the general case you would want the +/-, but here you always know it's +.

>  When there is only a SB/NB value, should it be center-aligned or aligned to the side it is going? For example in this screenshot, it is only SB and I would suggest it aligns on the right of the box.

Hmm.
I tried right-aligning the SB ones, and it felt "jumpy" to me when moving between SB-only and NB-only segments.

PatrickCleary:
Still getting the issue, but I can look into it if you're not able to reproduce.

As for the animations, I like no-animation when hovering between segments. I would only suggest an animation when the first one appears. Idk if that distinction affects your opinion. 

Yea, jumping from left-to-right side is probably not good. I'm good with how it is.

One other random though: might be cool to have the segment of the map visually change when it is hovered

devinmatte:
Noting an error in the console
```
client.js?4d30:1 Warning: React does not recognize the `pathDirective` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `pathdirective` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

idreyn:
The `pathDirective` issue above, as well as an issue with duplicate React keys, should now be fixed. Happy to merge this in immediately if someone can approve.

devinmatte:
Screenshot 2023-05-01 at 8 02 15 PM
Oops. Merge created a duplicate I guess

devinmatte (approved):
LGTM, probably want a second set of eyes to make sure I didn't miss anything in my merges before merging if possible

PatrickCleary:
When I switch from red line to orange line, the labels persist. Was happening on the old branch too at some point, then got fixed. Not a blocker, just something to keep an eye on.
Screen Shot 2023-05-01 at 5 12 03 PM

idreyn:
> When I switch from red line to orange line, the labels persist.

There's gotta be some `useEffect` dependency missing somewhere but for now I am just slapping a `key` on that `SlowZonesMap` instance to force it to remount when the line changes.

modules/slowzones/map/segment.ts
```diff
@@ -41,95 +48,108 @@ const getEffectiveDate = (desiredDate: Date, slowZones: SlowZoneResponse[]) => {
...
-    }
-    otherMatchingZones.forEach((zone) => handledSlowZones.add(zone));
-  });
-  return { segmentedSlowZones, effectiveDate };
+  const activeSpeedRestrictions = locateIntoSegments(
+    filterActiveElements(
+      speedRestrictions,
+      lineName,
+      effectiveDate,
+      (rs) => new Date(rs.cleared ?? '4000'),
```
PatrickCleary, line 145: Why 4000?

modules/slowzones/map/SlowZonesTooltip.tsx
```diff
@@ -5,66 +5,121 @@ import { getParentStationForStopId } from '../../../common/utils/stations';
...
+      const [oldest] = speedRestrictions
+        .filter((sr) => sr.reported)
+        .map((sr) => new Date(sr.reported).toISOString())
+        .sort();
+      const oldestString = prettyDate(oldest, false);
+      return (
+        <div
+          className={classNames(
+            styles.speedRestriction,
+            'bg-slate-200 text-base text-xs text-slate-700'
```
PatrickCleary, line 75: you have two text sizes here

PatrickCleary: Also, we've mostly used tailwind's `stone` or `gray` colors for grey

modules/slowzones/map/SlowZonesMap.tsx
```diff
@@ -32,78 +34,78 @@ const abbreviateStationName = ({ stationName }) => {
...
+    zones = [zones];
   }
-  // Shawmut to Ashmont — obscured by "North Quincy" label
-  if (!isHorizontal && toStationId === 'place-asmnl') {
-    return {
-      mapSide: '1' as const,
-    };
-  }
-  return null;
+  const zonesArray = zones as SlowZoneResponse[];
```
PatrickCleary, line 41: Is `zones` ever not an array? It's typed as an array up until here

PatrickCleary (approved):
looks good to me. Couple small comments

modules/slowzones/map/SlowZonesMap.tsx
devinmatte: Good catch, we shouldn't need to `as` anything

modules/slowzones/map/segment.ts
idreyn: It's sufficiently far in the future 😆

PatrickCleary: image

us in the year 4000

idreyn:
Closing in favor of #531, which has a different merge that preserves some newer work that was in commits to this branch.

PR #446, opened 2023-04-05
Adding new presets for datepicker

devinmatte:
Adding a few new preset date ranges and dates to make things simpler

Screenshot 2023-04-05 at 12 24 03 PM
Screenshot 2023-04-05 at 12 24 08 PM


PatrickCleary:
If you want to get rid of the April 2023/March 2023 presets that's fine with me. I haven't found them that useful.

Issue #436, opened 2023-04-03
Upgrade Pandas to 2.0.0+


devinmatte:
https://github.com/pandas-dev/pandas/releases/tag/v2.0.0

We should do this in v4

PR #451, opened 2023-04-06
Updating pandas

devinmatte:
Seems to just ... work

We should run this in v4 so that if there are any bugs, we can find them, and track them down before release. Worst case, we revert back to `1.5.3`, but, better to do it now than after launch

As far as I can tell though, it should work fine

#436


austinjpaul:
There is a lot of pandas code that is only run offline for data ingestion. I don't want to merge this without checking that also.

devinmatte:
I'll require we merge https://github.com/transitmatters/data-ingestion/pull/9 before merging this. Going to revert this to draft for now

austinjpaul:
The scripts I mentioned live in server/rapid and server/bus in this repo,
unrelated to the ingestion repo (unless we decide to move them there
someday) Sorry to confuse!

devinmatte:
No worries. I'll check those out. Either way though, probably not a bad idea to get all the repos on the same version anyway

PR #453, opened 2023-04-06
V4/new pages

PatrickCleary:
Going to do my best to break up the nav changes into chunks.

First off - adding new files and folders for range and single day trip explorers. No way to navigate to these yet, but you can go to the URL manually.


devinmatte (approved):
Looks good, although unsure what's going on with the package-lock

PR #454, opened 2023-04-06
V4/speed changes

PatrickCleary:
Remove default comparison ranges from speed widgets. I don't think this is feasible, better to just show deltas over given time period. 

**Before**
Screen Shot 2023-04-06 at 1 15 40 PM

**After**
Screen Shot 2023-04-06 at 1 14 27 PM


devinmatte:
Screenshot 2023-04-06 at 3 28 42 PM
I'm seeing two different speed values in this branch

PatrickCleary:
> I'm seeing two different speed values in this branch

Ah, good point.
The speed widget uses weekly/monthly aggregated values when longer date ranges are selected. So in this case, it is using the most recent weekly value, which differs from today's value.

I think there are three solutions (or lmk if you see another one):
1) show today's value on that widget (in which case it would not match with last point on the graph).
2) Update the widget to say "Speed for week of April 2nd"
3) Don't show the current value on that widget. Not sure what we could replace it with..

devinmatte:
@PatrickCleary I think the labels need to be clearer then. They both basically insinuate that they're current speed, when one is more current than the other. So the labels are the issue I think

PatrickCleary:
Screen Shot 2023-04-07 at 11 56 04 AM
Screen Shot 2023-04-07 at 11 56 36 AM
Screen Shot 2023-04-07 at 11 58 06 AM


What do you think about these titles? @devinmatte

devinmatte:
@PatrickCleary That's probably better and less likely to confuse

devinmatte (commented):
Rather not have the daterange changes as part of this. In this PR, the ability to change between single date and range for travel times no longer works

What I've seen for speed specifically all seem okay

modules/dashboard/LineHealth.tsx
```diff
@@ -11,18 +8,13 @@ import { SpeedWidget } from '../speed/SpeedWidget';
...
 export const LineHealth = () => {
-  const [timeRange, setTimeRange] = useState<TimeRange>('week');
   const { tab, line } = useDelimitatedRoute();
 
   return (
     <div>
-      <div className="flex w-full flex-col justify-between sm:flex-row">
-        <h1 className="text-xl">Line Health</h1>
-        <ButtonGroup pressFunction={setTimeRange} options={Object.entries(TimeRangeNames)} />
-      </div>
```
devinmatte, line 22: Unsure why this isn't replaced with the new `OverviewDateSelection`, even if not wired up

devinmatte (approved):
One question, but otherwise looks good

PR #455, opened 2023-04-06
Remove some codeowners for v4

PatrickCleary:
Removing people who aren't actively working on V4 from the `CODEOWNERS` file.

Will remove this change when merging back into main, but I figured I would save you guys a few hundred emails

@mathcolo @friendchris @nathan-weinberg if any of you don't want to be removed lmk and I can add you back.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,3 @@
 * @austinjpaul @skaplan-dev @devinmatte @PatrickCleary
+.github @nathan-weinberg
+deploy.sh @mathcolo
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @austinjpaul @mathcolo @friendchris @nathan-weinberg @skaplan-dev @devinmatte @PatrickCleary
+* @austinjpaul @skaplan-dev @devinmatte @PatrickCleary
```
devinmatte, line 1: We could do actual structured codeowners where people are tagged only on files they care about

```suggestion
* @austinjpaul @skaplan-dev @devinmatte @PatrickCleary
.github @nathan-weinberg
deploy.sh @mathcolo
```

PR #456, opened 2023-04-06
Change Date Picker to only be range OR Single day

PatrickCleary:
Date picker will only serve as a range date picker or a single day picker. Prop `range` controls that.

This is in preparation for upcoming nav change which will not have any pages which can be both ranged and single day pages.


devinmatte:
So on the 
>not have any pages which can be both ranged and single day pages.
What are we doing with slow zones? As slow zones both has a "today" value as well as a range that lets you see changes over time. Are we just going to have the range only apply to some, and have the today options always shown as today?

PatrickCleary:
Discussed with @devinmatte 

We came to the conclusion:
- Trip explorer will revert to having the swappable date range picker or something similar.
- all `Overview` sub pages will have the range only date picker. Slow zones map will be pinned to the end date with a visual indicator to show it is not a range view.

Issue #452, opened 2023-04-06
Classic View for Buses


PatrickCleary:
Seems our customQueries aggregate queries don't work for buses. Probably because no dwells for buses? Idk.

Issue #401, opened 2023-03-25
Mobile Navigation Cleanup


devinmatte:
Before launch, we need to make sure the mobile navigation is finished and cleaned up

Fixing spacing issues like this
<img width="411" alt="Screenshot 2023-03-25 at 4 34 22 PM" src="https://user-images.githubusercontent.com/9310513/227740537-e2ae73e9-218d-4f09-945c-a09be04d7c2b.png">

Fixing my initial range UI (we can do a lot better)
<img width="416" alt="Screenshot 2023-03-25 at 4 39 22 PM" src="https://user-images.githubusercontent.com/9310513/227740561-efda12be-4e6f-495b-9014-799894dd75ba.png">

devinmatte:
There are type errors showing up for bus navigation on mobile

PR #458, opened 2023-04-06
V4/nav rework

PatrickCleary:
## Motivation

The navigation needed a bit of an overhaul. There were a few inconsistencies:
- Our previous tab navigation gave the feeling that all the pages were at equal standings. This change shows the hierarchy among the pages.
- Some pages shared parameters and some did not. The new page organization takes that into account and organizes the pages such that each section has one set of parameters. `Trips` has date selection and station selection. `Line` has just date selection
- We have far less data for bus routes than we do for subways. Previously we were displaying them as equivalents. Now buses have their own tab, and that tab only has the pages relevant to buses.
- There was no place for system-wide metrics.
- Old nav was fairly rigid. No place to add new pages which may have a new set of parameters.

## Changes
- Moved all navigation to the sidebar
- Added a `bus`/`subway`/`system` tab
- Separated the old `overview` page into `today` and `overview`
- Put all station-based metrics into the `trips` section.
- Separated station-based metrics into `range` and `single day` tabs. This is up for debate. The plan is to share with the group on Thursday, and then make a decision.
**Sidebar nav**
Screen Shot 2023-04-11 at 9 00 37 AM

**Range/single day tabs** (undecided)
Screen Shot 2023-04-11 at 9 01 09 AM

## Testing Instructions

Click around between all the pages. Keep in mind, most of the logic for query params is coming in the next PR. So things like default parameters, and persistent parameters are not currently working. Should probably just ignore the query params for now.


devinmatte:
Initial thoughts (some of the same discussed in slack)

Personally I don't think we should seperate out range and single day views
Screenshot 2023-04-06 at 4 41 55 PM

As @skaplan-dev brought up, speed might not be the most useful to a rider, so making the metric front and center might not be the most advisable. Although, we can discuss this with more of labs to determine order
Screenshot 2023-04-06 at 4 47 11 PM

I also noticed that dates disappear from query params when jumping between pages. Ideally we'd persist the query params even if a page ignores them. Same with preferences between range and single select, like we do today

devinmatte:
One minor, nitpicky, thought towards not separating singe/range is when separated on my large monitor, I had to scroll to see all option in the sidebar. Less options on the sidebar brings more things above the fold for the average user
Screenshot 2023-04-06 at 4 26 59 PM

devinmatte:
Notes so far:

### 1: Love the floating datepicker
Screenshot 2023-04-09 at 6 32 55 PM

### 2: One bug with the tabs
One bug with the tabs, and maybe with the concept, is sometimes it can get into a single date selection state, despite selecting range, which looks much weirder than if you didn't have the tabs

Screenshot 2023-04-09 at 6 33 57 PM

### 3: Overview page should have white widget borders
They should have the same widget borders, or, one large white border for the section. It living on the background feels weird (I know this might be a different problem to solve)

Screenshot 2023-04-09 at 6 36 11 PM

### 4: Two overview pages
Not a fan of both of these pages being called the same thing. Overview works in the main `line` section, but I don't love the name for the Trip section

Screenshot 2023-04-09 at 6 42 58 PM

### 5: Alert taking full width while speed doesn't
It's odd to me that they're different sizes. Either they'd sit next to each other, or, both be full width

Screenshot 2023-04-09 at 6 48 33 PM

PatrickCleary:
> 1: Love the floating datepicker
😁

> 2: One bug with the tabs

True, though I think we could add a way to prevent this



> 3: Overview page should have white widget borders

Agreed

> ### 4: Two overview pages

Yea, I think you're right. I'm open to suggestions for other names.


> ### 5: Alert taking full width while speed doesn't

Yea, I agree. Ideally we will have a third thing there soon and can adjust these accordingly. I was thinking full alerts then two half-size below. Although now idk if alerts needs that much space. For now I can make them all full-width

common/components/inputs/DateSelection/DatePickers.tsx
devinmatte, line 1: I am not convinced that we should remove the code for a date picker to be swapped between single and range. It may be too much work, but I'd prefer a prop that disables the setting instead of removing the code

devinmatte:
Screenshot 2023-04-11 at 12 55 19 PM
🐛

devinmatte (approved):
I did as good of a pass as I felt I could. I've noted the errors I saw that should be fixed either in this PR or in https://github.com/transitmatters/t-performance-dash/pull/472

common/layouts/DashboardLayout.tsx
```diff
@@ -12,6 +15,17 @@ interface DashboardLayoutProps {
...
+  const {
+    page,
+    query: { single },
+  } = useDelimitatedRoute();
+
+  const section = ALL_PAGES[page]?.section;
+  const getDatePicker = () => {
+    if (section === 'trips' || section === 'line')
+      return <DateSelection type={single ? 'single' : 'range'} />;
+    if (section === 'overview') return <OverviewDateSelection />;
```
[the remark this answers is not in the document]
PatrickCleary: In this PR or new issue?

common/utils/router.tsx
```diff
@@ -17,29 +18,34 @@ const linePathToKeyMap: Record<string, Line> = {
   bus: 'BUS',
 };
 
-const getParams = (params: QueryParams) => {
+const getParams = (params) => {
```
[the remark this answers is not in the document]
devinmatte: Maybe a `Partial<QueryParams>`? But that might be too much effort. We can leave it for now. I do want to add more TS rules later, but we can follow up on that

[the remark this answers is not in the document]
PatrickCleary: Discussed in https://github.com/transitmatters/t-performance-dash/pull/458#discussion_r1162801742

devinmatte:
I should note a caveat to my review, that I am still not a fan of the tab `range` vs `single`. I much prefer our original method for date selection changes. I am however fine with merging in this manor since it is easily reversible should the majority of labs agree

PatrickCleary:
https://github.com/transitmatters/t-performance-dash/pull/458#issuecomment-1503756299

Will be fixed in next PR

modules/dashboard/DataPageHeader.tsx
```diff
@@ -1,40 +1,27 @@
...
-  } = useDelimitatedRoute();
-  const [range, setRange] = useState<boolean>(endDate !== undefined);
-
-  React.useEffect(() => {
-    if (!range && endDate !== undefined) {
-      setRange(true);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [endDate]);
+  const { line, page } = useDelimitatedRoute();
```
PatrickCleary, line 9: Don't need any of the `range` state because the date picker lives in the `DashboardLayout` now and is self-contained.

modules/navigation/BusSelection.tsx
PatrickCleary, line 1: This is mostly preexisting code that was in `modules/navigation/desktop/BusDropdownItem.tsx`

common/layouts/DashboardLayout.tsx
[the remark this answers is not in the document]
devinmatte: We can include it as part of https://github.com/transitmatters/t-performance-dash/issues/401

modules/navigation/LineSelection.tsx
PatrickCleary, line 1: Line Selection. Mostly reused old line selection. Could use a bit more UI work.
Screen Shot 2023-04-11 at 1 25 15 PM

modules/navigation/RangeTabs.tsx
PatrickCleary, line 1: Attempt to differentiate more between `range` and `single day` data types. As first suggested by @austinjpaul.

This change is definitely undecided. I put it in this pr just so we could try it out and demo it on Thursday. I think we'll be able to get a better sense for it if we can see it working. If consensus is opposed, will revert.

I also think maybe `range` and `single day` are not the best names. Because the real difference is `per day aggregates` vs. `per-trip` data. So maybe new names would be better.

Screen Shot 2023-04-11 at 1 26 21 PM

modules/navigation/SidebarTabs.tsx
PatrickCleary, line 1: Bulk of the changes. Open to naming, icon, and ordering suggestions for anything. These were all just what I put in to start.
Screen Shot 2023-04-11 at 1 29 44 PM

devinmatte: I feel like some variation of `Trip Explorer`, `Explore` is better than using `Overview` in two spots.

As for ordering I think we had discussed an order above that might be more valuable to a user
image

PatrickCleary: I like `Explore`. And just updated the ordering 👍

PatrickCleary: I was also using `Trip Overview` as the title. What about that?

PatrickCleary: `Summary` ?

devinmatte: Trip Explorer or Trip Summary both seem reasonable. Not sure on final naming

modules/dashboard/Today.tsx
```diff
@@ -0,0 +1,33 @@
...
+export default function TodaysCommute() {
+  const { tab, line, lineShort } = useDelimitatedRoute();
+  const allSlow = useQuery(['allSlow'], fetchAllSlow);
+  const canShowSlowZonesMap = lineShort === 'Red' || lineShort === 'Blue' || lineShort === 'Orange';
+
+  return (
+    <div className="flex flex-col gap-y-4">
+      <div className="flex flex-col gap-x-4 gap-y-4 xl:flex-row">
+        <Alerts />
+        {tab === 'Subway' && line !== 'GL' && <Speed />}
```
skaplan-dev, line 19: i feel like I'm seeing this repeated a lot. (checking if we are in subway view)
Maybe we should separate Subway and Bus `Today` views?

skaplan-dev, line 19: we also have a variable storing `line !==  "GL"` already above with canShowSlowZonesMap.

modules/navigation/SideNavigation.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { useDelimitatedRoute } from '../../common/utils/router';
+import type { Tab } from '../../common/types/router';
+import { DashboardSelection } from './DashboardSelection';
+import { BusNavMenu } from './BusNavMenu';
+import { SubwayNavMenu } from './SubwayNavMenu';
+
+interface SideNavigationProps {
+  setSidebarOpen?: React.Dispatch<React.SetStateAction<boolean>>;
```
skaplan-dev, line 9: why is this optional?

skaplan-dev (changes requested):
## 
Took a first pass. I think this would be a lot cleaner if we had some more containers that were checking bus/subway mode. Then you wouldn't have so many if statements all over the place checking `line`/etc.

Also pulled this down locally and here's a list of things that aren't working for me.

- Default route broken
- Speed widget broken
- Station picker under explore doesn’t change for headways / traveltimes when switching line
- Bus page is broken
- Aggregate apis aren’t working for me locally

## Questions
- Can you explain having travel times / headways / dwells under trips and line?

## Suggestions
- Tooltips on disabled nav items?

modules/navigation/BusNavMenu.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { BUS_OVERVIEW, BUS_PAGES } from '../../common/constants/pages';
+import { BusSelection } from './BusSelection';
+import { SidebarTabs } from './SidebarTabs';
+
+interface BusNavMenuProps {
+  setSidebarOpen?: React.Dispatch<React.SetStateAction<boolean>>;
```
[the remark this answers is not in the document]
PatrickCleary: It's only needed on the mobile nav sidebar. https://github.com/transitmatters/t-performance-dash/blob/dashboard-v4/modules/navigation/desktop/SideNavBar.tsx#L91-L112

modules/navigation/RangeTabs.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useDelimitatedRoute } from '../../common/utils/router';
+import { lineColorBorder } from '../../common/styles/general';
+import { ONE_WEEK_AGO_STRING, TODAY_STRING } from '../../common/constants/dates';
+
+export const RangeTabs = () => {
+  const route = useDelimitatedRoute();
+  const { query, line } = route;
+  const router = useRouter();
+
+  const selected = query.single === 'yes' ? 1 : 0;
```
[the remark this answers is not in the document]
PatrickCleary: Just signifies whether the page is in `single day` mode. I don't have a great idea for what to name it instead. 
single=`true`?
type=`singleday`?

PatrickCleary:
Thanks for the thorough review  @skaplan-dev  😁
I think you make a good point about checking for line. I will look into your suggestion - my immediate thought is that since most components just pull the line from `useDelimitatedRoute()` I would think any prior gating on `line` wouldn't be known to the component. I could maybe do something inside the router? idk. Do you have any ideas? or am I missing something?
 
 Also in top-level pages, they would have no way of being gated other than checking the params I would think.

> * Default route broken

fixed.

> * Speed widget broken

I can't seem to replicate this

> * Station picker under explore doesn’t change for headways / traveltimes when switching line

I can probably figure this out, but the next change is overhauling the station selector and then it will pull values from the query params. So that will fix it. Do you think it needs to be fixed in this PR?

> * Bus page is broken

Preexisting bug:
https://github.com/transitmatters/t-performance-dash/issues/452

> * Aggregate apis aren’t working for me locally

I also haven't been able to reproduce this

> * Can you explain having travel times / headways / dwells under trips and line?

Do you mean why they are not under trips and line? If so, the pages are organized by the parameters they take. The `line` section is station-agnostic. The `trips` section has the widgets which require a station to be chosen. Another reason to separate them is the "expected" date ranges are very different. A date range of a year or more in the `line` section is to be expected, whereas it will be very slow or not work in the `trips` section. If someone is clicking around in the `line` section then goes to "headways" and nothing loads, that will be a source of confusion/friction.

> * Why are we using `GL` and `Green` / etc . Shouldn't we stick to one denominator for the line? It seems like they're describing the same thing

https://github.com/transitmatters/t-performance-dash/pull/459



> * Tooltips on disabled nav items?

All for it. I can create an issue? https://github.com/transitmatters/t-performance-dash/issues/476

modules/navigation/SidebarTabs.tsx
```diff
@@ -0,0 +1,66 @@
...
+import React from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { useRouter } from 'next/router';
+import { handleTabNavigation, useDelimitatedRoute } from '../../common/utils/router';
+import type { NavTab } from '../../common/constants/pages';
+
+interface SidebarTabs {
+  tabs: NavTab[];
+  title: string;
+  setSidebarOpen?: React.Dispatch<React.SetStateAction<boolean>>;
```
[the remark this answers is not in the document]
PatrickCleary: answered aboce

PR #464, opened 2023-04-07
No unused props eslint rule

devinmatte:
## Motivation

Ensure we don't have any unused props on components

## Changes

- Adds `react/no-unused-prop-types` as a warn

## Testing Instructions

Remove all usage of a prop, ensure that the interface line is now a warning for that component


idreyn (approved):
LGTM and I would support making this an error unless there are edge cases it doesn't catch (spreading props into children, maybe?)

PatrickCleary (approved):

PR #467, opened 2023-04-08
Better opacity scaling for SZ map segments

idreyn:
@austinjpaul proposed using a linear scale on slow time instead of % slowdown. I think this does a much better job at emphasizing the problem areas. Thoughts?

_Before:_
Screenshot 2023-04-08 at 14-19-08 Screenshot

_After:_
Screenshot 2023-04-08 at 14-19-19 Screenshot


PatrickCleary (approved):
I like it

devinmatte (approved):

austinjpaul:
I wanted to play with a square root curve as well to give more variation in the low end, while retaining visible differences in the high end. 
Here's `return Math.min(Math.sqrt(zone.delay / 240), 0.9);`
image

I think something like this might be an improvement- even the 30s zones are dark enough to stand out, and the worst regions are even darker. Thoughts?

skaplan-dev (approved):

devinmatte (approved):

PR #469, opened 2023-04-10
API Cleanup and Ridership fixes

devinmatte:
## Motivation

We want to reduce API calls, especially when data has not changed since last fetch

## Changes

- Moves API calls into reusable hooks and defaults reasonable staleTimes
- Adds the ridership dayselector
Screenshot 2023-04-10 at 1 59 52 PM


skaplan-dev (approved):
🔥

PR #470, opened 2023-04-10
Updating python deps for April 2023

devinmatte:
## Motivation

Updating python deps up to latest, along with https://github.com/transitmatters/slow-zones/pull/26 and https://github.com/transitmatters/data-ingestion/pull/9

## Changes

Updates packages with new minor or patch updates, avoid major releases (see https://github.com/transitmatters/t-performance-dash/pull/451)

### DIFF
--- a/server/poetry.lock
+++ b/server/poetry.lock
@@ -7,22 +7,8 @@ optional = false
 python-versions = "*"
 
 [[package]]
-name = "attrs"
-version = "21.4.0"
-description = "Classes Without Boilerplate"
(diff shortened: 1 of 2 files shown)


devinmatte:
As a note, we are running with a lot of these versions in beta v4 with no issues. Could probably backport a few of these package bumps to v3 directly, so we don't have a large rift in case we need to maintain an instance of v3 for a while

PatrickCleary (approved):

PR #472, opened 2023-04-11
V4/date preset logic

PatrickCleary:
image

## Motivation

Our dashboards were set up to remember date params across sections, but not date presets. So if you returned to a section it would always say `custom` as the range.

Also, navigating to a URL with a date param matching a preset will now apply that preset to the date picker.

Lastly, improvements to range/single trip date swap.

## Changes

Adds `datePresetConfig` zustand state management. Keeps track of the preset for date picker across sections.

## Testing Instructions

1. From `overview` page, select preset then navigate to any subpage. Preset will stick.
2. From any `line` subpage, set preset date, navigate to new section, and return. Preset will stick. Same with `trips` section.
3. Navigate via URL to any page with a preset date range, like http://localhost:3000/red/slowzones/?endDate=2022-12-31&startDate=2022-01-01 you will see that the preset is automatically applied.
4. Switch back and forth between single/range tabs on `trips` page. Verify logic is intuitive.
5.  Try anything else that you think should work. Hopefully covered all cases, but please try to find edge cases.


common/constants/dates.ts
PatrickCleary, line 1: These keys are used to test whether the incoming URL is a date preset on first load.

common/layouts/DashboardLayout.tsx
PatrickCleary, line 1: Add logic to test whether incoming URL is an existing date preset, and apply it if so.

common/state/datePresetConfig.ts
```diff
@@ -0,0 +1,51 @@
...
+        return PRESET_DEFAULTS;
+      }
+      if (section === 'trips') {
+        if (query.queryType === 'range') PRESET_DEFAULTS.rangeTripPreset = preset;
+        else PRESET_DEFAULTS.singleTripPreset = preset;
+        return PRESET_DEFAULTS;
+      }
+      return PRESET_DEFAULTS;
+    }),
+
```
PatrickCleary, line 42: Set default preset configurations on first load. The `preset` would be the preset for the page the user navigated to.

modules/navigation/utils/rangeTabUtils.ts
PatrickCleary, line 1: Handles all logic for switching between `range` and `single day` tabs (which I renamed `Aggregate` and `Daily`)

Definitely open to suggestions on those names.

common/constants/dates.ts
devinmatte: I see we only do it on first load. So if someone manually sets the range to a set of values that matches a preset it still says custom. This is fine, just a note of something I noticed

devinmatte (approved):
✅ Usage test, seems great
Code looks good

Only one non-blocking comment

common/constants/dates.ts
PatrickCleary: That should be how it works. Was it not working?

PatrickCleary: I'm gonna merge this because I think this is how it works now. Lmk if you find any cases where it's not the case @devinmatte

PR #473, opened 2023-04-11
V4/datepicker change

PatrickCleary:
## Motivation

In preparation for #458 we need single day date pickers, range date pickers, and a combo. This reworks the date picker to have all three.

## Changes

Adds a `type` field to the DateSelection component which can be `combo`, `range`, or `single`. In this PR, we are only using `combo`. In upcoming changes we will use `range` and `single`.
**range**
Screen Shot 2023-04-11 at 10 18 12 AM

**Single**
Screen Shot 2023-04-11 at 10 18 31 AM

**Combo** (already exists)
Screen Shot 2023-04-11 at 10 21 23 AM

## Testing Instructions
Test the existing date picker, and go into the code to change the `type` field and test it after.

### DIFF
--- a/common/components/inputs/DateSelection/DatePickers.tsx
+++ b/common/components/inputs/DateSelection/DatePickers.tsx
@@ -2,71 +2,69 @@ import dayjs from 'dayjs';
 import type { SetStateAction } from 'react';
 import React from 'react';
 import classNames from 'classnames';
-import { faClose, faArrowRight } from '@fortawesome/free-solid-svg-icons';
+import { faArrowRight, faClose } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import Flatpickr from 'react-flatpickr';
 import 'flatpickr/dist/themes/light.css';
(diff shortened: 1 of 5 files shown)


common/components/inputs/DateSelection/DatePickers.tsx
```diff
@@ -2,71 +2,69 @@ import dayjs from 'dayjs';
...
-    setConfig({ ...config, selection: undefined });
+    setSelection(undefined);
   };
 
   const handleStartDateChange = (date: string) => {
     const updatedDate = dayjs(date);
     if (updatedDate.isSame(endDateObject)) {
-      // Convert to single day mode when startDate == endDate
-      handleRangeToggle();
+      // TODO: set to undefined?
```
devinmatte, line 58: Any reason we don't handle this anymore?

devinmatte (commented):
One question before I do a full review. Looks good, I just need to test before approval

common/components/inputs/DateSelection/DatePickers.tsx
PatrickCleary: Ah, good catch. That was an accident.

PR #479, opened 2023-04-12
V4/various fixes

PatrickCleary:
## Motivation

Minor changes that will simplify the upcoming nav-logic PR.

## Changes
1) Rename things
2) 

Add a condition to this useEffect to check if the query already contains the `to`/`from` parameters. fixes issue where it gets reset upon switching between `trips` tabs
```
useEffect(() => {
    if (!from) setFromStation(stations?.[3]);
    if (!to) setToStation(stations?.[stations.length - 3]);
  }, [lineShort, from, to, stations, setFromStation, setToStation]);
  ```
  

## Testing Instructions
switch between tt/hways/dwells and the station widget won't change values.


devinmatte:
Looks like `/trips/` ignores the `to` and `from` params
I'm also noticing when I pick `Alewife` and refresh, it switches to `Davis`

Are these expected to be fixed in this PR?

devinmatte (commented):
Code looks good, still experiencing the bugs I mentioned above. The weirdest one is the Alewife to Davis (and Oak Grove to Malden Center so it's a terminus issue)

PR #480, opened 2023-04-12
V4/context dashboard

PatrickCleary:
WIP


common/components/inputs/DateSelection/DateSelection.tsx
```diff
@@ -20,42 +20,34 @@ interface DateSelectionProps {
...
+  const { section } = ALL_PAGES[page];
   const handleSelection = (selection: number, range: boolean) => {
-    const newOptions = range ? DATE_PICKER_PRESETS.range : DATE_PICKER_PRESETS.singleDay;
+    const newOptions = range ? Object.values(RANGE_PRESETS) : Object.values(SINGLE_PRESETS);
     updateQueryParams(newOptions[selection].input ?? null, range);
     setRange(range);
     setSelection(selection);
   };
 
-  /*
```
PatrickCleary, line 43: This comment explains what this `useEffect` used to do. I removed it because my next change will do the same thbing, and more.

Now this useEffect is only used to set the `range` value appropriately from the props since they come from the URL params and can take a second to load.

common/constants/dashboardTabs.ts
```diff
@@ -1,7 +1,16 @@
...
+export const DASHBOARD_TABS: {
+  [key in Tab]: { name: Tab; path: string; disabled?: boolean; query?: QueryParams };
+} = {
   Subway: { name: 'Subway', path: '/red' },
-  Bus: { name: 'Bus', path: '/bus/trips?busRoute=1' },
+  Bus: {
+    name: 'Bus',
+    path: '/bus/trips',
+    query: { busRoute: '1', queryType: 'single', startDate: BUS_DEFAULTS.tripConfig.startDate },
+  },
```
PatrickCleary, line 14: Just reformatting these.

common/constants/dates.ts
PatrickCleary, line 1: 2 changes

1) added a key and key field to the preset objects. Makes it easier to work with them.
2) make a type which is the union of all these keys.

common/state/dashboardConfig.ts
PatrickCleary, line 1: This is the meat of the changes. 

Zustand state manager for dashboard config. This only stores dates for now.

Split up into 3 pieces of state
1. Overview - just needs to store current tab (view)
2. Line - stores the start and end dates
3. Trip - stores the start and end dates (Separately from `Line`)


And 4 functions:
1. `setLineConfig`: updates line date values
2. `setTripConfig`: updates trip date values. 
3. `swapDashboardTabs`: this lets us switch out all the state when switching between bus and subway.
4. `overviewPresetChanges`: When the user chooses a preset on the `overview` page, we want that date range to be applied to any of the subpages they go to. This function adjusts those values accordingly.

common/utils/router.tsx
```diff
@@ -104,12 +105,13 @@ export const getLineSelectionItemHref = (newLine: Line, route: Route): string =>
...
   const currentPath = currentPage.path;
   let href = `/${path}`;
-
   // Go to homepage if current line is selected or the selected page is not valid for the given line.
   const validPage = currentPage.lines.includes(newLine);
   if (key === line || currentPath === 'today' || !validPage) {
     return href;
   }
+  delete query.from;
+  delete query.to;
```
PatrickCleary, line 114: We want to remove station URL params when switching lines,

modules/speed/SpeedWidget.tsx
PatrickCleary, line 1: Was using a default here, now can take the `view` param

modules/speed/SpeedGraph.tsx
PatrickCleary, line 1: Was using a default here, now can take the view param

modules/navigation/DashboardSelection.tsx
```diff
@@ -4,16 +4,19 @@ import classNames from 'classnames';
...
   const router = useRouter();
+  const swapDashboardTabs = useDashboardConfig((state) => state.swapDashboardTabs);
   const dashboardTabs = Object.values(DASHBOARD_TABS);
   return (
     <Tab.Group
       manual
       onChange={(index) => {
-        router.push(dashboardTabs[index].path);
+        router.push({ pathname: dashboardTabs[index].path, query: dashboardTabs[index].query });
+        swapDashboardTabs(dashboardTabs[index].name);
```
PatrickCleary, line 19: Swaps out current config with defaults for new tab (bus/subway)

modules/dashboard/HomescreenWidgetTitle.tsx
PatrickCleary, line 1: Rework to use new `handlePageNavigation` rather than passing `hrefs`. More dynamic

common/utils/router.tsx
```diff
@@ -136,19 +142,44 @@ export const getBusRouteSelectionItemHref = (newRoute: string, route: Route): st
...
+        navigateWithinSection(router, linePath, page, query);
+        return;
+      }
+      saveDashboardConfig(pageObject.section, query, dashboardConfig);
+      navigateToNewSection(router, linePath, page, query, dashboardConfig);
+    },
+    [router, query, linePath, pageObject, dashboardConfig]
+  );
+  return handlePageNavigation;
+};
```
PatrickCleary, line 163: This function replaces `handleTabNavigation`. 

The additional functionality is to save the current configuration when exiting a section, and bringing in the previous confguration when returning to a previous section.

This also serves to inject our default values, because those defaults are in the configuration.

common/state/utils/utils.ts
PatrickCleary, line 1: Functions used in router to save/load config  when swapping sections.

common/constants/dashboardTabs.ts
PatrickCleary, line 1: Changes:
Reformat `DASHBOARD_TABS`
Reformat presets to have a `key` and `key` field
Add `DatePresetKey` type.

common/components/inputs/DateSelection/OverviewDateSelection.tsx
```diff
@@ -1,13 +1,25 @@
...
+  const router = useRouter();
+  const selectedView = router.query.view ?? 'year';
+  const selectedIndex = Object.keys(OverviewRangeTypes).findIndex((view) => view === selectedView);
+
+  const overviewPresetChange = useDashboardConfig((state) => state.overviewPresetChange);
+  const handlePresetSelection = (value: OverviewDatePresetKey) => {
+    overviewPresetChange({ view: value });
+    router.query.view = value;
+    router.push({ pathname: router.pathname, query: router.query });
   };
```
PatrickCleary, line 18: When preset is changed, update `line` config, add `view` to query, push new URL to router.

common/state/dashboardConfig.ts
PatrickCleary, line 1: This is the meat of the changes.

Zustand state manager for dashboard config. This only stores dates for now.

Split up into 3 pieces of state

Overview - just needs to store current tab (view)
Line - stores the start and end dates
Trip - stores the start and end dates (Separately from Line)
And 4 functions:

setLineConfig: updates line date values
setTripConfig: updates trip date values.
swapDashboardTabs: this lets us switch out all the state when switching between bus and subway.
overviewPresetChanges: When the user chooses a preset on the overview page, we want that date range to be applied to any of the subpages they go to. This function adjusts those values accordingly.

devinmatte:
I noticed right now if I select a range `say 2023` on slow zones, switch to travel times, pick `2022` and then go back to slow zones, it's still set to 2023, and back to travel times still 2022. To me, I was expecting it to update both.
I don't think this is a bug or worth fixing now, but just something to consider and maybe discuss with people when we start to elicit feedback. Should range be shared across all pages? Or not?

PatrickCleary:
@devinmatte 

I separated the date ranges for two reasons:
1. Date ranges over 18 months don't work on `trips` (and anything over a month is quite slow)
2. Single day date ranges don't make sense in `lines`

I also think they should have separate preset options for those reasons.

I could see why you would expect them to both update though... Perhaps we need a better visual indicator? I was hoping the section separation might suffice, but maybe not. Maybe a slight visual difference on the pickers?

devinmatte:
@PatrickCleary Yeah I totally see both sides, which is why this might just come down to user feedback. I think it's fine as is today, as it's still good, just not what I expected

PR #482, opened 2023-04-13
Fix Bus trip explorer page

### LINKED ISSUE #452: Classic View for Buses

PatrickCleary:
## Motivation

fixes #452 

## Changes

1. Main issue is that the trip explorer was not getting passed the `busRoute`. Fixed that

2. We also had no way to differentiate between bus and subway in `useCustomQueries` so I reworked that to take `BusOrSubway` as a parameter so we're not searching for non-existent bus dwell data.

3. Separated bus and subway trip graphs into two components.

## Testing Instructions

Go to bus trip explorer page without `from`/`to` in the URL. This didn't work previously, now it does.

### DIFF
--- a/common/api/datadashboard.ts
+++ b/common/api/datadashboard.ts
@@ -1,13 +1,14 @@
 import type * as ReactQuery from '@tanstack/react-query';
 import { useQueries } from '@tanstack/react-query';
+import { QUERIES, AggregateAPIParams, QueryNameKeys, SingleDayAPIParams } from '../types/api';
 import type {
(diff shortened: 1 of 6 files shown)


PatrickCleary:
Should be fixed by  #480 @devinmatte

devinmatte:
All good, I'll wait until this is rebased ontop of #480 before approving though to check

devinmatte (approved):

PR #491, opened 2023-04-13
[v4] Notice on Speed when shuttling

devinmatte:
## Motivation

Show a basic message when the line is shuttling and we don't have a MPH number to make it clear why

## Changes

Screenshot 2023-04-13 at 2 00 50 PM

- Only shows when line is currently shuttling and we don't have a computed mph number

## Testing Instructions

Hard to test when there isn't actually a shuttle running on a line at the time

### DIFF
--- a/common/utils/emoji.ts
+++ b/common/utils/emoji.ts
@@ -1,0 +1,21 @@
+export const randomUpsetEmoji = () => {
+  const emojis = [
+    '😭',
+    '😢',
+    '😔',
+    '😡',
+    '😞',
+    '😕',
+    '🙁',
+    '😠',
+    '😑',
+    '🤬',
+    '😬',
+    '😨',
+    '😯',
+    '😮',
+    '😳',
+  ];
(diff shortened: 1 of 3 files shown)


PatrickCleary:
Looks good. I really like the idea of showing the alert in the widget. Maybe if we have that, we should use a different emoji above? Maybe a sad face? I just used a <p> tag to make this but if you like the idea, I can send an svg of it.
Screen Shot 2023-04-14 at 10 08 57 AM


Also, we might want to check for suspension alerts, and maybe a general case because there's probably other reasons the line is not fully running that we're not accounting for.

devinmatte:
Yeah we can do a sad face, then we just put whatever the responsible alert is below.

We probably just use a `<p>` tag with the emoji value instead of svg's tbh. That way it uses the devices default emojiset

devinmatte:
Screenshot 2023-04-14 at 12 10 27 PM
Handling suspensions, and, emojis!

PR #492, opened 2023-04-13
Station can sometimes be undefined

devinmatte:
## Motivation

Screenshot 2023-04-13 at 2 45 42 PM
Screenshot 2023-04-13 at 2 46 07 PM

## Changes

Reflects the reality, that `stations` can be a list of `(Station | undefined)[]`. We probably should ensure that can't happen, but, in the meantime, that is the reality and we need to handle it

### DIFF
--- a/modules/commute/alerts/AlertUtils.ts
+++ b/modules/commute/alerts/AlertUtils.ts
@@ -2,17 +2,19 @@ import { stations } from '../../../common/constants/stations';
 import type { BusRoute, LineShort } from '../../../common/types/lines';
 import type { Station } from '../../../common/types/stations';
 
-export const getEndStations = (stations: Station[]) => {
+export const getEndStations = (stations: (Station | undefined)[]) => {
   let min = stations[0];
   let max = stations[0];
 
   stations.forEach((station) => {
-    if (station.station === 'place-jfk') station.stop_name = 'JFK/UMass'; // Sub out branch names of JFK
-    if (station.order < min.order) {
-      min = station;
-    }
-    if (station.order > max.order) {
-      max = station;
+    if (station) {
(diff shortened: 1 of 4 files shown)


PatrickCleary:
Is this because the alerts api was sending no stations? If so, I think that may just be part of the API and we might need to account for it by rewording the alert widget.

devinmatte:
It may be yes, I just wanted to address the symptom and type things accurately to ensure we catch potential issues. This handles at least this specific issue

PatrickCleary:
Thanks for tracking this down. I think the core issue is stemming from the `getStations` function being called when it shouldn't.

In `DelayAlerts` and `SuspensionAlerts` there is this check to make sure stops is not empty:
https://github.com/transitmatters/t-performance-dash/blob/dashboard-v4/modules/commute/alerts/DelayAlert.tsx#L18-L19

But it's missing in `ShuttleAlerts`: 
https://github.com/transitmatters/t-performance-dash/blob/dashboard-v4/modules/commute/alerts/ShuttleAlert.tsx#L15

I think that might be where this is coming from. I think when I wrote this, I assumed shuttling would always have stops. But I'm guessing yesterday's blue line shuttling was the entire line and so they didn't input stops.

We might want to add that check around `getStations` rather than allow Stations to be undefined.

PatrickCleary:
Oh I just realized that what I said above doesn't really make sense. Station array is not undefined, just the entries, as you noted.

When you made this change, what did the alert look like? Just empty strings where stops should go I presume?

That will probably make it easy to spot next time and then try to debug the backend.

PatrickCleary (approved):
lgtm. I realize now it's not related, but we should probably check for empty stations array in shuttling component.

PR #495, opened 2023-04-14
Updating package.json/lock to satisfy dependabot

devinmatte:
## Motivation

We have an open PR for https://github.com/transitmatters/t-performance-dash/pull/385

Although we're working on v4, we're probably going to keep an instance of v3 running for a little bit at launch, and, it's deployed today, so we should ensure there aren't any easily fixed vulnerabilities

## Changes

- Bumps minimum node version to be accurate (lock is generated with node 16.x, and CI runs on 16.x, we should require 16.x)
- Fixes placement of a dev dep
- Does basic lock updates and `audit fix`

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This is the repository for the TransitMatters Data Dashboard. Client code is wri
 
 ## Requirements to develop locally
 
-- node 16.x preferred, but should work with anything >=14.x and <17. (verify with `node -v`)
+- node 16.x required, works with anything >=16.x and <17. (verify with `node -v`)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -9,7 +9,7 @@ This is the repository for the TransitMatters Data Dashboard. Client code is wri
 
 ## Requirements to develop locally
 
-- node 16.x preferred, but should work with anything >=14.x and <17. (verify with `node -v`)
+- node 16.x required, works with anything >=16.x and <17. (verify with `node -v`)
```
nathan-weinberg, line 12: We love a documentation update :ok_hand:

nathan-weinberg (approved):

PR #502, opened 2023-04-18
Updating to allow Python 3.10

devinmatte:
## Motivation

Allow Python 3.10

## Changes

- Sets poetry to use python 3.10
- Runs CI on python 3.10
- Sets up matrix for potential two version CI

## Testing Instructions

Once merged we can update lambda in the aws console to use the newer runtime


.github/workflows/build.yml
```diff
@@ -9,17 +9,20 @@ on:
...
         uses: actions/setup-python@v3
         with:
-          python-version: 3.9
-      - name: Set up Node 18.x
+          python-version: '3.10'
+      - name: Set up Node ${{ matrix.node-version }}.x
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: ${{ matrix.node-version }}
```
nathan-weinberg, line 25: I like using `matrix` for node - why not do it for Python as well?

devinmatte: We can do it in both, I set it for the jobs that are specific to backend/frontend. I can add it everywhere though

nathan-weinberg (approved):
LGTM

PR #508, opened 2023-04-21
V4/service page

PatrickCleary:
## Motivation

We now have scheduled service data!

## Changes

1. Add service widget and service page.
2. Add date placeholder to `Line` page widgets
3. Updated speed widgets -> comparison to peak of date range rather than first day of range.

I think the graph is a little squished, but not gonna work on that now because that should be coordinated across the graphs I think.

And the widgets are not set in stone. Just some stuff I threw on to get started. This PR is mostly about getting the page on there and setting up the API route.

(Also changed the speed graph slightly - uses steps like this new service graph)
Screen Shot 2023-05-05 at 11 19 33 AM

Screen Shot 2023-05-05 at 11 19 46 AM


devinmatte:
Question? Do we have the data for "scheduled service levels"? So we can compare against a baseline of what the T says they're going to do?

PatrickCleary:
I think Ian is working on a GTFS ingestor which will get that information.

devinmatte:
We should move the Service Levels graph from the ridership page to this page. Ridership should show just that, ridership stats

modules/service/utils/utils.ts
PatrickCleary, line 1: This is all basically the same as the Speed utils

modules/speed/SpeedGraph.tsx
PatrickCleary, line 1: I like this more - makes it easier to see small speed increases.

modules/service/utils/utils.ts
skaplan-dev, line 1: this can be condensed
```
import type { SpeedDataPoint } from '../../../common/types/dataPoints';

const calcValues = (datapoints: SpeedDataPoint[], includePeak: boolean = false) => {
  const current = datapoints[datapoints.length - 1].count / 2;
  const delta = current - datapoints[0].count / 2;
  const average = datapoints.reduce((sum, speed) => sum + speed.count, 0) / datapoints.length / 2;
  const peak = includePeak ? { ...datapoints.reduce((max, speed) => speed.count > max.count ? speed : max, datapoints[0]), value: max.count / 2 } : undefined;
  
  return { current, delta, average, ...(includePeak && { peak }) };
};

export const getOverviewServiceWidgetValues = (datapoints: SpeedDataPoint[]): WidgetValues => calcValues(datapoints);
export const getDetailsServiceWidgetValues = (datapoints: SpeedDataPoint[]): WidgetValues => calcValues(datapoints, true);
```

skaplan-dev: we should combine them!

PatrickCleary: The functions are similar, but the data types are different which makes it hard. They're also starting to diverge as I adjust the widgets. I condensed them both though, per your suggestion.

devinmatte (approved):

modules/service/utils/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export const getOverviewServiceWidgetValues = (
+  datapoints: SpeedDataPoint[],
+  predictedData: number[]
+) => calcValues(datapoints, predictedData);
+export const getDetailsServiceWidgetValues = (
+  datapoints: SpeedDataPoint[],
+  predictedData: number[]
+) => calcValues(datapoints, predictedData);
```
skaplan-dev, line 36: is there a difference between these?

skaplan-dev (approved):

modules/service/utils/utils.ts
PatrickCleary: Not anymore, will remove one

PR #511, opened 2023-04-25
V4/szs nb sb

PatrickCleary:
This is currently live on the [V4 URL](https://dashboard-v4-beta.labs.transitmatters.org/red/slowzones/?startDate=2016-02-01&endDate=2023-04-26) if you want to check it out without setting it up.

## Motivation

1. We needed NB/SB selector SZ segments.
2. The Segment view had a couple issues, and weird UI in certain cases. Particularly with mobile.

## Changes

1. Add NB/SB selector
2. Standardize bar width. Previously the height of the chart was constant and the bars would squish accordingly.
3. Use shortened station names when screen size is limited
4. Add opacity to the bars - same logic as slow zone map
5. remove labels when they do not fit in the box. Had to do this based on the SZ time range, not the bar width. I tried to use bar width for awhile but was not having any luck.
Screen Shot 2023-04-26 at 6 43 42 PM

## Testing Instructions

Try out all configurations and screen sizes related to the slow zone line segments graph.


common/types/dataPoints.ts
```diff
@@ -48,13 +49,11 @@ export interface SlowZone {
   duration: number;
   delay: number;
   color: string;
-  fr_id: string;
-  to_id: string;
```
PatrickCleary, line 52: These weren't being used

modules/slowzones/TotalSlowTimeWrapper.tsx
```diff
@@ -32,7 +32,7 @@ export const TotalSlowTimeWrapper: React.FC<TotalSlowTimeWrapperProps> = ({
       <BasicWidgetDataLayout
         widgetValue={
           new TimeWidgetValue(
-            filteredDelayTotals[filteredDelayTotals.length - 1][lineShort],
+            filteredDelayTotals[filteredDelayTotals.length - 1]?.[lineShort],
```
PatrickCleary, line 35: This was causing issues when the sz data was not up to date.

modules/slowzones/charts/LineSegments.tsx
PatrickCleary, line 1: Most of the complication in these changes was to make the `x` and `y` axes switch between at `sm` screen breakpoint

devinmatte:
Screenshot 2023-04-28 at 3 48 03 PM
There's extra padding on the right when on mobile compared to the left

Screenshot 2023-04-28 at 3 48 28 PM
Also there's some less than ideal extra spacing on mobile with the zone names, not sure of a better way to deal with it though, so, probably okay

common/utils/slowZoneUtils.ts
```diff
@@ -30,27 +31,31 @@ const getDirection = (to: Station, from: Station) => {
...
-      start: x.start,
-      end: x.end,
-      from: from.stop_name,
-      to: to.stop_name,
-      id: from.stop_name + '-' + to.stop_name,
-      delay: +x.delay,
-      duration: +x.duration,
-      color: x.color,
-      fr_id: x.fr_id,
-      to_id: x.to_id,
```
devinmatte, line 49: You said above these aren't used, but looks like they were used here?

modules/slowzones/constants/chartConfig.ts
```diff
@@ -0,0 +1,22 @@
+import type dayjs from 'dayjs';
+import { enUS } from 'date-fns/locale';
+import type { CartesianScaleTypeRegistry, ScaleOptionsByType } from 'chart.js';
+import { getTimeUnitSlowzones } from '../../../common/utils/slowZoneUtils';
+
+export const getDateAxisConfig = (startDateUTC: dayjs.Dayjs, endDateUTC: dayjs.Dayjs) => {
```
devinmatte, line 6: This feels like a util, not a constant

devinmatte (approved):
Looks good to me overall, small questions?

common/utils/slowZoneUtils.ts
PatrickCleary: I just mean they weren't used after being assigned here.

PR #513, opened 2023-04-25
Shortened station names

PatrickCleary:
## Motivation
In certain places, we need shorter station names. This is a first draft on some of these names.
## Changes
added "short" field to stations.json
added short field for red, orange, and blue lines


devinmatte (approved):

devinmatte:
We probably should update the `Station` ts type to have an optional `short`

PR #515, opened 2023-04-28
Fixing deployments

devinmatte:
## Motivation

Python 3.10 deployments were failing, reverting back to 3.9 in the meantime

## Changes

- Sets some logical minimums in poetry
- Updating nodejs minimum since CI runs on 18.16.0
- Adding nextjs build cache to speed up builds


.github/workflows/deploy.yml
```diff
@@ -45,4 +55,5 @@ jobs:
         run: |
           curl -sSL https://install.python-poetry.org | python3 -
           npm ci
+          aws --version
```
devinmatte, line 58: Temporary for testing

server/pyproject.toml
```diff
@@ -15,6 +15,8 @@ numpy = "1.24.2"
 pandas = "1.5.3"
 
 [tool.poetry.dev-dependencies]
+pip = "^23.0"
+setuptools = "^67.6.1"
```
devinmatte, line 19: To potentially get rid of an error message in the logs

nathan-weinberg (approved):

PR #519, opened 2023-04-29
Labeler update

devinmatte:
## Motivation

Didn't like Javascript/Python labels, preferred `frontend` / `backend`

## Changes

Using `frontend` and `backend` labels


nathan-weinberg (commented):
One nit, could we change the action name from `Labeler` to `labeler` so it matches the case of the others?

nathan-weinberg (approved):

Issue #471, opened 2023-04-10
Display properly when station pair returns no data


devinmatte:
**In v3 we showed**
<img width="1035" alt="Screenshot 2023-04-10 at 5 30 39 PM" src="https://user-images.githubusercontent.com/9310513/231002870-557f7858-1bca-4e4a-8885-aeb01b8a6af3.png">

In v4 we currently show
<img width="638" alt="Screenshot 2023-04-10 at 5 30 57 PM" src="https://user-images.githubusercontent.com/9310513/231002916-3613219f-d36d-481f-8f9b-f4e8ecf7ff68.png">

We should be clearer that there's no data to show

PatrickCleary:
As a note for whoever is working on this, I _believe_ that the code to display this message is still here, but in most of our charts we are passing empty arrays which do not show this message. I believe it needs to receive undefined.

Another note is that some charts are using the `ChartPlaceholder` component and some are not. I believe the ones that are using the component would show an error message in this case. However if we want to differentiate between error and no data we could add some fairly simple logic to do so.

PR #521, opened 2023-04-30
V4 Missing Data Message on Charts

devinmatte:
## Motivation

#471 

## Changes

Screenshot 2023-04-30 at 12 15 07 PM

### DIFF
--- a/common/components/charts/AggregateLineChart.tsx
+++ b/common/components/charts/AggregateLineChart.tsx
@@ -18,6 +18,7 @@ import type { AggregateDataPoint, AggregateLineProps } from '../../types/charts'
 import { prettyDate } from '../../utils/date';
 import { CHART_COLORS } from '../../../common/constants/colors';
 import { DownloadButton } from '../general/DownloadButton';
+import { writeError } from '../../utils/chartError';
 import { LegendLongTerm } from './Legend';
 import { drawTitle } from './Title';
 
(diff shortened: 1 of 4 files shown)


common/components/charts/AggregateLineChart.tsx
```diff
@@ -178,6 +179,9 @@ export const AggregateLineChart: React.FC<AggregateLineProps> = ({
             {
               id: 'customTitleAggregate',
               afterDraw: (chart: ChartJS) => {
+                if (startDate === undefined || endDate === undefined || data.length === 0) {
+                  writeError(chart);
```
PatrickCleary, line 183: [This upcoming PR](https://github.com/transitmatters/t-performance-dash/pull/514) should prevent this component from being loaded without a startDate or endDate, so I think we only need to check for data being an empty array

common/components/charts/SingleDayLineChart.tsx
```diff
@@ -216,25 +215,13 @@ export const SingleDayLineChart: React.FC<SingleDayLineProps> = ({
...
-                  const { height } = chart;
-                  chart.clear();
-
-                  ctx.save();
-                  ctx.textAlign = 'center';
-                  ctx.textBaseline = 'middle';
-                  ctx.font = "16px normal 'Helvetica Nueue'";
-                  ctx.fillText('No data to display', width / 2, height / 2);
-                  ctx.restore();
+                if (date === undefined || date.length === 0 || data.length === 0) {
```
PatrickCleary, line 223: Was there a case when we were passing an empty string for date? Or is this just out of precaution?

PatrickCleary (approved):
LGTM

common/components/charts/SingleDayLineChart.tsx
devinmatte: There was in v3, so this was handling those cases from there. It doesn't hurt to check

Issue #380, opened 2023-03-10
Wire up Datadog to Dashboard V4


devinmatte:
We have access to a Datadog account now. We should wire up RUM, APM and logs to the V4 dashboard

PR #522, opened 2023-04-30
Wire up Datadog Tracing

devinmatte:
## Motivation

#380 

Give us more detailed information about the performance of the app

## Changes

- Adds datadog lambda tracing to chalice

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -90,7 +90,8 @@ aws cloudformation deploy --template-file cfn/packaged.yaml --stack-name $CF_STA
     TMFrontendHostname=$FRONTEND_HOSTNAME \
     TMFrontendCertArn=$FRONTEND_CERT_ARN \
     TMBackendCertArn=$BACKEND_CERT_ARN \
-    MbtaV2ApiKey=$MBTA_V2_API_KEY
+    MbtaV2ApiKey=$MBTA_V2_API_KEY \
+    DDApiKey=$DD_API_KEY
 
 popd > /dev/null
 aws s3 sync out/ s3://$FRONTEND_HOSTNAME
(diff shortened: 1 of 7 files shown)


server/frontend-cfn.json
```diff
@@ -80,7 +92,10 @@
       "Properties": {
         "Environment": {
           "Variables": {
-            "MBTA_V2_API_KEY": { "Ref": "MbtaV2ApiKey" }
+            "MBTA_V2_API_KEY": { "Ref": "MbtaV2ApiKey" },
+            "DD_SITE": { "Ref": "DDSite" },
```
mathcolo, line 96: How likely is it that DD_SITE will change? perhaps we can just hardcode that?

deploy.sh
```diff
@@ -90,7 +90,8 @@ aws cloudformation deploy --template-file cfn/packaged.yaml --stack-name $CF_STA
     TMFrontendHostname=$FRONTEND_HOSTNAME \
     TMFrontendCertArn=$FRONTEND_CERT_ARN \
     TMBackendCertArn=$BACKEND_CERT_ARN \
-    MbtaV2ApiKey=$MBTA_V2_API_KEY
+    MbtaV2ApiKey=$MBTA_V2_API_KEY \
+    DDApiKey=$DD_API_KEY
```
mathcolo, line 94: Missing DD_SITE here

server/frontend-cfn.json
mathcolo: (and if you choose to do so, vars that are static can be hard-coded here https://github.com/transitmatters/t-performance-dash/blob/dashboard-v4/server/.chalice/config.json#L13)

mathcolo (commented):
exceedingly close

server/frontend-cfn.json
devinmatte: Likely to never change, I'll do that then

mathcolo (approved):

PR #531, opened 2023-05-02
SZT speed restrictions take 2

idreyn:
#444 was languishing forever and had gotten way out of sync — in the end it was easiest to branch off of `dashboard-v4` and start from scratch. See that PR for context, details, etc.


devinmatte (approved):
Looking good, thanks for handling this

Let's get this in before any more conflicts can occur

PR #535, opened 2023-05-02
Adding favicon, meta and titles

devinmatte:
## Motivation

Display a title, TM logo and metadata for unfurls on twitter and more

## Changes

- Adds a title to every page
- Adds metadata to actually rendered document
- Updates favicon to use the TM one, and not the vercel one


PatrickCleary:
Would it be easier to wrap the whole app in `PageWrapper` and get the title from the route?

skaplan-dev:
> Would it be easier to wrap the whole app in `PageWrapper` and get the title from the route?

+1. Is there a way to do this more programmatically with `next`?

devinmatte:
This seems to be Next's recommended way of handling this (per route) when I was reading docs. I couldn't find a more programmatic way to handle things

PatrickCleary (approved):
Looks good. The icon really makes it look so much better


just one nitpick, I see a lot sites go in sort of a reverse breadcrumb order  (and use `|` as separation). So would be like `Slow Zones | Data Dashboard`. Could also do `Slow Zones | Red Line` to add even more detail.

couple examples from sites I happen to have open rn:
Screen Shot 2023-05-03 at 2 09 41 PM

devinmatte:
Yeah I kinda agree, I'll go update the order

PR #536, opened 2023-05-03
V4/null speed values

PatrickCleary:
**Specific testing instructions below**

## Motivation

Previously, the speed database did not have values for dates which were considered invalid (any day with shuttling or missing data). I am updating the db to have these values (for the service graph), and setting the `value` to null.  The frontend needed some updates to handle this better.

I also updated the graph to leave empty dates blank, rather than fill in the line. I think this is better for 2 reasons:
1) Easy way to see when there was shuttling in the past
2) Less misleading.

## Changes

Screen Shot 2023-05-02 at 6 33 24 PM

## Testing Instructions

I have updated the speed database to have null values only for the **red line in 2023**. So look at days in those values on the graph to sees how it has changed.


modules/speed/SpeedOverviewWrapper.tsx
```diff
@@ -26,12 +26,12 @@ export const SpeedGraphWrapper: React.FC<TotalSlowTimeWrapperProps> = ({
   endDate,
 }) => {
   const { current, delta, average } = getOverviewSpeedWidgetValues(data, line);
-
+  const dataNoNulls = data.filter((datapoint) => datapoint.value !== undefined);
```
devinmatte, line 29: `undefined` and `null` are different types. Just want to make sure the right things are being checked here

modules/commute/speed/utils/utils.ts
```diff
@@ -6,8 +6,9 @@ export const calculateCommuteSpeedWidgetValues = (
   speed: SpeedDataPoint[],
   line: string
 ) => {
+  const weeklyDataNoNulls = weeklyData.filter((datapoint) => datapoint.value != null);
```
devinmatte, line 9: We should be using `!==` here

devinmatte (approved):
Looks good, better than infinity lol

PR #538, opened 2023-05-03
API routes for service levels and ridership

idreyn:
## Motivation

For the Service and Ridership pages and related widgets, we want to use the latest data that we've begun ingesting into dynamoDB over in the [`data-ingestion`](https://github.com/transitmatters/data-ingestion) repo.

## Changes

This PR adds two new API routes that query the `TripCounts` and `Ridership` tables:

### `/api/tripcounts`

Accepts `start_date`, `end_date`, and `route_id` GET params and returns an object with by-day trip counts as a list of ints. Because it returns an entry for every date, I figure we can omit those date strings to save bytes on the wire and recover that list on the client from `start_date` and `end_date` if necessary.

It also holds objects containing the by-hour trip counts (a list of 24 ints) for weekday, saturday, and sunday service. This should be everything we need to power this widget or something like it:

image

n.b. the trip counts here include trips in both directions. In the UI, we'll definitely want to divide these values in half to get single-direction trip counts!

### `/api/ridership`

Accepts `start_date`, `end_date`, and `line_id` and returns a list of  `{ date, count }` objects.

While the `tripcounts` API accepts a `route_id`, this asks for a `line_id` because we don't have ridership data that differentiates between different _routes_ — e.g. `Green-B` vs. `Green-C`.

## Testing Instructions

I am spot-checking the API with various parameters:

http://localhost:3000/api/tripcounts/?route_id=66&start_date=2023-04-01&end_date=2023-05-02
http://localhost:3000/api/ridership/?line_id=line-1&start_date=2020-01-01&end_date=2023-05-03

### DIFF
--- a/server/chalicelib/service_levels.py
+++ b/server/chalicelib/service_levels.py
@@ -1,0 +1,93 @@
+from datetime import date, datetime
+from typing import List, Dict, TypedDict, Union, Literal
+
+from .dynamo import query_trip_counts
+from .data_funcs import index_by, date_range
+
+ByHourServiceLevels = List[int]
+DayKind = Union[Literal["weekday"], Literal["saturday"], Literal["sunday"]]
+TripCounts = List[Dict]
+
+
+class ByDayKindServiceLevels(TypedDict):
+    date: str
+    service_levels: ByHourServiceLevels
+
+
+class ByDayKindServiceLevels(TypedDict):
+    weekday: ByDayKindServiceLevels
+    saturday: ByDayKindServiceLevels
+    sunday: ByDayKindServiceLevels
+
+
+class GetTripCountsResponse(TypedDict):
+    start_date_service_levels: ByDayKindServiceLevels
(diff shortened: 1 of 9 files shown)


server/chalicelib/service_levels.py
```diff
@@ -0,0 +1,93 @@
...
+            counts.append(trips_today["count"])
+        else:
+            counts.append(counts[-1] if len(counts) and fill_zeros else 0)
+    assert len(counts) == (end_date - start_date).days + 1
+    return {
+        "counts": counts,
+        "start_date": start_date.isoformat(),
+        "end_date": end_date.isoformat(),
+        "start_date_service_levels": get_service_levels(trip_counts, False),
+        "end_date_service_levels": get_service_levels(trip_counts, True),
```
PatrickCleary, line 92: At first I was very confused, but now I am impressed by the foresight to include this.

PatrickCleary (approved):

Issue #533, opened 2023-05-02
Datepicker Selected Date Color


devinmatte:
<img width="328" alt="Screenshot 2023-05-02 at 3 53 34 PM" src="https://user-images.githubusercontent.com/9310513/235771406-0bac57d6-daa8-43e9-aeec-a34114b7c86e.png">

For the blue/green/orange line, we should update the color of the selected date on the picker

PR #541, opened 2023-05-04
Setting selected datepicker color to match line

devinmatte:
## Motivation

https://github.com/transitmatters/t-performance-dash/issues/533

Match the current line color

## Changes

Okay this is kinda hacky, but, there's no good interface to change the selected date color cleanly given the way flatpickr sets its styles. So, here's the function that pins a new classname onto the selected date to update the color

Screenshot 2023-05-03 at 8 39 54 PM
Screenshot 2023-05-03 at 8 40 01 PM
Screenshot 2023-05-03 at 8 40 06 PM
Screenshot 2023-05-03 at 8 40 12 PM


devinmatte:
Good catch, I'll fix that. I tried standard and light and felt like light looked better. Happy to change though, not strictly tied to light

skaplan-dev (approved):

PR #549, opened 2023-05-06
V4/chart annotations

PatrickCleary:
## Motivation

Feedback from labs folks

## Changes

Adds baselines (highest monthly median scheduled service)

And blocks for days with shuttling/no data

Screen Shot 2023-05-09 at 10 42 27 AM

## Testing Instructions
Try out different date ranges


modules/service/ServiceGraph.tsx
```diff
@@ -92,6 +97,11 @@ export const ServiceGraph: React.FC<ServiceGraphProps> = ({
             ),
             backgroundColor: pattern.draw('diagonal', '#FFFFFF', lineColor, 5),
           },
+          {
+            label: 'Baseline',
```
devinmatte, line 101: We will need to describe in the UI what baseline means. We probably should be adding a section to most pages that describes the dataset and what everything means more clearly

Side note: We should add some kind of baseline to our other graphs too!
- Travel times showing MBTA Benchmark and Transitmatters Benchmark/Baseline!

devinmatte:
Screenshot 2023-05-08 at 10 14 37 AM
Screenshot 2023-05-08 at 10 14 49 AM
This looks great, but also depressing

modules/service/ServiceGraph.tsx
PatrickCleary: Yea, completely agree. I think we probably want to add a blurb to every widget. Maybe 1 sentence directly on the widget and an info tooltip with more.

And yea, definitely want to add more baselines.

devinmatte: Maybe an expander like we use in https://dashboard-v4-beta.labs.transitmatters.org/opensource/ under some widgets with a description? I feel like we should make it easier to read than a tooltip.
But anyway, this looks great so far

devinmatte (approved):

skaplan-dev (approved):

PR #556, opened 2023-05-08
V4/side nav fix

PatrickCleary:
## Motivation
Add Donate and feedback buttons, as well as clean up some visual bugs with side nav.
## Changes

Add donate and feedback buttons. The nav menu feels a bit cluttered... If anyone has suggestions to improve lmk. Kinda want to add a header to put some of this stuff in. But that would be a lot of effort. Can also do that in v4.1 pretty easily.
**Desktop**
Screen Shot 2023-05-08 at 10 10 15 AM
**Mobile**
Screen Shot 2023-05-08 at 10 10 49 AM

## Testing Instructions
Click around, check out on different devices and screen sizes


devinmatte:
I think these buttons can be shrunk a bit. At least feedback I think should be skinnier. I also don't think it should be on top on mobile. I think ideally it's on the bottom of the nav in both views

PatrickCleary:
Screen Shot 2023-05-12 at 10 41 15 AM
IMG_7A369AB4AC5D-1

PatrickCleary:
Made your suggested changes @devinmatte. Good calls, looks better. I think I could probably shrink `Feedback` even more. But I think we'll also want to add a few more links in that section (like TM home page). Then we can make the `Donate` button the biggest and have everything else be small. But for the time being since there's only 2, I think the equal sizing will look the best.

devinmatte (approved):

PR #558, opened 2023-05-09
some style changes

PatrickCleary:
Made some updates to the style

- The brighter blue background matches v3
- shrunk the weekday selector
- Tried out a step function. I think it makes sense because the slope doesn't really mean much on this graph.  Also, the fact that it's divided into half-hour increments means each increment is a distinct datapoint and showing a curved line between them is somewhat misleading.

Screen Shot 2023-05-08 at 7 16 54 PM

Lmk what you think


devinmatte (approved):

PatrickCleary:
You got me addicted to stepped graphs @idreyn

Issue #532, opened 2023-05-02
Add Goatcounter to v4


devinmatte:
Add Goatcounter to V4 UI

PR #563, opened 2023-05-11
Adding basic goatcounter

devinmatte:
## Motivation

#532 

## Changes

- Adds `next-goatcounter` which is just a clean util to add goatcounter to a next app easily
- Points to our goatcounter instance


devinmatte:
Also opened https://github.com/haideralipunjabi/next-goatcounter/issues/5 to be able to remove redundant `scriptSrc` line, as it shouldn't be needed

devinmatte:
Waiting on https://github.com/haideralipunjabi/next-goatcounter/pull/6 / https://github.com/haideralipunjabi/next-goatcounter/issues/5

devinmatte:
This is ready to go now

pages/_app.tsx
```diff
@@ -23,6 +25,7 @@ export default function App({ Component, pageProps }) {
   return (
     <Layout>
       <DashboardLayout>
+        <GCScript siteUrl={'https://transitmatters-dd.goatcounter.com/count'} />
```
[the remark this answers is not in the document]
devinmatte: Good point, added a check like we do in v3

common/utils/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const PRODUCTION = 'dashboard-v4.transitmatters.org';
+export const PRODUCTION = 'dashboard.transitmatters.org';
```
PatrickCleary, line 1: Was this just an error?

PatrickCleary (approved):

common/utils/constants.ts
devinmatte: The prod address won't have a `v4` in the url so it was likely an error at first setup

PR #574, opened 2023-05-15
Updating packages and cleaning up

devinmatte:
## Motivation

Make sure we're still up to date, and clear up some warnings/log lines

## Changes

- Only runs datadog when not localhost
- Ensures we use SWC builds (faster)
- Updates all package.json packages
- Updating poetry packages
- Updating licenseinfo
- Running build CI on node 20 to prepare for node 20.x LTS


modules/service/ServiceGraph.tsx
```diff
@@ -218,7 +218,6 @@ export const ServiceGraph: React.FC<ServiceGraphProps> = ({
             if (showTitle) drawSimpleTitle(`Daily round trips`, chart);
           },
         },
-        Annotation,
```
devinmatte, line 221: Not needed in `3.0.0`

nathan-weinberg (approved):

PatrickCleary (approved):

Issue #408, opened 2023-03-27
Show real delta numbers on widgets


devinmatte:
Right now we are mostly showing a delta of `1` instead of a real comparison to week before or day before

PR #576, opened 2023-05-15
V4/titles

PatrickCleary:
## Motivation

Our graph titles were a bit scattered. This consolidates the style, adds a subtitle, and moves the location information to the widget rather than the individual graph.

The rationale to move the location information is that we want to put the widgets associated with a graph inside the same widget. So the location should look like it applies to the whole widget, rather than just the graph. It is set up to do ticket #408 

We are still using the old title style on the `Lines` page because the widget title is not always the same as the graph title.

## Changes
**Before**
Screen Shot 2023-05-15 at 11 28 18 AM

**After**
Screen Shot 2023-05-15 at 11 27 43 AM

Subtitles
Screen Shot 2023-05-15 at 11 28 44 AM

## Testing Instructions
Click around 😄

### DIFF
--- a/common/components/charts/AggregateLineChart.tsx
+++ b/common/components/charts/AggregateLineChart.tsx
@@ -19,11 +19,10 @@ import type { AggregateDataPoint, AggregateLineProps } from '../../types/charts'
 import { prettyDate } from '../../utils/date';
 import { CHART_COLORS } from '../../../common/constants/colors';
 import { DownloadButton } from '../general/DownloadButton';
-import { writeError } from '../../utils/chartError';
 import { useBreakpoint } from '../../hooks/useBreakpoint';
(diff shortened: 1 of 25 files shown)


devinmatte (approved):
Looks good, my only complaint is that there's quite a bit of spacing between the title and the graph. Ideally it would be a little bit shorter
Screenshot 2023-05-22 at 10 40 15 AM
Screenshot 2023-05-22 at 10 42 18 AM
Looking at the CSS, it looks like the extra spacing comes from chart.js itself, so may not be a simple fix

PatrickCleary:
Oh wait, nevermind. That padding was added for the old title. I thought I removed it from all the graphs but I left it on the single day and aggregate line charts. Good catch

Issue #562, opened 2023-05-11
Show running elevators/escalators


devinmatte:
It would be great if we can show how actually accessible the stations in the T are somewhere in the UI of the dashboard
```
Blue Line
Escalators 4/10 in service
Elevators 9/10 in service
```

https://www.mbta.com/alerts/access should show us elevators/escalators closures, and we can use that to create aggregate metrics of availability

kmoses228:
Additional info:

We could use GET [/facilities](https://api-v3.mbta.com/docs/swagger/index.html#/Facility/ApiWeb_FacilityController_index) from the V3 MBTA API to get a list of all elevators and escalators in the system, and periodically store the results of GET [/alerts](https://api-v3.mbta.com/docs/swagger/index.html#/Alert/ApiWeb_AlertController_index) to compute downtime in order to display history. 

I'm not familiar with how our periodic data scraping is done today, or whether it would be best to follow the same framework or do something new for this feature.

devinmatte:
@kmoses228 We do a lot of regular data processing in the `data-ingestion` repo which would probably be the place to look for keeping a historical record of the live info

nathan-weinberg:
@devinmatte @kmoses228 anything else for this or did the changes in #578 cover it and we can close this out?

devinmatte:
We never added aggregate stats, just showing the live alerts. Not sure if we want that still or not, since the live alerts tell a lot of the story

nathan-weinberg:
That would be cool to have - we could also use the alert history we collect to show percentages of how often an elevator/escalator is out of service (thinking about Porter :thinking:)

PR #578, opened 2023-05-16
V4.1/Accessibility alerts

kmoses228:
## Motivation

Related to https://github.com/transitmatters/t-performance-dash/issues/562

We are not currently showing any data related to accessibility on the dashboard, however this is essential for riders who, for example, cannot use stairs. We also want to hold the MBTA accountable for fixing these issues in a timely manner and keeping accessible features in good working order.

## Changes

This PR adds accessibility alerts to the "Today" dashboard for each subway line. Users can click on the alert to see the full description and workaround, if available.

I've also added an `/api/facilities` endpoint for fetching all elevators and escalators on a line. This is not in use yet but will be in the future.

The next PR will focus on adding a historical dashboard for elevator and escalator downtime per-line.

image

image
## Testing Instructions

Run the branch locally and interact with the alerts on each line. Also soliciting feedback on the design.


modules/dashboard/Today.tsx
```diff
@@ -16,12 +18,19 @@ export const Today: React.FC<TodayProps> = ({ lineShort }) => {
...
+  const accessibilityAlerts = useAccessibilityAlertsData(lineShort);
+
   return (
     <PageWrapper pageTitle={'Today'}>
       <div className="flex flex-col gap-y-4">
         <div className="flex flex-col gap-x-4 gap-y-4 xl:flex-row">
-          <Alerts />
+          <Alerts title="Alerts" alerts={rideAlerts} />
           {canShowSlowZonesMap && <Speed />}
+          <Alerts title="Accessibility" alerts={accessibilityAlerts} />
```
devinmatte, line 33: I think if we're going to reuse alerts here, we should add some kind of prop that get's rid of the `today` and `upcoming` sections, because we probably only care about current accessibility alerts

PatrickCleary:
Looks great. Sorry for the delay, we were in full "finish v4" mode. We ended up reworking this page quite a bit as you may have seen.

I think we should get a group together and have a discussion about how we want to portray accessibility data. We could have a design session to talk about some ideas!

devinmatte:
@kmoses228 Now that we've launched v4, it would be a good idea to readdress this and find a good spot for it

akh4rf:
https://github.com/transitmatters/t-performance-dash/assets/59891647/216a8d72-8063-4789-889f-858cf54e365e

common/api/alerts.ts
```diff
@@ -51,6 +58,31 @@ const fetchAlertsForBus = async (busRoute: string): Promise<AlertsResponse[]> =>
...
+
+  const lineStations = stations[route].stations;
+  let stationKeys = [] as string[];
+  if (isLineMap(lineStations)) {
+    stationKeys = lineStations.stations.map((station: Station) => station.station);
+  } else {
+    stationKeys = lineStations.map((station: Station) => station.station);
+  }
+
+  const options = { ...accessibilityAlertsAPIConfig, stop: stationKeys.join(',') };
```
idreyn, line 74: Also not a comment specific to this PR but I think we ought to have a utility function to handle this request setup for us, as we (myself included) have been duplicating it across these fns.

common/api/facilities.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { FacilitiesResponse } from '../types/facilities';
+import type { LineShort } from '../types/lines';
+import { APP_DATA_BASE_PATH } from '../utils/constants';
+import type { Station } from '../types/stations';
+import { isLineMap } from '../types/stations';
+
+export const fetchAllElevatorsAndEscalators = async (
+  line: LineShort
+): Promise<FacilitiesResponse> => {
+  const url = new URL(`${APP_DATA_BASE_PATH}/api/facilities`, window.location.origin);
```
idreyn, line 11: Same here (@all) we could consider an `apiFetch()` helper to handle the base path, params, etc.

common/types/facilities.ts
```diff
@@ -0,0 +1,30 @@
+export interface FacilityProperty {
+  value: string;
+  name: string;
+}
+
+export interface Facility {
+  type: string;
```
idreyn, line 7: Is there a known set of possible values here?

common/types/stations.ts
```diff
@@ -20,6 +20,10 @@ export interface Station {
   short?: string;
 }
 
+export const isLineMap = (obj: LineMap | Station[]): obj is LineMap => {
```
idreyn, line 23: Another suggestion for code structure over time (no need to address here): I think we could find a better name for the `LineMap` type, especially since we now have more map-related code elsehwere in the codebase. Maybe `UnidirectionalRouteDescription`?

modules/commute/alerts/AccessibilityAlert.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { AlertBoxInner } from './AlertBoxInner';
+
+interface DelayAlertProps {
+  alert: FormattedAlert;
+  type: UpcomingOrCurrent;
+  lineShort: LineShort;
+}
+
+export const AccessibilityAlert: React.FC<DelayAlertProps> = ({ alert, type, lineShort }) => {
+  const lineStations = stations[lineShort].stations;
```
idreyn, line 18: Here's another place where `getStationsForLine()` would be nifty.

modules/commute/alerts/AccessibilityAlert.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <AlertBoxInner
+      header={alert.header}
+      Icon={alert.type === AlertEffect.ESCALATOR_CLOSURE ? EscalatorIcon : ElevatorIcon}
+      alert={alert}
+      type={type}
+      showTimeSince={true}
+    >
+      <span>
+        {alert.type === AlertEffect.ESCALATOR_CLOSURE
+          ? 'Escalator out of service'
```
idreyn, line 37: Consider factoring out the inner string calculation for clarity.

modules/commute/alerts/AlertBox.tsx
```diff
@@ -73,9 +77,9 @@ export const AlertBox: React.FC<AlertBoxProps> = ({ alerts, lineShort, busRoute,
     } else {
       return (
         <div className="flex w-full flex-row-reverse gap-x-2 md:flex-col-reverse md:gap-x-0 md:gap-y-2">
-          {relevantAlerts.map((alert: FormattedAlert) =>
-            getAlertComponent(alert, lineShort, type, busRoute)
-          )}
+          {relevantAlerts
```
idreyn, line 80: Could we move the calculation of this array into a `useMemo()`? a `JSON.stringify()` is a somewhat expensive operation to do on every render.

modules/commute/alerts/Time.tsx
```diff
@@ -70,3 +70,13 @@ export const UpcomingTime: React.FC<TimeProps> = ({ times }) => {
     </>
   );
 };
+
+export const SinceTime: React.FC<TimeProps> = ({ times }) => {
```
idreyn, line 74: I think either name is fine but I wonder if we should use `TimeSince` to match the `showTimeSince` prop elsewhere.

idreyn (approved):
This looks great to me and I think we should land it as soon as possible! (Sorry for letting it languish a bit).

I do think we probably want to put the alerts below the graphs on the Line Overview page. Thoughts, all?

idreyn (approved):
This looks great to me and I think we should land it as soon as possible! (Sorry for letting it languish a bit).

I do think we probably want to put the alerts below the graphs on the Line Overview page. Thoughts, all?

PatrickCleary:
Looks great, I'm excited to get this in. Thanks Austin and Kevin.

I have a few minor suggestions:
1. I agree with Ian, this should be below the graphs (or just smaller).
2. I don't think this belongs in a separate page. It's all real-time data, so it doesn't adhere to the date configurations. It can just be on the overview I think.
3. I _believe_ we don't need the `upcoming` section for accessibility. I don't think they are posted in advance generally? Not certain.

modules/commute/alerts/Time.tsx
akh4rf: Changing to `EffectiveTime`

modules/commute/alerts/AlertBox.tsx
akh4rf: Not even really sure why this is here, I'll just remove it

common/types/facilities.ts
akh4rf: It seems like it can only be `facility`, I'll change this.

common/api/facilities.ts
akh4rf: Added!

common/api/alerts.ts
akh4rf: Added!

devinmatte:
I'm with @PatrickCleary in thinking we don't need a separate Alerts page right now. I think we could make something like this but for now I think these living exclusively on the overview page is fine by me

PR #579, opened 2023-05-16
Shrink Legend

PatrickCleary:
## Motivation

The legend was taking up a bit too much space on mobile. We're trying to pack a lot in, so every pixel counts. Also, I think it looked a bit scattered.

## Changes
Before / After
Screen Shot 2023-05-16 at 4 47 25 PM

## Testing Instructions

Look at various screen sizes

### DIFF
--- a/common/components/charts/Legend.tsx
+++ b/common/components/charts/Legend.tsx
@@ -9,9 +9,9 @@ export const Legend: React.FC = () => {
     >
       <div className="col-span-2 flex flex-row items-baseline gap-2 pb-1 lg:pb-0">
         <p>
-          Compare to (
-          <span className="top-[1px] inline-block h-2.5 w-2.5 items-center border-t-2 border-[#bbb] bg-[#ddd]"></span>{' '}
-          MBTA benchmark):
+          Compare to{' '}
+          <span className="top-[1px] inline-block h-2.5 w-2.5 items-center border-t-2 border-[#bbb] bg-[#ddd] shadow-sm"></span>{' '}
+          MBTA benchmark:
         </p>
       </div>
       <p>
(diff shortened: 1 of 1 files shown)


common/components/charts/Legend.tsx
```diff
@@ -4,40 +4,43 @@ export const Legend: React.FC = () => {
   return (
     <div
       className={
-        'grid w-full grid-cols-2 items-baseline gap-2 p-1 text-left text-xs sm:flex sm:flex-row sm:gap-4'
+        'grid w-full grid-cols-2 items-baseline p-1 text-left text-xs lg:flex lg:flex-row lg:gap-4'
       }
     >
+      <div className="col-span-2 flex flex-row items-baseline gap-2 pb-1 lg:pb-0">
+        <p>
+          Compare to (
```
devinmatte, line 12: I don't think we need the parentheses

common/components/charts/Legend.tsx
```diff
@@ -4,40 +4,43 @@ export const Legend: React.FC = () => {
...
+          Compare to{' '}
+          <span className="top-[1px] inline-block h-2.5 w-2.5 items-center border-t-2 border-[#bbb] bg-[#ddd] shadow-sm"></span>{' '}
+          MBTA benchmark:
+        </p>
+      </div>
       <p>
         <span className="mr-1 inline-block h-2.5 w-2.5 rounded-full border border-[#57945B] bg-[#64b96a]"></span>
-        {'On time'}
+        {'Below 125%'}
       </p>
```
skaplan-dev, line 20: typo (i think)

PatrickCleary: I'm trying to see green dots are below 125% of the baseline. I think that's accurate (?). Maybe I need to reword it

devinmatte (approved):

PR #580, opened 2023-05-17
ci: added optional check for conventional commits

nathan-weinberg:
## Motivation

This PR adds an action for conventional commits, but it will not be required as a prerequisite for merging

## Changes

Adds new action, will begin running against Pull Requests after merge

## Testing Instructions

Tested already with the Slow Zone Bot

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -1,0 +1,13 @@
+name: commit
+
+on:
+- pull_request_target
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      - name: Check if pull request commits are conventional
+        uses: webiny/action-conventional-commits@v1.1.0


devinmatte:
So this means every PR/commit that doesn't follow it will have a failed job? Since we can't have a "pass with failures" kind of job like gitlab it's going to make it harder to tell if a commit actually passes codebuilding CI at a glance

nathan-weinberg:
> So this means every PR/commit that doesn't follow it will have a failed job?
Correct
> Since we can't have a "pass with failures" kind of job like gitlab it's going to make it harder to tell if a commit actually passes codebuilding CI at a glance
If you look at the checks that are currently running, you see some are marked as `Required` and some are not - so long as the `Required` checks are passing, the Pull Request will show as mergable. The README badges and top-level commits in `main` will continue to show green as those checks are not only separate but only run against the master branch.

devinmatte:
As I stated in slack, since we always "Squash and Merge" this is wasted effort as it requires that all commits in the working branch match the conventional commits, but, then all those commits are squashes and never make it to `main`, thereby, being all for nothing.

nathan-weinberg:
Discussed on Slack - going to close this and potentially replace with just a PR title check after trying it out on the Slow Zone Bot

PR #582, opened 2023-05-17
V4/delta widget

### LINKED ISSUE #561: Double seconds label
Screenshot 2023-05-10 at 6 34 29 PM

13 sec sec

PatrickCleary:
## Motivation

1. Upcoming Widget PR needs this.
2. Fixes #561 
3. Improves UX for slow zones map

## Changes

Shrink widget text slightly

Properly use `sec`/`min` labels. Uses the same logic as elsewhere: sub 100 seconds = `sec`, sub 90 min = `min`, anything large =`hrs`

Add a `DeltaTimeWidget` to our widgets classes.

In this screenshot you can see we use seconds instead of minutes for the value below 90. Thee values are different because I am comparing local to beta URL which is out of date.

Screen Shot 2023-05-17 at 11 20 35 AM

## Testing Instructions

Look at all the slow zone tooltips

### DIFF
--- a/common/components/widgets/internal/BasicWidgetDataLayout.tsx
+++ b/common/components/widgets/internal/BasicWidgetDataLayout.tsx
@@ -13,6 +13,7 @@ export type BasicWidgetDataLayoutProps = {
   widgetValue: WidgetValueInterface;
   sentimentDirection?: SentimentDirection;
   layoutKind?: LayoutKind;
+  isLarge?: boolean;
 };
 
 export const BasicWidgetDataLayout: React.FC<BasicWidgetDataLayoutProps> = ({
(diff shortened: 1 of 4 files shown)


common/components/widgets/internal/BasicWidgetDataLayout.tsx
```diff
@@ -21,26 +22,30 @@ export const BasicWidgetDataLayout: React.FC<BasicWidgetDataLayoutProps> = ({
   widgetValue,
   layoutKind = 'total-and-delta',
   sentimentDirection = 'negativeOnIncrease',
+  isLarge = true,
```
skaplan-dev, line 25: could we make this `size` instead and pass a t-shirt size `s, m, l, xl, etc`. Then we can use `text-${prop}`

PatrickCleary: It's used in two places, so would need two props.

And I think from a design standpoint we don't want much variability in font sizes. Should only have two sizes based on screen size imo.

If it's an issue of code cleanliness, I could do this:
Screen Shot 2023-05-17 at 2 56 03 PM

devinmatte (approved):

PR #584, opened 2023-05-17
Data Widget Carousel

PatrickCleary:
🚨**This is only present on the multi-day trips view for travel times and headways**🚨
Will add to all graphs in separate PRs

## Motivation

Improve look and functionality of data widgets within graphs. Specifically:
- Widgets separated from associated graphs was confusing (see Service page)
- Widgets should be "screenshottable" with _only_ agg stats
- A screenshot of a graph should not be overwhelming. That is why we only show one at a time, along with most obvious metric (average or current)

## Changes

Adds `<WidgetCarousel/>` component
Screen Shot 2023-05-31 at 2 49 35 PM

https://github.com/transitmatters/t-performance-dash/assets/46229349/b4717b1e-e6a3-4e95-9a5d-0e4b42a76950

## Testing Instructions
View on various screen sizes/devices


devinmatte:
Not sure I'm a fan of the side by side for the trips page. I personally still think it looks better stacked on top of each other. Which it looks like still happens on a `md` pagesize. Maybe we do do the breakout at `xl` pages? So it only shows like that on big big monitors imo
Screenshot 2023-05-31 at 8 25 54 PM
Screenshot 2023-05-31 at 8 26 11 PM

Luckily, I don't feel too too strongly about this, but I suspect some members of labs may

devinmatte:
Screenshot 2023-05-31 at 8 28 45 PM
Unrelated to this PR, but, we should probably remove this option for v4

devinmatte (approved):
If the split is at `xl`, then I'm good with this

PR #589, opened 2023-05-30
Add Layouts

PatrickCleary:
## Motivation

We want to be able to customize layouts depending on page. Specifically, the landing page does not need controls and a header.

## Changes
Add secondary layout component to _app.tsx which applies layout for each page.

## Testing Instructions
Go to root URL `/` and see that title and controls are gone. Other pages should remain the same.


devinmatte:
https://github.com/transitmatters/t-performance-dash/assets/9310513/6802dbb2-8d88-4976-97c6-7c999e92e10f

So when going to `/` this happens when trying to use the sidebar

PatrickCleary:
> So when going to `/` this happens when trying to use the sidebar

@devinmatte 
This was preexisting - the navbar shouldn't really be displaying "System" it is just defaulting to that because a tab is not selected. Really we shouldn't show this selection on the landing. I added a fix to show a different nav on the landing page, which I think is good because it also simplifies our nav for first-time users.

But I could actually just make it a second PR, because now this is 28 files.

common/layouts/layoutTypes.ts
```diff
@@ -0,0 +1,6 @@
+export enum LayoutType {
```
skaplan-dev, line 1: can we change this to just be `Layout`? Its self explanatory that its a type

skaplan-dev (approved):

PR #595, opened 2023-06-02
Basic system slowzones page

devinmatte:
## Motivation

Show a basic graph so we can map v3 urls to v4 easily and show _something_

## Changes

Screenshot 2023-06-01 at 8 30 52 PM


PatrickCleary (commented):
Page looks good to me. Could you just make it so clicking the Slow Zones tab loads the page with the past year set? The preset seems to be working but the dates aren't

I forget where that logic is exactly. You may have to do something in the `getHref` function in `router.ts`

devinmatte:
Yeah I can dig deeper to find that. I was trying to set that up but struggled to find where it was

skaplan-dev (approved):

PatrickCleary:
Assigning myself to remember to figure out the automatic date selection

PatrickCleary (approved):

PR #621, opened 2023-06-05
V4/no data notice

PatrickCleary:
## Motivation

We had no checks on empty data for wrappers with widgets. Errors were thrown when data was empty, this shows a notice instead.

## Changes
Screen Shot 2023-06-05 at 11 15 00 AM

## Testing Instructions
Go to date range with no data, look at different graphs.


devinmatte:
Like with v4, I would prefer if it said `No data available. Select a different date or station pair`. Maybe the second sentence as a sub-title under no data available. I just feel like we should call out that it's not necessarily an error, there's just no data

PatrickCleary:
Screen Shot 2023-06-05 at 1 57 27 PM

devinmatte (approved):

PR #636, opened 2023-06-07
V4/landing page service graph

PatrickCleary:
## Motivation
Landing page. 

Add graph for service data.

UI cleanup coming in later PR.

## Changes
Screen Shot 2023-06-09 at 9 37 36 AM


common/constants/dates.ts
```diff
@@ -22,6 +22,8 @@ export const ONE_WEEK_AGO = TODAY.subtract(7, 'days');
 export const ONE_WEEK_AGO_STRING = ONE_WEEK_AGO.format(DATE_FORMAT);
 export const ONE_YEAR_AGO = TODAY.subtract(1, 'year');
 export const ONE_YEAR_AGO_STRING = ONE_YEAR_AGO.format(DATE_FORMAT);
+export const THREE_MONTHS_AGO = TODAY.subtract(90, 'days');
+export const THREE_MONTHS_AGO_STRING = TODAY.subtract(90, 'days').format(DATE_FORMAT);
```
skaplan-dev, line 26: can we remove `STRING` from the consts in this file?

https://hackernoon.com/why-you-should-always-avoid-encoding-type-into-names-f62y37ba

skaplan-dev: Nit: most of the names in this file can be renamed. Doesn't need to happen now though 👍

skaplan-dev (approved):

common/constants/dates.ts
PatrickCleary: That's an interesting article. What do you do in this case where the two variables represent the same thing (the date of three months ago) and the only difference is the data type?

I've found this is helpful because when I'm trying to use this value elsewhere I can quickly grab the object or the string version by inputting the appropriate name.

PR #640, opened 2023-06-07
V4/landing graph style

PatrickCleary:
Waiting on the graph PRs
## Motivation

Adding style similar to Figma mockup

## Changes
Screen Shot 2023-06-12 at 1 31 47 PM

## Testing Instructions
Screen sizes


modules/landing/charts/LandingPageChart.tsx
```diff
@@ -45,6 +45,7 @@ export const LandingPageChart: React.FC<LandingPageChartsProps> = ({ datasets, l
       <Line
         id={id}
         height={240}
+        redraw={true}
```
devinmatte, line 48: When this is set, ensure we do some memoization where possible to avoid unnecessary redraws

devinmatte (approved):

modules/landing/charts/LandingPageChart.tsx
PatrickCleary: This whole component is memoized. That's what you mean right?

devinmatte: Yep. Just making sure

PR #645, opened 2023-06-09
Rename dashboardConfig -> dateConfig

PatrickCleary:
## Motivation

With the upcoming trips section changes, it became evident that storing both the `date` and `station` configuration as one object was an issue. Mainly because we want multi day and single day trips to share `station` configuration but not `date` configuration. It also will improve code quality and remove complicated logic.

This change renames `dashboardConfig` to `dateConfig`. It also removes `to` and `from` from the `tripConfig`, which means the dashboard no longer has state to track the station. This will be fixed in the upcoming PRs. If we want to avoid that we could keep this as a separate branch and merge followup changes to this branch.

## Changes

The only change should be that station selection should not persist between stations. This is a degradation that should be fixed soon.

## Testing Instructions

Test that date persists across sections with previous behavior.

### DIFF
--- a/common/constants/pages.ts
+++ b/common/constants/pages.ts
@@ -30,7 +30,7 @@ export enum PAGES {
   tripDwells = 'tripDwells',
 }
 
-export type Section = 'landing' | 'today' | 'line' | 'overview' | 'trips' | 'system';
+export type DateConfigOptions = 'landing' | 'today' | 'line' | 'overview' | 'trips' | 'system';
 export type SectionTitle = 'Today' | 'Line' | 'Overview' | 'Trips' | 'System';
 
 export type PageMetadata = {
@@ -39,7 +39,7 @@ export type PageMetadata = {
   name: string;
   lines: Line[];
   icon: IconDefinition;
-  section: Section;
+  section: DateConfigOptions;
   sectionTitle?: SectionTitle;
   sub?: boolean;
   title?: string;
--- a/common/state/types/dateConfigTypes.ts
+++ b/common/state/types/dateConfigTypes.ts
@@ -10,8 +10,6 @@ export type TripsSectionParams = Partial<{
   startDate: string;
   endDate: string;
   queryType: QueryTypeOptions;
-  from: string;
-  to: string;
 }>;
 
 export type SystemSectionParams = Partial<{
@@ -23,7 +21,7 @@ export type OverviewPresetParams = {
   view: OverviewDatePresetKey;
 };
 
-export interface FullDashboardConfig {
+export interface FullDateConfig {
   lineConfig: LineSectionParams;
   tripConfig: TripsSectionParams;
   systemConfig: SystemSectionParams;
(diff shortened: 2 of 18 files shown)


common/state/utils/dateConfigUtils.ts
```diff
@@ -1,31 +1,31 @@
...
 import type { QueryParams } from '../../types/router';
-import { getParams } from '../../utils/router';
-import type { DashboardConfig } from '../dashboardConfig';
+import { getDateParams } from '../../utils/router';
+import type { DateConfig } from '../dateConfig';
 
-export const saveDashboardConfig = (
-  section: Section,
+export const saveDateConfig = (
+  section: DateConfigOptions,
```
skaplan-dev, line 10: `section` name doesn't really make sense here anymore. can this be changed to reflect the new type better?

common/state/types/dateConfigTypes.ts
```diff
@@ -23,7 +21,7 @@ export type OverviewPresetParams = {
   view: OverviewDatePresetKey;
 };
 
-export interface FullDashboardConfig {
+export interface FullDateConfig {
```
skaplan-dev, line 24: `FullDateConfig` implies the existence of a `PartialDateConfig`. Can we rename this to be clearer?

why not just have
`DateConfig`

and then use the TS `Partial<DateConfig>` if you wanted to

common/state/utils/dateConfigUtils.ts
```diff
@@ -1,31 +1,31 @@
...
+    dateConfig.setLineConfig(params);
   }
 };
 
-export const getDashboardConfig = (section: Section, dashboardConfig: DashboardConfig) => {
-  if (section === 'trips') return dashboardConfig.tripConfig;
-  if (section === 'system') return dashboardConfig.systemConfig;
-  if (section === 'line') return dashboardConfig.lineConfig;
-  if (section === 'overview') return dashboardConfig.overviewPreset;
+export const getDateConfig = (section: DateConfigOptions, dateConfig: DateConfig) => {
```
skaplan-dev, line 24: same comment about `section` name here.

common/constants/pages.ts
```diff
@@ -39,7 +39,7 @@ export type PageMetadata = {
   name: string;
   lines: Line[];
   icon: IconDefinition;
-  section: Section;
+  section: DateConfigOptions;
```
skaplan-dev, line 42: same comment about section naming

skaplan-dev (changes requested):

common/state/utils/dateConfigUtils.ts
PatrickCleary: I'm realizing now the state object should probably be called like `dateStore` (like Zustand docs uses) and the object it stores should be called dateConfig :/

PatrickCleary:
Good call on the renaming @skaplan-dev, I ended up renaming everything to differentiate between the object used to control the date config (DateStore) and the configuration itself. I think that makes things a lot clearer.

skaplan-dev (approved):
names are lookin much cleaner 👍 🔥

PR #650, opened 2023-06-09
Split trips into two top-level tabs.

PatrickCleary:
## Motivation
As discussed in most recent v4 working session
1. Simplifies nav-bar.
2. trips detail pages were not really being utilized.

**Currently station config will not persist between trips pages. That will be fixed in upcoming PR.**

## Changes
1. Change query param `startDate` to `date` when looking at a single date. 
2. Update v3 -> v4 logic
3. Remove trips subpages and leaves just `trips` and `multiday trips`

Open to suggestions on titles/icons. Hopping to merge as is for now and can come back to update those later.
Screen Shot 2023-06-09 at 2 51 44 PM

## Testing Instructions

Hopefully all date logic should still work. Station logic will not.

Can test to confirm that is the case.


common/components/inputs/DateSelection/DatePickers.tsx
```diff
@@ -47,36 +48,41 @@ export const DatePickers: React.FC<DatePickerProps> = ({ range, setRange, type,
...
   };
 
+  const handleDateChange = (newDate: string) => {
+    updateQueryParams({ date: newDate }, false);
+    clearPreset();
+  };
+
   React.useEffect(() => {
     updateColor(line);
   });
```
skaplan-dev, line 88: don't we need an empty dependency array here?

skaplan-dev (approved):

devinmatte (approved):

common/components/inputs/DateSelection/DatePickers.tsx
PatrickCleary: yes, I think so. I will add and check it doesn't break anything.

Issue #651, opened 2023-06-09
Correctly redirect to 404 pages on deploy


PatrickCleary:

PatrickCleary:
The other question is whether this is getting overwritten by deploys

mathcolo:
Most definitely... this is all configured in code
S3 ErrorDocument: https://github.com/transitmatters/t-performance-dash/blob/dashboard-v4/server/frontend-cfn.json#L100
CloudFront 404 handler: https://github.com/transitmatters/t-performance-dash/blob/dashboard-v4/server/frontend-cfn.json#L155

Both of those can probably be removed in v4?

PatrickCleary:
yes. Will open a PR to redirect those to our 404 page.

PatrickCleary:
Making a note to myself - these are mostly fixed but there's a couple small issues with the sidebar links on the 404 page

PR #652, opened 2023-06-09
improved 404 page

### LINKED ISSUE #648: Better 404 Page
Important because some old links won't work

PatrickCleary:
closes #648 

## Motivation

404 page was essentially empty

## Changes
Add explanation and link to home
Screen Shot 2023-06-09 at 5 20 39 PM

## Testing Instructions
🤷‍♂️

### DIFF
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -1,14 +1,20 @@
 import Head from 'next/head';
 import React from 'react';
+import Link from 'next/link';
 import { Layout } from '../common/layouts/layoutTypes';
 
 export default function Page() {
   return (
-    <div>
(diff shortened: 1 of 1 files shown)


mathcolo:
So [over here](https://github.com/transitmatters/t-performance-dash/issues/651#issuecomment-1585862819) I suggested removing those custom 404 pages entirely because I didn't know this PR existed. But now I do! So those paths can probably just be nextjs generates out of this?

devinmatte (approved):
I do think it would be nice to have something fun here, but that's hardly a 4.0 or even 4.1 necessity

PR #654, opened 2023-06-10
Fix system tab controls and title bug on mobile

### LINKED ISSUE #639: No controls on mobile on system > slow zones
Controls don't show up on this page on mobile.

https://dashboard-v4-beta.labs.transitmatters.org/system/slowzones/?startDate=2022-06-07&endDate=2023-06-07

PatrickCleary:
## Motivation
Closes #639 

also there was an issue with the page title logic on mobile. Seems when we added the slow zones tab we only updated the desktop header title logic. 

Still needed: `DatePresetConfig` logic for systems page.

Ideally there would only be one source of truth for mobile and desktop.... but not worth it at the moment imo.
## Changes
**Before** (no controls and `•` before the title)
 Screen Shot 2023-06-10 at 3 41 08 PM
 
 **After**
 Controls and correct title.
Screen Shot 2023-06-10 at 3 41 51 PM

## Testing Instructions
Go to System > Slow zones page.

### DIFF
--- a/common/components/controls/MobileControlPanel.tsx
+++ b/common/components/controls/MobileControlPanel.tsx
@@ -7,8 +7,8 @@ import { DateControl } from './DateControl';
 
 interface MobileControlPanelProps {
   dateStoreSection: DateStoreSection;
-  line: Line;
-  busRoute: BusRoute | undefined;
+  line?: Line;
+  busRoute?: BusRoute;
   queryType?: QueryTypeOptions;
 }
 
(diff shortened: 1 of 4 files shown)


common/components/controls/MobileControlPanel.tsx
```diff
@@ -7,8 +7,8 @@ import { DateControl } from './DateControl';
 
 interface MobileControlPanelProps {
   dateStoreSection: DateStoreSection;
-  line: Line;
-  busRoute: BusRoute | undefined;
+  line?: Line;
+  busRoute?: BusRoute;
```
devinmatte, line 11: Just a note: So the reason to do `| undefined` instead of `?` is that it requires a param be prop be set to a variable always, even if that var can be `undefined`. So we should try to only use `?` when we might not even have a bar to send in

devinmatte (approved):

common/components/controls/MobileControlPanel.tsx
PatrickCleary: Ah okay, will update.

Issue #398, opened 2023-03-25
Backup hosting for v3 after V4 launch.


PatrickCleary:
We should keep the old dashboards available at a different domain after launching V4.

Issue #428, opened 2023-03-31
Link conversion from v3 -> v4


PatrickCleary:
Middleware to convert V3 links to V4

devinmatte:
We still need to handle `/slowzones`

devinmatte:
Middleware doesn't seem to be working in beta right now: https://dashboard-v4-beta.labs.transitmatters.org/rapidtransit?config=Blue,70051,70039,2023-06-12,
It just seems to drop you on the 404 page

PatrickCleary:
But it works locally right? Did we ever test in beta?

devinmatte:
Yep works locally, so this may not work in static bundles

PatrickCleary:
Whoops 😔
![Screen Shot 2023-06-15 at 1 27 44 PM](https://github.com/transitmatters/t-performance-dash/assets/46229349/fd7a5af0-9d1f-4276-8125-5e8f7d14a430)

PatrickCleary:
Also doesn't include redirects and headers. Is there any way to fix this without those...?

devinmatte:
Yeah. I'll have to take a look at this again. We may have to manually handle this by making routes and using browser redirects

PR #656, opened 2023-06-11
[Release day] v3 hostname to dashboard-v3.labs.transitmatters.org

mathcolo:
Fixes https://github.com/transitmatters/t-performance-dash/issues/398

We'll still want v3 up at dashboard-v3.labs.transitmatters.org. Intent is for v3 to remain on this branch, "v3". This PR likely won't ever actually get merged

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -133,7 +133,7 @@ class App extends React.Component {
       this.state.error_message = this.checkForErrors(this.state.configuration);
     }
 
-    if (window.location.hostname !== 'dashboard.transitmatters.org') {
+    if (window.location.hostname !== 'dashboard-v3.labs.transitmatters.org') {
       showBetaTag();
     }
 
--- a/deploy.sh
+++ b/deploy.sh
@@ -28,16 +28,16 @@ done
 # Setup environment stuff
 # By default deploy to beta, otherwise deploys to production
 
-$PRODUCTION && ENV_SUFFIX=""                                    || ENV_SUFFIX="-beta"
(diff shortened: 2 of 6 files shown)


devinmatte:
People may not be aware of v3/V4 outside of transitmatters. Maybe `dashboard-old` or something of the sort? Similar to old.reddit

mathcolo:
Well we're only using this internally to check things for a little while we transition, right? I figure we're not actually supporting v3 use after v4's out.

devinmatte:
Oh I was under the impression we would keep it around publicly for a little while so that people very against change can still use the old interface until the performance API is shut down

PatrickCleary (approved):

mathcolo:
yeah I hear you @devinmatte—I think with @PatrickCleary's point in addition to https://github.com/transitmatters/t-performance-dash/issues/428 we should be in an ok spot

That said this shouldn't be merged until release day

src/App.js
```diff
@@ -133,7 +133,7 @@ class App extends React.Component {
       this.state.error_message = this.checkForErrors(this.state.configuration);
     }
 
-    if (window.location.hostname !== 'dashboard.transitmatters.org') {
+    if (window.location.hostname !== 'dashboard-v3.labs.transitmatters.org') {
```
devinmatte, line 136: May I recommend 
```suggestion
    if (window.location.hostname !== 'dashboard-old.labs.transitmatters.org') {
```
since most people might not realize it was even v3 before

mathcolo:
closing, as this is already on `v3` branch

PR #659, opened 2023-06-12
Fix 404 bugs

PatrickCleary:
## Motivation

Some pages that should be 404 were throwing errors. Mainly because the sidebar logic was trying to work with broken links. Updated sidebar logic to not throw errors when given incorrect pages.

For example, going to `/red/invalidPage` would show a client-side error message.

## Changes

for line/busRoute selection - send links to default page (overview for rail, trips for bus).

Make sure invalid pages are not trying to save state.

## Testing Instructions

Go to any invalid page. `/invalid`, `/<valid-line>/invalid`, `/system/invalid`, etc. and the sidebar _should_ show a reasonable value and all links still work.

### DIFF
--- a/common/utils/router.tsx
+++ b/common/utils/router.tsx
@@ -174,7 +174,7 @@ export const getHref = (
   linePath: LinePath
 ) => {
   const pageObject = ALL_PAGES[currentPage];
-  if (pageObject && pageObject?.dateStoreSection === newPage.dateStoreSection) {
+  if (pageObject && pageObject.dateStoreSection === newPage.dateStoreSection) {
     return navigateWithinSection(linePath, newPage, query);
   }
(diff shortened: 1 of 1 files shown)


common/utils/router.tsx
```diff
@@ -174,7 +174,7 @@ export const getHref = (
   linePath: LinePath
 ) => {
   const pageObject = ALL_PAGES[currentPage];
-  if (pageObject?.dateStoreSection === newPage.dateStoreSection) {
+  if (pageObject && pageObject?.dateStoreSection === newPage.dateStoreSection) {
```
devinmatte, line 177: By adding `pageObject` we don't need the `?.` anymore, and can just use `pageObject.dataStoreSelection`

Not necessary to change, just a note

devinmatte (approved):

PR #675, opened 2023-06-15
V4 single date default

### LINKED ISSUE #570: Default date selector to yesterday before ~6 AM
If you open the dashboard after 12 AM, the graphs all show today, but we probably want to show yesterday's values until we start receiving today's data.

JNuss71:
## Motivation

Closes #570

## Changes

This keeps the default selected date in single-day pages to be the prior day between midnight and 6AM. 

Introduces a new variable TODAY_SERVICE_STARTED for whether the service day started which is a proxy for meaningful data probably being available in the charts and indicating that the default single-day date should be today rather than yesterday.

## Testing Instructions

Either test before and after midnight and before and after 6AM ET or modify the logic for TODAY_SERVICE_STARTED to make testing not rely on it being those specific times of the day.


PatrickCleary (approved):
Thanks Jared!

PR #682, opened 2023-06-16
Adding historical alerts

devinmatte:
## Motivation

Adding historical alerts to same day view similar to v3

## Changes

Screenshot 2023-06-16 at 2 17 11 PM


skaplan-dev (approved):

PatrickCleary:
Discussed with Devin offline. I would like to rework some parts of this before adding it. Don't want to put it in as is, because then we likely won't get around to changing it. I will assign this to myself as a reminder.

PR #685, opened 2023-06-19
Using replace instead of push to fix back nav

### LINKED ISSUE #681: Back Navigation Broken
Right now the history seems to be crowded with extra instances. Making it so hitting back doesn't do much (especially on mobile where it's hard to hit back twice)

I think part of the issue is links in the page:
- load the page
- then after loading, the url is updated with date and station params, leading to two entries in the history

I think we can resolve this with `replace` instead of `push` url events

Screenshot 2023-06-16 at 10 31 58 AM

devinmatte:
## Motivation

Fixes #681 

## Changes

Changes from a `push` to a `replace` to stop polluting the history allowing logical back/forward nav

## Testing Instructions

Click around links, and then use browser back nav one page at a time and make sure there isn't 2 entries per click, but at least 1

### DIFF
--- a/common/utils/router.tsx
+++ b/common/utils/router.tsx
@@ -120,7 +120,7 @@ export const useUpdateQuery = () => {
 
       if (!isEqual(router.query, newQuery)) {
         const query = pickBy(newQuery, (attr) => attr !== undefined);
-        router.push({ pathname: router.pathname, query }, undefined, { shallow: true });
+        router.replace({ pathname: router.pathname, query }, undefined, { shallow: true });
       }
     },
     [router]


PatrickCleary (approved):

PatrickCleary:
This makes it so changing the dates then clicking back doesn't go back to previous date? Is that the only thing this changes or was it fixing something else?

Because in my experience, with most apps with date pickers, when you change the date then go back it goes back to the previous selected date range.
For example: google flights/datadog/aws

@devinmatte

devinmatte:
This was fixing the fact that before it wouldn't let you go back at all. Because each page click would have 2 entries in the history. So clicking back would go to the same page and then repopulate the default params. Basically making the back button never work. We can likely fix this by making the default population of params be a replace and the subsequent changes a push

PR #688, opened 2023-06-19
Reroute v3 urls in static build

### LINKED ISSUE #428: Link conversion from v3 -> v4

devinmatte:
## Motivation

Fixing v3 routing in static builds

Fixes #428 finally

## Changes

- Adds basic routing handling for v3 urls


PatrickCleary:
This won't work for buses, right?

Is there anything we could do for that? Like check for `config` in query params, and then rewrite to our URLS?

devinmatte:
It does work for buses. I added the redirect to the bus index

PatrickCleary:
Looks great! Thanks for finishing this up!

Only thing I noticed was the date presets are set to defaults rather than what they would be if you were to navigate there directly. I confirmed it was working on `dashboard-v4`. 

My suspicion would be `usePresetsOnFirstLoad` in `common/utils/firstLoad.ts` needs to run but is not? Lmk if I can help debug

devinmatte:
Fixed @PatrickCleary

PatrickCleary (approved):

PR #697, opened 2023-06-23
[v3] hide secrets in healthcheck output

mathcolo:
## Motivation

Exceptions can contain outbound http request URLs, which can expose API keys.

## Changes

- Hide keys from public healthcheck output

## Testing Instructions

1. Change the API key environment variable to an invalid value, causing the healthcheck to fail
2. The healthcheck output should contain the string `HIDDEN`.


nathan-weinberg (approved):
Good call @mathcolo!

PR #698, opened 2023-06-23
[v4] hide secrets in healthcheck output

mathcolo:
See [v3 equivalent](https://github.com/transitmatters/t-performance-dash/pull/697) for info


nathan-weinberg (approved):
Good call @mathcolo!

PR #708, opened 2023-06-27
Change time widget unit format

PatrickCleary:
## Motivation

discussion on widget units. Still need to update tooltips to match.
https://transitmatters.slack.com/archives/GSJ6F35DW/p1687459523389939?thread_ts=1687459081.589399&cid=GSJ6F35DW

## Changes
Screenshot 2023-06-27 at 5 21 22 PM

## Testing Instructions

Look at all widget values please! Not sure I checked them all.


common/types/basicWidgets.tsx
```diff
@@ -57,7 +59,12 @@ export class DeltaZonesWidgetValue extends BaseWidgetValue implements WidgetValu
   }
   getFormattedValue() {
     if (this.delta === undefined) return '...';
-    return `${this.delta > 0 ? '+' : '-'}${Math.abs(this.delta)}`;
+    return (
+      <p>
+        <WidgetText text={`${this.delta > 0 ? '+' : '-'}${Math.abs(this.delta)}`} />{' '}
```
mathcolo, line 64: could you screenshot me what it looks like if it's positive?

PatrickCleary: Screenshot 2023-06-27 at 5 28 27 PM

PatrickCleary: Just realized it was showing a `-` for 0 change.

Changed it to 0 because I think `+0 zones` makes more sense

mathcolo: you're killing me with that uppercase Z

but OK looks good! also unless you think -0 is a risk, you don't need the abs()

mathcolo: oh yes I agree, +0 makes more sense indeed

mathcolo (approved):

common/types/basicWidgets.tsx
PatrickCleary: Screenshot 2023-06-27 at 5 40 20 PM

PR #710, opened 2023-07-01
shrink station selector items

PatrickCleary:
## Motivation
In v3 you can see all the stations immediately on dropdown. New station selection items were too big and the container was too small.

I also think we should remove handicap and bike icons, as that is not really the purpose of our app. I think it would be more useful to use that space for branches (A/B)

## Changes
**Before**
Screenshot 2023-07-01 at 10 45 49 AM

**after**
Screenshot 2023-07-01 at 10 45 33 AM


devinmatte:
I feel strongly about keeping the bike and accessibility icons

devinmatte:
My main concern with shrinking them is a touch radius on mobile. Want to make sure it's not too easy to misclick

PatrickCleary:
**Update**

Reverted size of mobile items.

Added branch info

Fixed bug where if `to` station is branch `D` and a `from` station is selected on another branch, the app breaks. Now the `to` station will be auto-updated to a core station (gov center for green, park for red).,

devinmatte (approved):

PR #714, opened 2023-07-05
Pull speed restrictions data from DynamoDB

idreyn:
In https://github.com/transitmatters/data-ingestion/pull/26 I added the MBTA's data on track speed restrictions to our DynamoDB tables. This PR updates the DD's speed restriction info to query from that table. It ultimately goes here, in the SZ map tooltip:

image


common/api/slowzones.ts
```diff
@@ -14,10 +16,20 @@ export const fetchAllSlow = (): Promise<SlowZoneResponse[]> => {
...
+  options: FetchSpeedRestrictionsOptions
+): Promise<SpeedRestriction[]> => {
+  const { lineId, date: requestedDate } = options;
+  const params = new URLSearchParams({ line_id: getGtfsRailLineId(lineId), date: requestedDate });
+  const speedRestrictionsUrl = new URL(
+    '/api/speed_restrictions?' + params.toString(),
     window.location.origin
   );
-  return fetch(speed_restrictions_url.toString()).then((resp) => resp.json());
+  const today = new Date();
```
devinmatte, line 28: So this may cause issues after 8pm EST as it'll be tomorrow in GMT. I think we try to use dayJs dates throughout the code to avoid this

devinmatte (approved):
Looks good and worked for me in testing. Thanks!

Only noticed one thing in the code that I'm concerned about, but, can't test until it gets later today (or if it would even break anything)

common/api/slowzones.ts
idreyn: That's a good callout but in this case we're never going to display the date, we just treat it as a timestamp.

PatrickCleary (approved):

Issue #717, opened 2023-07-06
Bus date selector presets


devinmatte:
<img width="245" alt="Screenshot 2023-07-06 at 7 30 38 PM" src="https://github.com/transitmatters/t-performance-dash/assets/9310513/10d4e50e-b0bc-4188-afee-f5ef92427ae0">

These presets don't make any sense for bus data. We should change the presets 

- Most recent date
- January 2023
- December 2022
- One year ago

PR #720, opened 2023-07-07
Add bus date presets

devinmatte:
## Motivation

The rapid presets don't make sense for bus data #717 

I'm not 100% happy with these, but they're better than what we had before

## Changes

Screenshot 2023-07-06 at 9 06 16 PM
Screenshot 2023-07-06 at 9 06 25 PM


PatrickCleary (approved):
🙏

skaplan-dev (approved):

Issue #712, opened 2023-07-04
Back button fixes


PatrickCleary:
Discussed here:

https://github.com/transitmatters/t-performance-dash/pull/685#issuecomment-1620696209

We want back button to work with param changes.

PR #724, opened 2023-07-07
Fixes back navigation to account for date changes

devinmatte:
## Motivation

#712

## Changes

- Handles default loading values to not pollute history, but preserves date changes done by the user in the history

## Testing Instructions

Visit pages, hit back, make sure things feel natural and you can actually go back


PatrickCleary:
Looks good, I think we should just wait to merge until we have presets and stations working with the back button as well.

devinmatte:
That's what this fixes

PatrickCleary:
(and the presets stay on the most recent selection if you are using those)

devinmatte:
Okay we should all be good now

PatrickCleary (approved):
LGTM

Issue #702, opened 2023-06-25
"0 sec" travel time average for a Green Line trip


mathcolo:
https://dashboard-v4-beta.labs.transitmatters.org/green/trips/single/?from=70510&to=70508&date=2023-05-25

travel times are getting floored maybe? 0 sec average is weird

<img width="617" alt="image" src="https://github.com/transitmatters/t-performance-dash/assets/193453/1c794ba9-daec-46c1-acb7-abb28cf14eed">

PR #725, opened 2023-07-07
Fix averages with 0s

devinmatte:
## Motivation

#702 

## Changes

If the value is 0, it reads as `false` in the if statement.
Changing to undefined check to provide real averages

Screenshot 2023-07-07 at 3 27 25 PM


PatrickCleary (approved):
🐐

PR #731, opened 2023-07-09
feat: add travel time copy

skaplan-dev:
## Motivation
- Noticed the travel time widget didn't have a subtitle and it looked a little off next to the other widgets that do.

## Changes
- Adds travel time subtitle

| Before | After|
| ------------- | ------------- |
| Screen Shot 2023-07-09 at 3 29 25 PM | Screen Shot 2023-07-09 at 3 28 08 PM |


devinmatte:
I'm not sure about removal on mobile. I know when I first moved to Boston I needed to learn what headways meant, so this text would have been helpful to me 2 years ago.

PatrickCleary:
+1

Looks better, but we can’t remove it unless we explain it elsewhere (which
isn’t really a priority).

Issue #530, opened 2023-05-02
Slow Zones map should reflect currently selected date


idreyn:
Right now there's still a shim `new Date()` here:

https://github.com/transitmatters/t-performance-dash/blob/29dcb8965f1da6b844ba87bada58b5c96234f47b/modules/slowzones/map/SlowZonesMap.tsx#LL68C1-L68C1

We'll want to pass in the upper bound of the currently selected date range. We'll also want to make it very clear in the UI that we're showing slow zones for _that_ date (which could be far in the past).

PatrickCleary:
Something like this:
(and if the date is today, show `Today`)

There's a function somewhere to check if date === today and convert to "Today" somewhere already.

![Image](https://github.com/transitmatters/t-performance-dash/assets/46229349/25e3429f-f690-4d5c-91bf-4df98f7512f8)

PR #737, opened 2023-07-11
Slow Zones Update

akh4rf:
## Motivation

#530 

### Before

image

### After

image


PatrickCleary:
Awesome! Thanks for taking care of this!

I pulled it down locally and realized the red may look too much like a button. Can we do grey instead? I was messing around with it locally, and pushed that here https://github.com/transitmatters/t-performance-dash/tree/tailwind-update if you want to use that. (I also converted the styles to tailwind).

Last thing which I should've mentioned before: I think we only want the `today` for the maps. The other widget dates can keep their current format.

Thank you!

modules/slowzones/SlowZonesWidgetTitle.tsx
```diff
@@ -0,0 +1,54 @@
...
+        color: 'white',
+        background: LINE_OBJECTS[line].color,
+        padding: '5px 15px',
+        borderRadius: '5px',
+      }}
+    >
+      {date}
+    </p>
+  ) : null;
+};
```
skaplan-dev, line 54: super nit but I prefer this instead of the ternary.

```javascript
if (!line) {
 return null 
}

return <p></p>
```

skaplan-dev (approved):
looks good to me once @PatrickCleary 's comment is addressed 👍

modules/slowzones/SlowZonesWidgetTitle.tsx
devinmatte: Agreed here. Normally I prefer a ternary but when it's an undefined check that returns a null, I prefer @skaplan-dev's suggestion

PR #748, opened 2023-07-12
V4/remove grid

PatrickCleary:
## Motivation
I don't like the grids. 

Also, the graphs were getting stretched too far on wide screens. Set limit to width (max-w-7xl)Screenshot 2023-07-12 at 12 48 07 PM
8)


skaplan-dev:
I kinda like the grids tbh

PatrickCleary:
The main issue with the grids is that every entry needs to be the same height, or else you end up with weird gaps and whitespace. Not having a grid will give us much more flexibility in the future. Like for example, if we want to add back some of those little widgets, they won't fit in the grid structure too well. It takes a lot more work to align everything.

The other issue is it makes the widths of the charts too variable. The slope of the charts varies pretty dramatically when they get squished/stretched. This reduces that effect.

PatrickCleary:
Doing something like this in the future would be cool, but I don't think it's realistic for us at this point. Needs to be very specifically arranged. The grid is an attempt at that, but it's unfinished. 
Screenshot 2023-07-12 at 2 23 13 PM

devinmatte:
I was against the grids at first, but now they've grown on me 😞 
I'm mixed on this change but I understand the reasoning

devinmatte (approved):

PR #749, opened 2023-07-12
add histogram back

PatrickCleary:
## Motivation

Forgot to add the headways histogram back.

## Changes

Adds headways histogram back.
Removing the grid in #748 which makes the whitespace at the botom go away.
Screenshot 2023-07-12 at 12 52 53 PM

### DIFF
--- a/modules/tripexplorer/SubwayTripGraphs.tsx
+++ b/modules/tripexplorer/SubwayTripGraphs.tsx
@@ -154,7 +154,7 @@ export const SubwayTripGraphs: React.FC<SubwayTripGraphsProps> = ({
           <WidgetDiv>
             <WidgetTitle
               title="Headway histogram"
-              subtitle="Time spent at station"
+              subtitle="Time between trains"
               location={location}
               line={line}
             />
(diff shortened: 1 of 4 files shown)


skaplan-dev (changes requested):
We should have this work irrespective of #748 (the whitespace)

PatrickCleary:
I don't think that is realistic. To do that we need all the charts be the exact same height, which is the issue with the grid.

modules/tripexplorer/SubwayTripGraphs.tsx
```diff
@@ -150,6 +151,19 @@ export const SubwayTripGraphs: React.FC<SubwayTripGraphsProps> = ({
             />
             <DwellsSingleWrapper query={dwells} toStation={toStation} fromStation={fromStation} />
           </WidgetDiv>
+          <WidgetDiv>
+            <WidgetTitle
+              title="Headway histogram"
+              subtitle="Time spent at station"
```
devinmatte, line 157: This isn't correct, unless you mean "time rider spends at station waiting for train"

PatrickCleary: whoops. Good catch.

devinmatte:
I don't love using "histogram" in the visible display text. Maybe "distribution"?

devinmatte (approved):

PR #757, opened 2023-07-14
Adding trailing slashes for redirects

devinmatte:
## Motivation

We need trailing slash to avoid a 302


mathcolo (approved):
Let's see what happens!

PR #763, opened 2023-07-14
Widget fix

PatrickCleary:
## Motivation

There were a few people mentioning the widget slider button looked like a dropdown.

## Changes
Make the widgets move right to left rather than down. Still looks kinda wack, and I have a longer term plan to improve, but I think this is better for short term.

https://github.com/transitmatters/t-performance-dash/assets/46229349/51489508-8a51-4301-9c9c-c3635c5162aa

Make the widgets vertical on mobile and desktop. The horizontal was just way too cramped.
Screenshot 2023-07-14 at 12 12 54 PM

## Testing Instructions

Scroll through widgets

### DIFF
--- a/common/components/general/WidgetCarousel.tsx
+++ b/common/components/general/WidgetCarousel.tsx
@@ -2,7 +2,7 @@ import React, { Children, cloneElement, useCallback, useMemo, useState } from 'r
 import classNames from 'classnames';
 import type { ComponentProps, FC, ReactElement } from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faChevronCircleLeft, faChevronCircleRight } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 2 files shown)


devinmatte:
If we have a arrow pointing left, could we not have one on the complete other side now pointing right? I feel like that might look nice and allow people to go back and forth on the stats and complete the look

PatrickCleary:
yea... That would be better. I am being kind of lazy, cuz if you do that you have to deal with where to place them. On the full screen graph it looks weird to have the arrow on the complete other side of the chart. If you have it hug the right side of the widget, then it moves around because they are variable widths.

I want to do something completely different for these tbh, but not enough time.

If this is too weird, I'm fine to keep the down arrow, but I don't think I have time to clean it up further. If you want to look into it though, feel free.

devinmatte:
I think it's fine to keep the down arrow. If people are confused at first. The second they click it, they're no longer confused, so I don't think we need to make this change now

devinmatte (approved):

PR #766, opened 2023-07-14
Prepare deploy script for beta/prod deploy

devinmatte:
## Motivation

Deploy to normal beta, not just v4 beta

## Changes

- Copied over what felt missing from v3 deploy script
- Deploy to normal beta instead of v4 specific beta


PatrickCleary (approved):

mathcolo (approved):
yolo

PR #767, opened 2023-07-14
Upgrade data dashboard to python 3.10

devinmatte:
## Motivation

Get us on the latest runtime

## Changes

- Updates from allowing both Python 3.9 and 3.10 to only 3.10
- Changes deploy to use Python 3.10

## Testing Instructions

We'll need to deploy this, and then, switch to python 3.10 runtime in AWS

**There could be up to a minute of downtime**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         node-version: ['16']
-        python-version: ['3.9', '3.10']
+        python-version: ['3.10']
     steps:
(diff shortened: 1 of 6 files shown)


skaplan-dev (approved):

mathcolo:
V4 deployment on a new hostname will involve a fully fresh CloudFormation “stack”- meaning it’ll get new lambdas and everything for itself. For that reason you don’t really need to do this- it’s nearly all risk for not much benefit! ❤️

PR #768, opened 2023-07-14
Hit static landing page data

PatrickCleary:
## Motivation

The landing page data was slow to load on the first query.

## Changes

Fetch landing page data from `/static` folder.  Uploaded by a lambda in this pr: https://github.com/transitmatters/data-ingestion/pull/33

Add a "primer query" to landing page to get the lambdas warm for when the user goes to next page.

## Testing Instructions
Check landing page -> make sure data is all there.


modules/landing/LandingCharts.tsx
```diff
@@ -1,9 +1,28 @@
...
 import { OverallServiceChartWrapper } from './charts/OverallServiceChartWrapper';
 import { OverallRidershipChartWrapper } from './charts/OverallRidershipChartWrapper';
 import { LandingChartWidget } from './LandingChartWidget';
+
 export const LandingCharts: React.FC = () => {
+  // This query is just to get the lambdas "warm"
+  useQuery(
+    ['primer'],
+    () =>
+      fetchActualTripsByLine({
```
devinmatte, line 16: Thoughts about making this a query that might actually get called later so we can also benefit from the local cache? Right now this is under the key `primer` and therefore this cached call will never be used. Either remove the ONE_HOUR local cache or hit an endpoint with params/keys that might be called on a subsequent page

PatrickCleary: I think that's an interesting point, and I was thinking about it for a bit. I think if we try to use a useful value, we're more likely to create more wasteful operations. We have no idea what the user is gonna do next, so the query is likely to end up not being used anyway.


I also think this solution is a bit hacky and I hope to have a better fix in the future, so I think keeping it clearly a "useless" function with the name `primer` reduces the odds that it causes confusion down the line. Whereas the performance benefits seem very minor at best.

devinmatte: At least use `useDeliveredTripMetrics` with the set of params you used here, so the cache _could_ be used

devinmatte (approved):
Looks good, just one thought about trying to be practical with cache

PR #769, opened 2023-07-15
V4/widget improvements

PatrickCleary:
## Motivation

The old widgets were not cutting it.

This is better for a few reasons:
1. Standardization - all the widgets in the `trips` are more or less the same now. For more intricate data widgets where more text/explanation is needed, I would propose we add them outside of the graph widgets. These are just showing values on the chart.
2. Smaller/less obtrusive. The old widgets were kinda in the way in every screenshot, and I think the graphs are probably the highlight, not the data widgets.

## Changes

This PR Is just changing the `trip` widgets. Next I will do the `line` widgets.
Screenshot 2023-07-14 at 10 39 17 PM
Screenshot 2023-07-14 at 10 39 30 PM

## Testing Instructions
I'm not 100% certain there won't be sizing issues. So testing out various dates and screen sizes would be useful. ALthough I'm certain that will become apparent as people use it either way.


devinmatte (commented):
Despite still being a draft, I really like this and would be good to approve now. Since we're launching soon, it may logically be a 4.1 or 4.01 release

PatrickCleary:
Okay cool, I might try to polish these up and get them out today.

PatrickCleary:
Couple small changes: Not using the delta widgets for now. Preston brought up the point that making comparison to the end date on the graph is applying to much weight to one singular data point. Can think about how we want to use those in the future.

Also, making them a little less prominent.
Screenshot 2023-07-16 at 12 54 57 PM

Screenshot 2023-07-16 at 12 54 23 PM

For now we can just stick with avg/median/min/max


Last thing: Leaving the `line` widgets as is for now. Will try to apply similar changes before tomorrow, but if I don't get to it I think that's okay

common/components/widgets/internal/SmallDelta.tsx
PatrickCleary, line 1: Not used for now, but keeping it in case we want to use it soon.

common/components/charts/Legend.tsx
PatrickCleary, line 1: Hide the legend behind dropdown always

common/components/widgets/internal/WidgetForCarousel.tsx
PatrickCleary, line 1: Removing the option to make widgets horizontal which we were using on mobile

skaplan-dev:
What is the "legend" drop-down? Sorry am on mobile and can't pull it down yet.

mathcolo (approved):
+1 to @devinmatte, this is enough of an improvement to just do today!

PatrickCleary:
@skaplan-dev on mobile, the single trips legend used to always show, now it's under a dropdown. This puts it under the dropdown on mobile and desktop

PR #774, opened 2023-07-16
fix: Render SZ segment labels properly in Safari

idreyn:
## Motivation

The time labels for Slow Zones map segments were previously rendered with the `<foreignobject>` SVG element, which doesn't really work in Safari. This change renders them in pure SVG.

## Changes

SVG layout is pretty manual work and it was easiest to resort to Unicode characters for the directional arrows. I found that the ones that work best (i.e. have no weird subpixel glitches) are these chevron ones:

On mobile, though, we'll get a slightly different arrow look:

## Testing Instructions

Visit `/red/slowzones` in desktop/mobile Safari and check that these elements render properly.

### DIFF
--- a/common/components/maps/LineMap.stories.tsx
+++ b/common/components/maps/LineMap.stories.tsx
@@ -26,7 +26,7 @@ const redLineSegments: SegmentRenderOptions[] = [
       {
         mapSide: '0',
         boundingSize: 40,
-        content: (
+        content: () => (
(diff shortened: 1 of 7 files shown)


devinmatte (approved):
Poked around and tested in both Safari (13.4.1 (22F82)) and Chrome (114.0.5735.198) and everything looks good to me

PatrickCleary (approved):
👑

Issue #605, opened 2023-06-03
Create a Metrics overview page


PatrickCleary:
1. Agree on copy
2. Create page

mcplanner:
Step one is facilitating agreement on the language for this page.
Step two is implementing it.

PR #777, opened 2023-07-16
add metrics page

PatrickCleary:
## Motivation

#605 

We need some explanation of the metrics. Could definitely use a lot more, but can start with this.

## Changes

Metrics button and metrics page
Screenshot 2023-07-16 at 1 51 36 PM


mathcolo (commented):
I like the intent, but I think it would probably be better to link to parts of this information from individual charts (an info ℹ️ icon or something?).

skaplan-dev:
> I like the intent, but I think it would probably be better to link to parts of this information from individual charts (an info ℹ️ icon or something?).

+1

Issue #687, opened 2023-06-19
Slow zone map popover behind navbar


devinmatte:
<img width="605" alt="Screenshot 2023-06-18 at 8 37 40 PM" src="https://github.com/transitmatters/t-performance-dash/assets/9310513/c9f8d087-5242-47ae-8aac-9a66ae24aeb7">

We need to fix the z-indexing of the slow zone map/navbar

devinmatte:
Maybe something for @idreyn to look at with the other slow zone map tickets?

PR #782, opened 2023-07-17
fix: Better SZ tooltip positioning

### LINKED ISSUE #573: Slow zone map shifting on mobile

idreyn:
## Motivation

Resolves #573 and #687

## Changes

This changes moves the SZ tooltip to the left side of the map when there's more real estate for it there (e.g. on the Braintree branch). It also makes sure the tooltip is never hidden by the left-hand sidebar. I kept the tooltip beneath the mobile bottom controls, though — seems better; just scroll.

## Testing Instructions

Check the tooltip rendering for the Red Line map on both desktop and mobile.

### DIFF
--- a/common/components/maps/LineMap.module.css
+++ b/common/components/maps/LineMap.module.css
@@ -9,6 +9,12 @@
   position: absolute;
 }
 
+@media screen and (min-width: 770px) {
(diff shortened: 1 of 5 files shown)


devinmatte (approved):
LGTM

PR #792, opened 2023-07-17
Set API Gateway name to the backend hostname, for debugging ease

mathcolo:
## Motivation
This morning we couldn't figure out which API Gateway entity belonged to which hostname when we were trying to figure out what was wrong with the stacks.

## Changes

This sets the API Gateway title to the backend hostname, which makes this suck less.

## Testing Instructions

These should be the hostname instead of `data-dashboard`: https://us-east-1.console.aws.amazon.com/apigateway/main/apis?region=us-east-1

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -25,6 +25,12 @@ while getopts "pc" opt; do
   esac
 done
 
+# Ensure required secrets are set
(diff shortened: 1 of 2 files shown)


devinmatte:
Does this require a delete-redeploy?

mathcolo:
shouldn't, the name is customizable in the gui

I can check on beta before merging tho

devinmatte (approved):

PR #804, opened 2023-07-23
Pin to newer version of poetry, and cleanup package pinning

devinmatte:
## Motivation

People have been using various versions of poetry, each which generates a different lockfile. This should ensure everyone is using a newer version, leading to less lockfile issues

## Changes

- Pins poetry to use something greater than poetry core `1.5.0`
- Updates lockfile with latest poetry version
- Pins `ddtrace` to `1.15.0` until datadog fixes its size (https://github.com/DataDog/datadog-lambda-python/issues/349)
- Updates `black` and runs `black` formatting
- Pins `botocore` so it doesn't get out of step with `boto3`
 
## Testing Instructions

Run `poetry self update`

If the above fails, then uninstall old versions of poetry, then install from scratch using docs https://python-poetry.org/docs/


devinmatte:
Running in Beta now

mathcolo (approved):
Awesome, thanks!!

PR #830, opened 2023-08-11
Fix data labels

PatrickCleary:
## Motivation
I made an oopsie. ChartJS data labels plugin for some reason applies to all charts... Add code to remove that default.

## Changes
Fixes bug with chartJS data labels.
Live on https://dashboard-beta.labs.transitmatters.org

## Testing Instructions
Look at charts.


devinmatte (approved):
Bad plugin 😞

PR #836, opened 2023-08-16
Made v4 the default build version

nathan-weinberg:
## Motivation

Now that v4 has launched, it probably makes sense to make that the default `build` command and regulate v3 to a custom command `build-v3` 

## Changes

- All former `build` references changed to `build-v3`
- All former `build-v4` references changed to `build`

## Testing Instructions

Run new commands and compare


devinmatte:
We probably don't need the v3 method anymore at all

devinmatte (approved):

Issue #869, opened 2023-09-29
Unwanted overlap in Slow Zone chart


nathan-weinberg:
**Describe the bug**
Some of the slow zone times are overlapping with station names - we should bump out the distance between the branches to compensate 

**To Reproduce**
1. Go to the Red Line Slow Zone page on the Dashboard

**Expected behavior**
No overlap

**Screenshots**
![image-1](https://github.com/transitmatters/t-performance-dash/assets/31703736/742773b8-5fc7-4515-b2cb-842b785e79ea)

**Device Info (please complete the following information):**
 - Device: Laptop
 - OS: Fedora 36
 - Browser: Firefox 
 - Browser Version: 112.0.1

**Additional context**
N/A

PR #873, opened 2023-10-04
[869] Fix overlap in Red Line slow zone chart

akh4rf:
## Motivation

Addresses #869 

## Changes

image

### DIFF
--- a/common/components/maps/diagrams/lines.ts
+++ b/common/components/maps/diagrams/lines.ts
@@ -8,6 +8,7 @@ import type { Turtle } from './types';
 type DiagrammableLineName = 'Red' | 'Orange' | 'Blue';
 
 type CreateDiagramOptions = {
+  /** Number of pixels between each station */
   pxPerStation?: number;
 };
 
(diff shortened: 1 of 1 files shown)


common/components/maps/diagrams/lines.ts
```diff
@@ -76,8 +77,10 @@ export const createDefaultDiagramForLine = (
   lineName: DiagrammableLineName,
   options: CreateDiagramOptions = {}
 ) => {
-  if (lineName === 'Red') {
-    return createRedLineDiagram(options);
+  switch (lineName) {
+    case 'Red':
+      return createRedLineDiagram(options);
+    default:
+      return createStraightLineDiagram(lineName, options);
```
akh4rf, line 84: I mainly just changed this in anticipation of the green line map, which I hope to work on once we extract the map stuff into a library. It's not necessary but idk to me this is easier to read.

common/components/maps/diagrams/lines.ts
```diff
@@ -8,6 +8,7 @@ import type { Turtle } from './types';
 type DiagrammableLineName = 'Red' | 'Orange' | 'Blue';
 
 type CreateDiagramOptions = {
+  /** Number of pixels between each station */
```
devinmatte, line 11: Thank you for this! This code is sometime hard to follow given it's complexity so docs like this are very helpful

devinmatte (approved):
LGTM but maybe @idreyn should take a quick pass to confirm?

common/components/maps/diagrams/lines.ts
akh4rf: No problem! I'm definitely going to add more of this but I figured I didn't want to bloat this PR with too many comments.

Issue #863, opened 2023-09-11
SZ chart minimum should be 0


PatrickCleary:
Slow zone chart y axis goes negative sometimes. Change minimum to be 0.
![image](https://github.com/transitmatters/t-performance-dash/assets/46229349/d904eb42-73e5-4644-a712-a792dccba54e)

PR #874, opened 2023-10-04
[863] Set Total Slow Time minimum to zero

akh4rf:
## Motivation

Addresses #863

## Changes

image

### DIFF
--- a/modules/slowzones/charts/TotalSlowTime.tsx
+++ b/modules/slowzones/charts/TotalSlowTime.tsx
@@ -111,6 +111,7 @@ export const TotalSlowTime: React.FC<TotalSlowTimeProps> = ({
             scales: {
               y: {
                 display: true,
+                min: 0,
                 ticks: {
                   color: COLORS.design.subtitleGrey,
                 },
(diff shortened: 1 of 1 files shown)


modules/slowzones/charts/TotalSlowTime.tsx
```diff
@@ -111,6 +120,7 @@ export const TotalSlowTime: React.FC<TotalSlowTimeProps> = ({
             scales: {
               y: {
                 display: true,
+                min: dataMin < 0 ? 0 : undefined,
```
devinmatte, line 123: Can we not just set this to 0 always?

devinmatte (approved):
Looks good just wondering why we need to find the min at all?

modules/slowzones/charts/TotalSlowTime.tsx
akh4rf: No, because for periods where the min is 55 and the max is 70 (for example), all of the data would be clumped at the top if we had the min at zero.

PR #879, opened 2023-10-19
feature: Unified `TimeSeriesChart` component

idreyn:
## Motivation

We have an awful lot of duplicated code in our charts. Usually when I need to make one I copy some code from a similar chart and make some very minor adjustments. So I've decided to take a crack at a unified `TimeSeriesChart` component that we should be able to use as the basis for basically all the charts in the DD where the x-axis is time.

## Changes

No changes to existing functionality, only a new component that I'll definitely use for new work, and may start retrofitting into existing dashboard tabs.

## Testing Instructions

`npm run storybook` and visit the `TimeSeriesChart` component page. Here are some examples of what this can do with a fairly minimal configuration:

image

image

image


PatrickCleary (approved):
This will be great. Thanks!

PR #881, opened 2023-10-21
Pc/feat/nav bar update

PatrickCleary:
## Motivation

**WIP**

Still neeed to:
- Add bus route selection

### DIFF
--- a/modules/navigation/SidebarTabs.tsx
+++ b/modules/navigation/SidebarTabs.tsx
@@ -26,8 +26,8 @@ export const SidebarTabs: React.FC<SidebarTabs> = ({ tabs, setSidebarOpen }) => 
   };
 
   return (
-    <div className="rounded-md bg-white bg-opacity-5 p-1" role={'navigation'}>
-      <ul className="space-y-1">
+    <div className="px-1" role={'navigation'}>
+      <ul>
         {tabs.map((tab: PageMetadata) => {
           const enabled = line ? tab.lines.includes(line) : true;
           const selected = page === tab.key;
@@ -45,21 +45,11 @@ export const SidebarTabs: React.FC<SidebarTabs> = ({ tabs, setSidebarOpen }) => 
                   selected
                     ? 'bg-stone-900 text-white'
                     : enabled && 'text-stone-300 hover:bg-stone-800 hover:text-white',
-                  'group flex select-none items-center gap-x-3 rounded-md px-2 text-sm font-semibold leading-6',
+                  'group flex select-none items-center gap-x-3 rounded-sm text-xs font-semibold leading-6 pl-1',
                   enabled ? 'cursor-pointer' : 'cursor-default  text-stone-600',
-                  tab.sub ? 'ml-2 py-1 text-xs' : 'py-2'
                 )}
               >
-                <FontAwesomeIcon
-                  icon={tab.icon}
-                  aria-hidden={true}
-                  className={classNames(
-                    selected ? 'text-white' : 'text-stone-200',
-                    enabled ? 'group-hover:text-white' : 'text-stone-600',
-                    'shrink-0',
-                    tab.sub ? 'h-4 w-4' : 'h-6 w-6'
-                  )}
-                />
+                <span className="text-lg">•</span>
                 <span className="truncate">{tab.name}</span>
               </Link>
             </li>
--- a/modules/navigation/SidebarTabs.tsx
+++ b/modules/navigation/SidebarTabs.tsx
@@ -1,6 +1,5 @@
 import classNames from 'classnames';
 import React from 'react';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import Link from 'next/link';
 import {
   useDelimitatedRoute,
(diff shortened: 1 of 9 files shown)


.prettierrc
PatrickCleary, line 1: I couldn't get prettier to work until I changed this. Is this a bad idea? @devinmatte

devinmatte: Hmm did a new update break `.js` prettierrcs?

idreyn (commented):
This is a really huge improvement, I think. I have a few stray design thoughts for now, I haven't looked at the code yet.

image

1. With reference to the above image, I wonder whether we can drop the _Bus_ and _Subway_ headings. Red/Orange/Blue/Green/Bus as top-level items seem fine. Having a _Bus_ heading with one subentry, Buses, is throwing me off a bit.

If we keep them, I think we should consider padding _below_ the headings instead of to the _left_ of the sub-items. For example:

image

2. Some menu items like global Slow Zones don't have hover/active states yet. I also think those should retain some icons.

3. Are icons for the per-line subpages (service etc) a no-go at this resolution? I kinda liked them but I can see how they'd be a bit busy here.

4. The bus line dropdown (white text on yellow) is really hard for me to read. Could we use a more traditional black text on white here? It might also look better to expand the dropdown so that it entirely fills the yellow rectangle that contains it, with no padding or rounded corners. 

image

PatrickCleary:
Thanks for checking this out. All good points.

1. Works for me
2. I thought I had added those. Might’ve forgot to push the last commit?
either way, definitely agree those are needed.
3. Yea, I had the icons at first and it was a bit much. I can send a pic of
what it would look like. Maybe we can incorporate them elsewhere?
4. will do

PatrickCleary:
@idreyn – Just updated with commit I had forgot to merge. Should have hover/active states for top two options (which should also still have icons).

Here is a pic of icons in the sidebar:
Screenshot 2023-11-27 at 9 59 50 PM

and without:
Screenshot 2023-11-27 at 9 51 27 PM

**Bus color/style adjustment**

Idk if this is what you were thinking. Let me know. Also, if you want to go in and try anything, please do.
Screenshot 2023-11-27 at 9 57 54 PM

PatrickCleary:
(most recent commit has no sidebar icons and updated bus style)

idreyn:
The bus dropdown is 👨‍🍳👌 and I still kinda like the icons in the sidebar tbh but that's just one opinion!

My vote is to land this basically as-is and then iterate if we want to change it. I'm setting a reminder to review the code tomorrow when I am feeling less zonked.

devinmatte:
So this looks great. Pushed it to beta to make it easier to test, and see it with real data

So I'll say that I'm also partial to the icons on the sections returning, but otherwise this is really clean 👌 

I found a bug where I can still click the disabled green line slowzone section and open the broken page
Screenshot 2023-11-30 at 5 54 11 PM

modules/navigation/SidebarTabs.tsx
```diff
@@ -45,21 +50,11 @@ export const SidebarTabs: React.FC<SidebarTabs> = ({ tabs, setSidebarOpen }) =>
...
-                  icon={tab.icon}
-                  aria-hidden={true}
-                  className={classNames(
-                    selected ? 'text-white' : 'text-stone-200',
-                    enabled ? 'group-hover:text-white' : 'text-stone-600',
-                    'shrink-0',
-                    tab.sub ? 'h-4 w-4' : 'h-6 w-6'
-                  )}
-                />
+                <span className="text-lg">•</span>
```
mathcolo, line 57: Out of curiosity, what if the • wasn't there?

PatrickCleary:
Okay I think everyone's comments have been addressed. Added back icons based on slack poll results.

@devinmatte @mathcolo @idreyn

mathcolo:
Could we get the most recent changes on beta just to check it out before merge?

modules/navigation/SidebarTabs.tsx
```diff
@@ -11,23 +11,29 @@ import type { PageMetadata } from '../../common/constants/pages';
 
 interface SidebarTabs {
   tabs: PageMetadata[];
-  setSidebarOpen?: React.Dispatch<React.SetStateAction<boolean>>;
+  close?: (
```
idreyn, line 14: This never gets called with an argument so I think it could be just `close: () => void`. There are a few other components wrapping this one that have the same prop, and afaict that change could be made there too.

idreyn (approved):
Latest looks very classy.

I have one final nitpick (and really just one): in certain states, it looks like the elements could use some padding — between the selected Orange Line and the not-selected Red Line, and between the Red circle-icon and the left-edge of its rounded container.

image

But of course good to land as-is.

devinmatte (approved):
Looks great. I'll echo that perhaps we need a little additional margins between the lines, but, nothing major. Looks solid!

PatrickCleary:
I won't have time again today to look at it, but the feedback is all
reasonable to me. If we're tryna get this in soon and someone wants to push
those changes, please do.

PR #885, opened 2023-10-30
feature: Service hours chart

idreyn:
## Motivation

We want to graph the total number of (scheduled and delivered) service hours for Red, Orange, and Blue lines. This is a complementary metric to trip counts, which also happens to be more robust to shuttling.

## Changes

Adds a new graph to `/<line>/service`:

image

For now weekly/monthly aggregation is done on-the-fly instead of using agg tables, so large queries (2+ years) sometimes fail. There's also no reliable data before about 2018.

Also: I'm using `<TimeSeriesChart>` for the first time here! It powers this chart and now the scheduled/delivered trips chart, too.

## Testing Instructions

Visit the Service pages for Red, Orange, Blue and confirm that this graph loads and renders correctly.


tsconfig.json
```diff
@@ -24,6 +24,6 @@
       }
     ]
   },
-  "include": ["next-env.d.ts", "globals.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "include": ["next-env.d.ts", "globals.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts", "common/components/charts/TimeSeriesChart"],
```
devinmatte, line 27: Can I ask why? I would expect this would already be covered by `"**/*.ts", "**/*.tsx"`

idreyn: Very strange. I did not make this change manually, some tool might have done it

PatrickCleary (approved):

PR #886, opened 2023-10-30
Show latest slowzone values

devinmatte:
## Motivation

We have access to both the average and the latest slowzone values, we should show both, and the map should prioritize latest when showing `today`

## Changes

Screenshot 2023-10-30 at 4 14 01 PM

Screenshot 2023-10-30 at 3 51 28 PM


nathan-weinberg:
nit: should `latest` be `current`?

devinmatte:
@nathan-weinberg The field exists in slowzones that have since ended. It shows the last slow speed before the slowzone was lifted/resolved, so it's latest for all except the ongoing slowzones

idreyn (approved):
Great fix for this issue, lgtm!

Issue #861, opened 2023-09-06
dropdown menus moving with page scroll


PatrickCleary:
**Describe the bug**

The dropdown menus seem to scroll with the page. Seems to only be Firefox.

Video + Details:
https://transitmatters.slack.com/archives/GSJ6F35DW/p1693958876191559

ascampos:
confirmed that this issue is no longer happening on any browser, including Firefox

PR #890, opened 2023-11-19
fix scrolling issue in dropdown menu

ascampos:
## Motivation
https://github.com/transitmatters/t-performance-dash/issues/861

The previous `position: fixed` caused unexpected behavior during scrolling due to a `transform` applied to its containing element as part of the animation that happens when you open the dropdown menu. This is due to an interesting quirk of CSS that you can read about [here](https://meyerweb.com/eric/thoughts/2011/09/12/un-fixing-fixed-elements-with-css-transforms/)

## Changes

Changed `position: fixed` to `position: absolute` for the dropdown menu container to address scrolling issues. By switching to `position: absolute`, the element is now fixed with respect to its nearest positioned ancestor's context without being affected by `transform`s.

## Testing Instructions

Tested the change across different browsers and mobile views to validate the fix's effectiveness in resolving the scrolling issue.

### DIFF
--- a/common/components/inputs/StationSelector.tsx
+++ b/common/components/inputs/StationSelector.tsx
@@ -69,7 +69,7 @@ export const StationSelector: React.FC<StationSelector> = ({
             leaveFrom="transform opacity-100 scale-100"
             leaveTo="transform opacity-0 scale-95"
           >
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):
Perhaps one of our frontend gurus wants to do the final merge but from my end LGTM!

devinmatte (approved):

PR #896, opened 2023-12-10
Add commuter rail colors and cleanup

devinmatte:
## Motivation

Prepare to add commuter rail to the dashboard in the future, by adding the colors and the sidebar selector ready to go

## Changes

- Adding commuter rail keys and colors
- Adding a commuter rail tab to codebase, but not adding it to the UI yet
- Fixes Bus sidebar icon

## Testing Instructions

There's nothing currently to test. Nothing is set to display

### DIFF
--- a/common/components/nav/CommuterRailDropdown.tsx
+++ b/common/components/nav/CommuterRailDropdown.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import { SidebarTabs } from '../../../modules/navigation/SidebarTabs';
+import { TRIP_PAGES, COMMUTER_RAIL_OVERVIEW } from '../../constants/pages';
+import { CommuterRailRouteSelection } from './CommuterRailRouteSelection';
+
+interface CommuterRailDropdownProps {
+  close?: () => void;
+}
+
+export const CommuterRailDropdown: React.FC<CommuterRailDropdownProps> = ({ close }) => {
+  return (
+    <div className="rounded-b-md">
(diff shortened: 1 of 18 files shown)


common/components/nav/CommuterRailDropdown.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { SidebarTabs } from '../../../modules/navigation/SidebarTabs';
+import { TRIP_PAGES, COMMUTER_RAIL_OVERVIEW } from '../../constants/pages';
+import { CommuterRailRouteSelection } from './CommuterRailRouteSelection';
+
+interface CommuterRailDropdownProps {
+  close?: () => void;
+}
+
+export const CommuterRailDropdown: React.FC<CommuterRailDropdownProps> = ({ close }) => {
```
skaplan-dev, line 10: my take is we shouldn't use React.FC unless we're using typed children / the other props we get from it. It muddles the type for the component.

skaplan-dev (approved):
one nit

common/components/nav/CommuterRailDropdown.tsx
devinmatte: I tend to disagree. I think it helps well label which functions are components vs just functions personally feel it reads cleaner

skaplan-dev: Everything is just functions 😛  its react baby

PR #898, opened 2023-12-10
Commuter rail in Data Dashboard

devinmatte:
> [!IMPORTANT]
> ### Merging Soon
>  - We should notify the T ahead of releasing so they are aware of the data issues we've found
> - We also should make sure that language is sufficient to note that this data is far from perfect, and should not be relied on as well as our other data sources

## Motivation

Add the commuter rail to the data dashboard

## Changes

Screenshot 2023-12-26 at 4 12 24 PM

- Adds all commuter rail lines to the dashboard using Gobble data


devinmatte:
Okay so for commuter rail we will need to handle stations with more than 1 stop id per direction. For south station there are cases of 8+ ids. Right now the dashboard only handles one of them at a time. We're probably going to want to change the logic handling for at least the commuter rail.
But I suspect this means we're not handling park street properly

devinmatte:
Relies on https://github.com/transitmatters/data-ingestion/pull/63 being merged first

modules/tripexplorer/CommuterRailTripGraphs.tsx
rudiejd, line 1: could we format the times on the y axes here to be `{h}hr {m} m`?

austinjpaul:
just saw this, check green line for examples of many stop ids for the same
stop simultaneously

devinmatte:
Yep @austinjpaul i discovered that and deleted the comment. We're all good. Just took a moment to find it in the code

common/components/notices/BetaDataNotice.tsx
```diff
@@ -1,28 +1,14 @@
...
-  const isStartDateAfterBusMaxDay =
-    (startDate !== undefined && dayjs(startDate).isAfter(BUS_MAX_DAY)) ||
-    (date !== undefined && dayjs(date).isAfter(BUS_MAX_DAY));
-  const isEndDateAfterBusMaxDay = endDate !== undefined && dayjs(endDate).isAfter(BUS_MAX_DAY);
-
-  if (
-    (line === 'line-commuter-rail' || linePath === 'commuter-rail') &&
-    (isStartDateAfterBusMaxDay || isEndDateAfterBusMaxDay)
-  ) {
+  if (line === 'line-commuter-rail' || linePath === 'commuter-rail') {
```
skaplan-dev, line 11: whats the difference between these?

modules/tripexplorer/CommuterRailTripGraphs.tsx
```diff
@@ -0,0 +1,134 @@
...
+          </WidgetDiv>
+          <WidgetDiv className="flex flex-col justify-center">
+            <WidgetTitle title="Travel times by hour" location={location} line={line} both />
+            <AggregateChartWrapper
+              query={traveltimes}
+              toStation={toStation}
+              fromStation={fromStation}
+              type={'traveltimes'}
+              timeUnit={'by_time'}
+              peakTime={peakTime === 'weekday' ? true : false}
```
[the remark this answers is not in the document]
skaplan-dev: good suggestion datadog

server/chalicelib/s3_historical.py
```diff
@@ -78,8 +78,7 @@ def headways(stop_ids: list, start_date: date, end_date: date):
         headway_time_sec = delta.total_seconds()
 
         # Throw out any headways > 120 min
-        # TODO: We can't do this anymore for CR data
-        if headway_time_sec > 120 * 60:
+        if headway_time_sec > 120 * 60 and not s3.is_cr(this["route_id"]):
```
skaplan-dev, line 81: 120 minute headways lets goooo

server/scripts/generate_line_files.py
```diff
@@ -92,15 +92,17 @@ def parse_stop_name(stop_name: str):
...
-            )
+            station_json = {
+                "stop_name": stop["attributes"]["name"],
+                "station": stop["id"],
+                "branches": None,
+                "order": index + 1,
+                "stops": stop_layout[stop["id"]],
+            }
+            if index == 0 or index == len(stops["data"]) - 1:
+                station_json["terminus"] = True
```
skaplan-dev, line 103: is this so we can show a notice that terminal data is somewhat unreliable?

common/components/notices/BetaDataNotice.tsx
devinmatte: I believe the issue is that sometimes one of the values may be `undefined` so we need to check both

server/scripts/generate_line_files.py
devinmatte: Yep!

common/components/notices/BetaDataNotice.tsx
rudiejd: could we consistently use commuter rail? I feel like `line-commuter-rail` makes it sound like a GTFS route, which could be confusing since commuter rail has many GTFS routes

common/constants/cr_constants/cr-fitchburg.json
```diff
@@ -0,0 +1,204 @@
...
+        "order": 17,
+        "stops": {
+          "0": ["CR-Fitchburg_0_FR-0034-02", "CR-Fitchburg_0_FR-0034-01"],
+          "1": ["CR-Fitchburg_1_FR-0034-02"]
+        }
+      },
+      {
+        "stop_name": "North Station",
+        "station": "place-north",
+        "branches": null,
```
rudiejd, line 183: what does branches represent here? Can it be removed?

common/components/notices/BetaDataNotice.tsx
rudiejd: although I guess it's fine since we currently have bus as `line-bus` lol

modules/tripexplorer/CommuterRailTripGraphs.tsx
rudiejd: I still think this is relevant - the data shows up as (hour, minute) on the data point tooltip but just (minute) on the y axis.

common/constants/cr_constants/cr-fitchburg.json
devinmatte: It is part of the dataset we use for Red and Green line. It just keeps it consistent with the other lines. We can probably remove it but then it'll be undefined instead of null so we'd need to make sure the code can handle that

common/components/notices/BetaDataNotice.tsx
devinmatte: Yeah this is probably something we should clean up someday

modules/tripexplorer/CommuterRailTripGraphs.tsx
devinmatte: Screenshot 2025-03-01 at 5 44 45 PM
Done

rudiejd (approved):

skaplan-dev (approved):

Issue #803, opened 2023-07-21
Speed for station pairs


PatrickCleary:
**Is your feature request related to a problem? Please describe.**
Show MPH speed for a station pair

**Describe the solution you'd like**
The first thing that would need to be done is to create a mapping of station pairs to distances. Currently we only have full-line distances.

**Additional context**
I (@PatrickCleary) have ideas on how this would look. If anyone wants to work on this, let me know and we can talk design.

rudiejd:
I can take this one

rudiejd:
https://github.com/transitmatters/data-ingestion/blob/bd84c4ccd6594ee83de86de6525334481cf370a8/ingestor/chalicelib/daily_speeds.py#L20 it looks like this data is already there, we just aren't storing it for each station pair

rudiejd:
it seems like the [speed comes from the miles traveled divided by seconds of time aggregated in](https://github.com/rudiejd/t-performance-dash/blob/5c4f84477c8c543241343b4faa0425f080cb2184/server/chalicelib/speed.py#L73) `DeliveredTripMetrics`.  [we fetch all the data for each station pair,](https://github.com/transitmatters/data-ingestion/blob/bd84c4ccd6594ee83de86de6525334481cf370a8/ingestor/chalicelib/daily_speeds.py#L20) we just don't write it to any table. the easiest way to solve this seems to be a trip metrics table that includes data for each station pair to drive this new graph. I'll do the data ingestion work first and then work on the frontend stuff

PatrickCleary:
Thanks for looking at this!

Is it possible we could calculate it when a stop-pair page is opened, rather than storing it all? I'm concerned it will be a lot of data to store.

rudiejd:
no problem! My concern there would be getting rate limited, but I think we're fine as long as we only do this speed calculation on the trip metrics page for each line. I'll try coding up something to do that

rudiejd:
There could also be a more sophisticated structure for storing this than just paid to pair trip seconds. For example, you could store the accumulation of average trip seconds and distance for each station in a direction, then use that data structure to calculate trip times / distances. For example inbound from alewife on the red line

{
Alewife: {
 miles_traveled: 0,
 trip_seconds_inbound: 0
}
...
{
Kendall: {
miles_traveled_inbound: dist(alewife, Kendall)
trip_seconds_inbound: trip_seconds(alewife, Kendall)
}, 
Charles: {
[...]

PR #909, opened 2023-12-17
feat(trip-explorer): speed for station pairs

rudiejd:
## Motivation

Closes https://github.com/transitmatters/t-performance-dash/issues/803

## Changes
- Add speed graph on trip explorer

image

## Testing Instructions
- Navigate to trip explorer page for all lines
- Verify that the new speed graph is populated correctly


rudiejd:
this is really speed for stop pairs if we're being technical 🤓 

I will try to fix the nomenclature to consistently call them stops

common/constants/stations.json
```diff
@@ -641,8 +899,12 @@
         "station": "place-mvbcl",
         "order": 8,
         "stops": {
```
rudiejd, line 901: I think this auto-formatted because I opened it 😮‍💨 

I wanted to add distances here, but I figured it would be either to make another file. Let me know your thoughts

server/chalicelib/speed.py
```diff
@@ -80,3 +82,6 @@ def is_invalid_range(start_date, end_date, max_delta):
     start_datetime = datetime.strptime(start_date, DATE_FORMAT_BACKEND)
     end_datetime = datetime.strptime(end_date, DATE_FORMAT_BACKEND)
     return start_datetime + timedelta(days=max_delta) < end_datetime
+
+if __name__ == 'main':
```
rudiejd, line 86: [ ] remove

server/scripts/update_stations.py
```diff
@@ -74,5 +74,112 @@
...
+    # there is definitely a more efficient way to do this.
+    # maybe hardcode the termini and start traversal from there?
+    # this also generates extraneous data for trips that are impossible
+    # for example: 70093 => 70096 by this is 20 miles.  that is JFK U/Mass (braintree) to Ashmont..
+    def populate_distances(target_stop_id):
+        stop_stack = []
+        seen_stops = set()
+        
+        # assume each stop starts with one destination 
+        first_dest, first_dist = next(iter(stop_distances[target_stop_id].items()))
```
rudiejd, line 141: you could probably make this better by doing a seperate DFS for each route 🤷

modules/landing/utils.ts
```diff
@@ -45,6 +47,24 @@ export const convertToSpeedDataset = (data: { [key in Line]?: DeliveredTripMetri
   });
 };
 
+export const convertToStationSpeedDataset = (
+  fromStationId: string,
+  toStationId: string,
+  data: SingleDayDataPoint[]
+) => {
+  const intervalDistance = getDistancesForStopId(fromStationId, toStationId);
+  console.log(`${fromStationId}, ${toStationId}, ${intervalDistance}`);
```
rudiejd, line 56: - [ ] remove

common/constants/stations.json
devinmatte: The other file seems ideal given how large it is 👍

devinmatte (commented):
This is awesome! A few initial thoughts in no particular order:

We should display both of the two station pairs since this is between two
Screenshot 2023-12-17 at 7 38 49 PM
Screenshot 2023-12-17 at 7 38 52 PM

We should add the same aggregate numbers below the chart like we do in travel times
Screenshot 2023-12-17 at 7 38 40 PM

Personally I think speed should go between headways and dwells in the chart order

I was able to get the page to bug out with `http://localhost:3000/red/trips/single/?from=70073&to=70093&date=2023-12-17` so there's probably some missing stops in the dataset?

Overall, this is very cool, just needs some polish. We can try and get this into beta before the next meeting to get people in labs to poke around with it

rudiejd:
> This is awesome! A few initial thoughts in no particular order:
> 
> We should display both of the two station pairs since this is between two Screenshot 2023-12-17 at 7 38 49 PM Screenshot 2023-12-17 at 7 38 52 PM
> 
> We should add the same aggregate numbers below the chart like we do in travel times Screenshot 2023-12-17 at 7 38 40 PM
> 
> Personally I think speed should go between headways and dwells in the chart order
> 
> I was able to get the page to bug out with `http://localhost:3000/red/trips/single/?from=70073&to=70093&date=2023-12-17` so there's probably some missing stops in the dataset?
> 
> Overall, this is very cool, just needs some polish. We can try and get this into beta before the next meeting to get people in labs to poke around with it

thanks for taking a look!! yep, this is definitely still a WIP - just threw it together yesterday/today. Those are all good. There were a couple other things I wanted to take a look at too, let me know your thoughts: 
- some of the speeds seem a bit suspect, for example average of 20mph some hours for davis to quincy center. I wonder if I should be including the dwells in addition to `travel_time_sec`
- there is no monthly/weekly aggregated version of this chart
- there is no benchmark. I could calculate a benchmark using the benchmark travel time. 
- the distance calculations from traversing that list are right, but unfortunately the dataset includes some bizarre distances - for example, JFK / UMass (Braintree) => Ashmont (Ashmont) was listed as 20 miles. I haven't debugged this super deeply yet

devinmatte:
> there is no monthly/weekly aggregated version of this chart

This certainly isn't necessary for a first pass. We can start with a single day chart only

> there is no benchmark. I could calculate a benchmark using the benchmark travel time.

there's no benchmark for dwells, so it's not critical

> some of the speeds seem a bit suspect, for example average of 20mph some hours for davis to quincy center. I wonder if I should be including the dwells in addition to travel_time_sec

Yeah maybe @PatrickCleary or @idreyn could help you track down the accuracy of these since they worked on the line-wide speed numbers. I can also poke around after the holiday

rudiejd:
> I was able to get the page to bug out with http://localhost:3000/red/trips/single/?from=70073&to=70093&date=2023-12-17 so there's probably some missing stops in the dataset?

I poked around a bit on this. It seems that `70073` is Charles MGH on the Alewife - Braintree route and `70093` is Ashmont on the Alewife - Ashmont route. By this dataset, that is unreachable since you can't access `70093` on the same route. I originally used `stop_id` for the distance calculation since it was readily available on the FE, but it seems like it might be better to use `station_id` for the distances based on this quirk 😅

rudiejd:
I changed this to use stations instead of stops, fixing the Charles => Ashmont issue, fixed the distance calculation by only traversing in one direction at a time, added a legend, and updated the screenshots. 

I also added a benchmark speed calculation. I can show it on the graph if that's something we want, but I think it might be a little confusing considering it'll be the inverse of the travel time benchmark

devinmatte:
Damn seeing full speeds for tracks that are repaired is wild
Screenshot 2023-12-20 at 8 25 18 PM

I do think this will be more interesting in aggregate (multi-day) datasets to see improvement/decline than it will be for most single day views (since it really is just another way to describe travel times) but this is quite interesting to see when a track is fixed.

rudiejd:
image
Added aggregate data.

One small issue is that some of the green line station distances e.g. Ball Square => Back of the Hill are still undefined, and the aggregate charts don't accept undefined data, so you end up with these weird graphs of 0 speed. I'll

PatrickCleary:
The green line station distances we use are just measured on Google Maps. So if you're interested in replacing some of those null values, you could add them that way!

rudiejd:
@PatrickCleary  I ended up figuring out the problem with the data - I was relying on the  `route-id` field being the line name, but it was not `Green` for all of the green, but rather `Green-A`, `Green-B`, etc. but thank you! Green line looks to be fixed now
image

I refactored the script for calculating station distances and moved it out to its own file, because the logic was getting pretty complicated.

modules/tripexplorer/SubwayTripGraphs.tsx
```diff
@@ -62,6 +64,20 @@ export const SubwayTripGraphs: React.FC<SubwayTripGraphsProps> = ({
               toStation={toStation}
             />
           </WidgetDiv>
+          <WidgetDiv>
+            <WidgetTitle
+              title="Speeds"
+              subtitle="Speed between stops"
```
devinmatte, line 70: We should say 
```suggestion
              subtitle="Average speed between stops"
```
since there are several speed limits between a pair

modules/tripexplorer/SubwayTripGraphs.tsx
```diff
@@ -130,6 +146,21 @@ export const SubwayTripGraphs: React.FC<SubwayTripGraphsProps> = ({
             />
           </WidgetDiv>
 
+          <WidgetDiv>
+            <WidgetTitle
+              title="Speeds"
+              subtitle="Speed between stops"
```
devinmatte, line 152: Same here 
```suggestion
              subtitle="Average speed between stops"
```

server/pyproject.toml
```diff
@@ -32,3 +32,10 @@ target-version = ['py311']
 [build-system]
 requires = ["poetry-core>=1.6.0"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.pyright]
+reportMissingImports = true
+reportMissingTypeStubs = false
+useLibraryCodeForTypes = false
```
devinmatte, line 39: I'm guessing this is from local environment, we probably don't want this committed

modules/tripexplorer/SubwayTripGraphs.tsx
```diff
@@ -62,6 +64,20 @@ export const SubwayTripGraphs: React.FC<SubwayTripGraphsProps> = ({
               toStation={toStation}
             />
           </WidgetDiv>
+          <WidgetDiv>
```
devinmatte, line 67: Since this is basically an inverse of travel times, we shouldn't show it back to back with travel times. We should move it below Headways at a minimum. Probably after dwells imo

modules/tripexplorer/SubwayTripGraphs.tsx
```diff
@@ -130,6 +146,21 @@ export const SubwayTripGraphs: React.FC<SubwayTripGraphsProps> = ({
             />
           </WidgetDiv>
 
+          <WidgetDiv>
```
devinmatte, line 149: Same here, below dwells ideally, definitely below headways

devinmatte (commented):
Some initial comments

server/pyproject.toml
rudiejd: this is actually  setting up the language server I use for neovim, [pyright](https://github.com/microsoft/pyright) to not scream at me because we don't use type stubs - it is in strict type-checking mode by default. it shoudn't affect anything other than pyright, and the settings should be universal (you can think of it like `.editorconfig`) but I can remove if it's confusing

devinmatte: Okay maybe @nathan-weinberg has an opinion about us keeping this in or removing then

nathan-weinberg: If the additional config indeed only affects `pyright` and won't have any other effects (i.e. no change to the project) I don't see the harm in leaving this in - don't think we have any other devs using this so we can always adjust if that ever changes

austinjpaul:
Mentioned in the meeting, just writing here for not forgetting - I think my preference would be to add speed as a toggle for the travel-times chart, since it is the same data displayed two different ways, just scaled by a factor. That way we don't end up with too many charts on page.

rudiejd:
@austinjpaul totally agree! I'll rework it as a toggle on the travel times graph

rudiejd:
I reworked this to make speed a toggle on the travel times graph. I currently have it set as the first option for travel time, because I think it's the easier to read given our users might not know what the travel times for a given station pair should be

rudiejd:
image

rudiejd:
I switched the graph around to make travel times the default since that's what users are more used to following comments from @austinjpaul in slack. This should be ready to go now

server/poetry.lock
devinmatte, line 1: I'm not sure why the lock file is getting updated here, especially with py12 binaries

server/scripts/update_stations.py
```diff
@@ -74,5 +73,5 @@
             stations_by_id[station_id][indx]["pedal_park"] = True
 
 
-with open("common/constants/stations.json", "w") as f:
+with open("common/constants/stations_distances.json", "w") as f:
```
devinmatte, line 76: This doesn't seem necessary

devinmatte (commented):
Looking good, two small things but overall, basically ready to approve

server/scripts/update_stations.py
rudiejd: good catch, that actually would have broken this script 💀

server/poetry.lock
rudiejd: no clue either. It's adding a PyYaml binary for musl, but I use glibc on my machine in WSL. I can remove this

skaplan-dev (approved):

PR #910, opened 2023-12-19
Handle new updated_on date in json

devinmatte:
## Motivation

Merge before https://github.com/transitmatters/slow-zones/pull/41

## Changes

Handles both current and new json formats for slowzones


idreyn (approved):
LGTM — could probably DRY the code out a little but it doesn't matter since this is temporary.

PR #911, opened 2023-12-19
Green line slow zones

devinmatte:
## Motivation

All green line to our slow zones views

## Changes

Screenshot 2023-12-18 at 9 35 38 PM


austinjpaul:
hell yeah! were going to have to validate the data before we publish these -- i remember seeing some things like that huge spike in march and it wasn't real.

devinmatte:
Yeah there's probably some adjustments we can make. Or we at the least should put a disclaimer for green line data in general since it's just a much more variable line even the portions with a dedicated ROW

austinjpaul:
I think a disclaimer isn't quite enough here - if there's garbage data we should probably go remove it manually from the backing files so we don't display it.

devinmatte:
Relies on https://github.com/transitmatters/stripmap/issues/2 to move forward

Issue #921, opened 2023-12-28
Green Line slow zone page should not be selectable


nathan-weinberg:
**Describe the bug**
Despite being grayed-out, users can select the Green Line `Slow zones` page, which leads to the standard widget page with infinite loading widgets. This should either not be clickable or lead to a page explaining why we don't do this.

**To Reproduce**
1. Navigate to https://dashboard.transitmatters.org/green/
2. Select `Slow zones` in left-side nav bar

**Expected behavior**
See suggestions above

**Screenshots**
![image](https://github.com/transitmatters/t-performance-dash/assets/31703736/ac000920-1027-461c-87a7-5d73b6cf96ec)

**Device Info (please complete the following information):**
 - Device: Laptop
 - OS: Fedora 36
 - Browser: Firefox
 - Browser Version: 112.0.1

PR #922, opened 2023-12-29
fix(slow_zones): make green line slow zones page unselectable

rudiejd:
## Motivation

Closes https://github.com/transitmatters/t-performance-dash/issues/921

## Changes

I went with just removing the slow zones tab for the green line. This seems like a better temporary fix until we have stuff ready for that page 
image
image

## Testing Instructions

Navigate to http://localhost:3000/green/slowzones/?startDate=2022-12-28&endDate=2023-12-28 and verify a 404. Verify that the slow zones tab for the green line is gone


devinmatte (approved):
Thanks, I tried to fix this bug a while ago but for some reason it remained so this seems like the best option

skaplan-dev (approved):

PR #931, opened 2024-01-09
Add by-hour trips histogram to Service details page

### LINKED ISSUE #904: Add error boundaries to widget wrappers
We sometimes get unexpected errors that occur within a single widget on the Overview page — in these cases, we'd like to keep the error localized to the widget so it doesn't crash the whole page. 

We can use [Error Boundaries](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) from React, and [this helper library](https://github.com/bvaughn/react-error-boundary), to accomplish this. Basically, we'll wrap certain components with an `<ErrorBoundary>`, using as a fallback the `<ErrorNotice>` component already in our codebase.

### LINKED ISSUE #914: Coloring on Overview date selector component
**Describe the bug**
The overview date selector colors are off. It looks like it's improperly reading the line color and applying DEFAULT colors.

Screenshot 2023-12-23 at 2 31 02 PM

**To Reproduce**
go to:
https://dashboard.transitmatters.org/red/
hover/click on a date range

**Expected behavior**
Should be similar coloring to other buttons

idreyn:
## Motivation

In the Covid Recovery Dash we have a little graph breaking down total scheduled trips by hour of the day.

## Changes

I'm adding something like that here:

image

I've decided to use more of a histogram look — a time series would work here, but the t-axis would be different than the rest of the page, which is confusing at a glance.

A few other things:

- I added a new `<Widget>` wrapper that handles waiting for dependencies, error states (w/ error boundaries) etc. This resolves #904 
- I also resolved #914 while I was in there.

## Testing Instructions

Visit the Service page for red/orange/blue/green and verify that you can see the histogram for a number of date ranges.


devinmatte:
@idreyn Is 1am showing up on weekends expected?
Screenshot 2024-01-09 at 8 59 22 PM

idreyn:
@devinmatte I think so but I'll double check

idreyn:
@devinmatte that's what's in the database, and the calculation logic looks right. possible the trip is a deadhead in practice but I'm not sure off the top of my head if we can find that out with GTFS

devinmatte:
I feel like it looks weird showing at 1am before 5am instead of 1am after midnight like we do for trips. But I'm assuming that's showing a 1am Friday trip and 1am Saturday trip, which is why it's there. Just seemed odd

devinmatte (approved):

Issue #906, opened 2023-12-16
Allow CSV downloads of data


devinmatte:
Add CSV downloads to many datasets like ridership, service, prediction accuracy and speed so people can use our data in research without needing to reach out to us

PR #935, opened 2024-01-11
add csv download link to speed, prediction accuracy, service, ridership

cdwhitt:
## Motivation

Commonly used widgets need a link to download a .csv file so people can use our data in research without needing to reach out to us [#906](https://github.com/transitmatters/t-performance-dash/issues/906)

## Changes
Screenshot 2024-01-11 at 12 32 58 PM
Screenshot 2024-01-13 at 5 22 18 PM
Screenshot 2024-01-13 at 5 22 36 PM
Screenshot 2024-01-13 at 5 22 57 PM
Screenshot 2024-01-13 at 5 23 10 PM
Screenshot 2024-01-13 at 5 23 24 PM
Screenshot 2024-01-13 at 5 23 37 PM
Screenshot 2024-01-13 at 5 23 47 PM
Screenshot 2024-01-13 at 5 23 59 PM
Screenshot 2024-01-13 at 5 24 16 PM
Screenshot 2024-01-13 at 5 24 26 PM

## Testing Instructions
1. Load the app locally
2. Widgets on the Line Overview page should have download links (Speed, Service, Ridership). Arrival Predictions should have a download button as well. 
3. Test by adjusting time frame on the top right, click download button for widget, ensure data is within time range


modules/predictions/charts/PredictionsGraph.tsx
```diff
@@ -207,6 +212,17 @@ export const PredictionsGraph: React.FC<PredictionsGraphProps> = ({
           ]}
         />
       </ChartDiv>
+      <div className="flex flex-row items-end justify-end gap-4">
+        {startDate && (
+          <DownloadButton
+            data={dataWithPercentage as unknown as (DataPoint | AggregateDataPoint)[]}
```
skaplan-dev, line 218: where do we lose type information for this?

skaplan-dev (commented):
Could you add some images and more information to the PR description?

It also looks like there were some moving around of `ChartBorder`. Were we using it inconsistently in places?

common/components/buttons/DownloadButton.tsx
```diff
@@ -7,55 +7,23 @@ import type { DataPoint } from '../../types/dataPoints';
...
-  const date2 = endDate ? `-${endDate.replaceAll('-', '')}` : '';
-  const when = `${date1}${date2}`;
-
-  return `${where}_${what}_${when}.csv`;
-}
+import { filename } from '../../utils/csv';
 
 interface DownloadButtonProps {
   datasetName: string;
   data: (DataPoint | AggregateDataPoint)[];
```
idreyn, line 14: I think we can loosen this type to something like `Record<string, any>[]`.

common/utils/csv.ts
```diff
@@ -0,0 +1,35 @@
...
+const directionAbbrs = {
+  northbound: 'NB',
+  southbound: 'SB',
+  eastbound: 'EB',
+  westbound: 'WB',
+  inbound: 'IB',
+  outbound: 'OB',
+};
+
+export function filename(
```
idreyn, line 12: This is a pretty good abstraction. Few thoughts:

- It could have a more specific name, like `getCsvFilename`
- The `bothStops` argument should be optional since many callers don't need to think about stops at all
- With this many arguments I'd consider letting users pass in an object:

```ts
type GetCsvFilenameOptions = {
  datasetName: string;
  startDate: string;
  endDate: string;
  location?: Location;
  includeBothStopsForLocation?: Location;
}

const getCsvFilename = (options: GetCsvFilenameOptions) => { ... }
```

modules/predictions/utils/utils.ts
```diff
@@ -77,3 +77,14 @@ const calcValues = (predictions: TimePredictionWeek[]) => {
 export const getDetailsPredictiondWidgetValues = (datapoints: TimePredictionWeek[]) => {
   return calcValues(datapoints);
 };
+
+export const addAccuracyPercentageToData = (data: TimePredictionWeek[]) => {
```
idreyn, line 81: Maybe this belongs in the csv utils since it's most directly related to that.

modules/speed/utils/utils.ts
```diff
@@ -26,3 +26,13 @@ const calcValues = (speeds: DeliveredTripMetrics[], isOverview = false) => {
 export const getDetailsSpeedWidgetValues = (datapoints: DeliveredTripMetrics[]) => {
   return calcValues(datapoints);
 };
+
+export const addMPHToSpeedData = (data: DeliveredTripMetrics[]) => {
```
idreyn, line 30: Also possibly belongs in the csv utils

modules/ridership/RidershipGraph.tsx
```diff
@@ -198,6 +200,17 @@ export const RidershipGraph: React.FC<RidershipGraphProps> = ({
             ]}
           />
         </ChartDiv>
+        <div className="flex flex-row items-end justify-end gap-4">
+          {startDate && (
+            <DownloadButton
+              data={data as unknown as (DataPoint | AggregateDataPoint)[]}
```
idreyn, line 206: ❗ We should loosen the type of `data` so we don't need these casts.

modules/service/ServiceGraph.tsx
```diff
@@ -68,15 +75,29 @@ export const ServiceGraph: React.FC<ServiceGraphProps> = (props: ServiceGraphPro
...
+        benchmarks={benchmarks}
+        startDate={startDate}
+        endDate={endDate}
+        agg={config.agg}
+      />
+
+      <div className="flex flex-row items-end justify-end gap-4">
+        {startDate && (
+          <DownloadButton
+            data={data as unknown as (DataPoint | AggregateDataPoint)[]}
```
idreyn, line 93: ❗ Here too

modules/speed/SpeedGraph.tsx
```diff
@@ -192,6 +196,17 @@ export const SpeedGraph: React.FC<SpeedGraphProps> = ({
           ]}
         />
       </ChartDiv>
+      <div className="flex flex-row items-end justify-end gap-4">
+        {startDate && (
+          <DownloadButton
+            data={dataWithMPH as unknown as (DataPoint | AggregateDataPoint)[]}
```
idreyn, line 202: ❗ Here too

idreyn (changes requested):
Overall looks pretty good to me! My only hard request for changes is that we loosen the `data` type that `<DownloadButton>` accepts — it should work for any `Record<string, any>[]`, and if we use that instead, we won't need to coerce TypeScript with the `as unknown as X` casts (marked with ❗ in my comments)

Thanks for diving in to add this. This is an excellent first contribution!

modules/ridership/RidershipGraph.tsx
cdwhitt: Thanks for commenting on this! I'm working on changing now, but having some issues locally that I posted about in the slack channel. I made it so the `DownloadButton` can optionally accept multiple data types.

modules/service/ServiceGraph.tsx
cdwhitt: Same as above.

modules/speed/SpeedGraph.tsx
cdwhitt: Same as above.

cdwhitt:
I responded to probably half of the comments here, will finish up tomorrow (1/14). @skaplan-dev @idreyn

cdwhitt:
> Could you add some images/gifs and more information to the PR description?
> 
> It also looks like there were some moving around of `ChartBorder`. Were we using it inconsistently in places?

Also wanted to mention, this needed to be moved up a level as it caused some inconsistency when I added the DownloadButton to one of the widgets. I checked, and moving it didn't affect anything else.

common/utils/csv.ts
cdwhitt: I'm a little confused by this - what's the need for `includeBothStopsForLocation` if we already have an optional `location` param that's the same shape?

idreyn: Oh, I mistyped that. I meant `includeBothStopsForLocation: boolean`. That's just me riffing on a potentially clearer name for `bothStops`.

common/components/buttons/DownloadButton.tsx
```diff
@@ -3,59 +3,26 @@ import React from 'react';
...
-  return `${where}_${what}_${when}.csv`;
-}
+import { getCsvFilename } from '../../utils/csv';
 
 interface DownloadButtonProps {
   datasetName: string;
-  data: (DataPoint | AggregateDataPoint)[];
-  location: Location;
-  bothStops: boolean;
+  data: Record<string, any>[];
```
idreyn, line 13: fyi: `unknown` should work here instead of `any` and then the linter won't complain.

idreyn (approved):
Looking great! Please remove the remaining `as unknown as X` casts before you merge, now that the `data` prop has a more permissive type, pretty much anything we throw at it should work.

common/components/buttons/DownloadButton.tsx
devinmatte: Yeah in general we should prefer unknown as it throws more warnings whereas any basically disables checks.
Probably should add some linting for prefer unknown

PR #943, opened 2024-01-29
Shows gobble data in the dashboard

devinmatte:
## Motivation

Get the app in a state ready for gobble data

## Changes

Only shows gobble data when manually navigating to a bus url with a date past the current maximum.

Nobody can reach gobble data through the UI, but could by changing `startDate=2024-01-22` if they wanted to.

Adds a new warning banner to the bottom explaining that we parsed this data ourselves

### DIFF
--- a/server/chalicelib/s3.py
+++ b/server/chalicelib/s3.py
@@ -28,16 +28,27 @@ def upload(key, bytes, compress=True):
     s3.put_object(Bucket=BUCKET, Key=key, Body=bytes)
 
 
-def is_bus(stop_id):
+def is_bus(stop_id: str):
     return ("-0-" in stop_id) or ("-1-" in stop_id)
 
 
-def download_one_event_file(date, stop_id):
-    """As advertised: single event file from s3"""
-    year, month = date.year, date.month
+def get_live_folder(stop_id: str):
+    if is_bus(stop_id):
+        return "daily-bus-data"
+    else:
+        return "daily-data"
 
-    folder = "monthly-bus-data" if is_bus(stop_id) else "monthly-data"
-    key = f"Events/{folder}/{stop_id}/Year={year}/Month={month}/events.csv.gz"
+
+def download_one_event_file(date, stop_id: str, live=False):
+    """As advertised: single event file from s3"""
+    year, month, day = date.year, date.month, date.day
+
+    if live:
+        folder = get_live_folder(stop_id)
+        key = f"Events-live/{folder}/{stop_id}/Year={year}/Month={month}/Day={day}/events.csv.gz"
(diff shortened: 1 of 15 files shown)


mathcolo:
whoa! is this on beta?

devinmatte:
We have a version of this that includes commuter rail on beta. I can push this branch up instead if we want

mathcolo:
That would be great, would you mind?

devinmatte:
@mathcolo on beta now

server/chalicelib/s3.py
```diff
@@ -28,16 +28,27 @@ def upload(key, bytes, compress=True):
...
+def is_bus(stop_id: str):
     return ("-0-" in stop_id) or ("-1-" in stop_id)
 
 
-def download_one_event_file(date, stop_id):
+def get_live_folder(stop_id: str):
+    if is_bus(stop_id):
+        return "daily-bus-data"
+    else:
+        return "daily-data"
```
hamima-halim, line 39: the other folders in `Events-live/` are `daily-cr-data/` and `daily-rapid-data/`! `daily-data` is a subfolder of `Events/`

server/scripts/generate_line_files.py
```diff
@@ -20,52 +20,70 @@
     "CR-Needham",
     "CR-Newburyport",
     "CR-Providence",
-    "CR-Foxboro",
```
hamima-halim, line 23: why remove foxboro?

server/chalicelib/s3.py
devinmatte: I have this in the other branch https://github.com/transitmatters/t-performance-dash/pull/898
I guess I can have all three in this branch

idreyn (approved):
Lovely to see how well Gobble slots into our existing tools!

server/scripts/generate_line_files.py
```diff
@@ -74,16 +92,42 @@ def cr_stop_info_to_s3_prefix(line, direction, stop_id):
...
+            stops_formatted.append(
+                {
+                    "stop_name": stop["attributes"]["name"],
+                    "station": stop["id"],
+                    "branches": None,
+                    "order": index + 1,
+                    "stops": stop_layout[stop["id"]],
+                }
+            )
+        except KeyError:
```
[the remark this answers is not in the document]
devinmatte: So this was only occurring for Worcester, but I gave this a try for everything, and then many stops on some lines returned less possible stop ids than generating the list from s3. Meaning gobble is detecting the trains at stops that the MBTA api doesn't think they will stop at. So really what we need here is a merge imo

PR #950, opened 2024-02-10
Fixing aggregate query performance

devinmatte:
## Motivation

Things work a lot faster when we don't load the same file 30 times for a month...

## Changes

- Switches back from `date_range` to `month_range`


idreyn (approved):
nice find!

PR #951, opened 2024-02-14
Adding a function to properly handle gobble date ranges

devinmatte:
## Motivation

For data where we have monthly data, we don't need every date, but for gobble, we do. 
So we need a function that handles that for getting data from s3

## Changes

- Replaces `month_range` with `s3_date_range`
- Adds a `MAX_MONTH_DATA_DATE` which is the date we have monthly data for

## Testing Instructions

Test fetching data for many months in aggregate
Test fetching data from Jan 1 - present for buses via url params

### DIFF
--- a/server/chalicelib/s3_historical.py
+++ b/server/chalicelib/s3_historical.py
@@ -1,3 +1,4 @@
+from datetime import date
 from chalicelib import s3
 from chalicelib.constants import EVENT_ARRIVAL, EVENT_DEPARTURE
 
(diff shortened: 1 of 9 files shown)


hamima-halim (commented):
nit: can we also add return type annotations?

lg otherwise, but ill let the requested reviewers stamp (:

server/chalicelib/s3_historical.py
```diff
@@ -32,7 +32,7 @@ def unique_everseen(iterable, key=None):
                 yield element
 
 
-def dwells(stop_ids: list, sdate, edate):
+def dwells(stop_ids: list, sdate: str, edate: str):
```
idreyn, line 35: I kinda like `start_date` and `end_date` for names here generally, but that's just an opinion ¯\\\_(ツ)_/¯

idreyn (approved):
LGTM

Issue #969, opened 2024-04-06
Highlight/call out bunched trips


devinmatte:
When a bunched trip occurs, as in one way longer than normal headway followed by a way shorter than normal headway in quick succession, we should call it out in the charts

We can do this for buses and rapid transit (since this happens on the Green line fairly often as well)

Perhaps we need a way to denote (50%- below baseline) in the chart as a first step in this, and then we can create a bunching analysis

Idea brought up by: https://x.com/Barry_abides/status/1776017894897279004

PR #972, opened 2024-04-06
Highlight unusually low headways in chart

devinmatte:
## Motivation

Show headways which are 50% under the benchmark, often meaning that there's bunching

relates to #969 

## Changes

Screenshot 2024-04-06 at 5 25 30 PM

### DIFF
--- a/common/components/charts/Legend.tsx
+++ b/common/components/charts/Legend.tsx
@@ -3,7 +3,11 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Disclosure } from '@headlessui/react';
 import React from 'react';
 
-export const LegendSingleDay: React.FC = () => {
+interface LegendProps {
+  showUnderRatio?: boolean;
+}
+
+export const LegendSingleDay: React.FC<LegendProps> = ({ showUnderRatio = false }) => {
   return (
     <Disclosure>
       {({ open }) => (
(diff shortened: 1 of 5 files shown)


nathan-weinberg (commented):
I love this idea, but I almost wonder if we're sending the wrong message here - the blue coloring almost suggests this is a _good_ thing, which on the surface could appear to be the case - such a short wait for a train! - but if we're trying to highlight _bunching_ here, I feel like we should try to more clearly communicate what it is and why it's a bad thing

devinmatte:
I'm not married to blue as the color, I just didn't want to use a color we already use on the legend

idreyn:
I think we should consider repurposing yellow for bunched trips. Then we could do something like this:

image

devinmatte:
You thinking we shouldn't differentiate 25% late vs 50% late with different colors like we do today and instead just do red for all late trips?

idreyn:
Well, I think the system that we have works well, but if we're introducing a color for bunched trips, I think it should be a "warning" color like orange or yellow. My first thought was that we'd no longer be able to use it for slightly late trips, but maybe we could keep it for both?

idreyn:
> You thinking we shouldn't differentiate 25% late vs 50% late with different colors like we do today and instead just do red for all late trips?

So basically yes, that's the proposal. But I don't feel very strongly about it.

devinmatte:
Perhaps we can call out late trips (25%-) and bunched trips (50%-) with the same colors as 25%+ and 50%+?

idreyn:
yes I think that makes the most sense!

devinmatte:
Screenshot 2024-04-14 at 8 05 56 PM
How about this @idreyn?

idreyn (approved):
lgtm!

PR #973, opened 2024-04-07
Updates all npm packages Apr 2024

devinmatte:
## Motivation

Keep us up to date with modern software

## Changes

Updates all packages to the latest versions and fixes issues from changes


nathan-weinberg (commented):
Needs rebase @devinmatte but otherwise LGTM

devinmatte:
Okay fixed and pushed to beta

modules/landing/charts/OverallRidershipChartWrapper.tsx
```diff
@@ -5,7 +5,8 @@ import { OverallRidershipChart } from './OverallRidershipChart';
 
 export const OverallRidershipChartWrapper: React.FC = () => {
   const ridershipData = useRidershipDataLanding();
-  const ridershipDataReady = !ridershipData.isLoading && !ridershipData.isError;
+  const ridershipDataReady =
```
idreyn, line 8: Not a huge deal but this seems like a bit of regression in the TS definitions of `useQuery`. Presumably it used to be able to infer `data` was present because those other two keys are absent, but now it can't?

idreyn (approved):
code lgtm, thanks for doing this!

modules/landing/charts/OverallRidershipChartWrapper.tsx
devinmatte: Yeah it seems that they now include `undefined` into the definition of the resulting data object. Which was I think technically possible before but the typescript definition didn't include it. Overall I think this is an improvement although an annoying one

PR #976, opened 2024-04-08
Replace calls to performance API with LAMP data (Version 4.1)

devinmatte:
## Motivation

Replace all calls to the performance API with LAMP data, and gobble as fallback

## Changes

- Removes calls to the performance API
- Fetches LAMP csvs properly
- Uses v3 Alerts when available
- **Bumps release version to v4.1!**
- Runs `npm update` to do some package cleanup and updating

## Depends on

Do not consider merging until we've finished these issues

- https://github.com/transitmatters/mbta-performance/issues/2
- https://github.com/transitmatters/mbta-performance/issues/5
- https://github.com/transitmatters/mbta-performance/issues/4
- https://github.com/transitmatters/t-performance-dash/pull/978

## Testing Instructions

Go to beta, compare results to prod


server/app.py
```diff
@@ -47,19 +46,11 @@ def mutlidict_to_dict(mutlidict):
 def healthcheck():
     # These functions must return True or False :-)
     checks = {
-        "API Key Present": (lambda: len(secrets.MBTA_V2_API_KEY) > 0),
+        "API Key Present": (lambda: len(secrets.MBTA_V3_API_KEY) > 0),
         "S3 Headway Fetching": (
             lambda: "2020-11-07T10:33:40"
             in json.dumps(data_funcs.headways(date(year=2020, month=11, day=7), ["70061"]))
         ),
-        "Performance API Check": (
```
salil-khanna, line 55: Briefly read architecture, but based on this PR, I'm assuming that LAMP data is held in a separate repo through an s3 bucket? So the headway fetching thats done above covers the check for LAMP as well?

Also why does MbtaPerformanceAPI need a check but not MbtaAPIV3?

devinmatte: LAMP data is processed through mbta-performance and stored in s3. We read from s3 here.

We don't have any checks for the v3 api mostly because we haven't relied on the v3 api for much in the app. Now we only use it to show current accessibility alerts, so we could add a check here, just haven't

skaplan-dev (approved):
🚢

PR #988, opened 2024-06-10
Adding more bus routes!

devinmatte:
## Motivation

Adding more and more bus routes to the dashboard

## Changes

- Adds a large collection of new bus routes to the dashboard


server/bus/gen_bus_data.sh
```diff
@@ -2,7 +2,7 @@
 
 routes="$@"
 if [ -z "$routes" ]; then
```
rudiejd, line 4: i feel like we could be storing these in a csv at this point 😿

rudiejd (approved):

server/chalicelib/date_utils.py
```diff
@@ -8,7 +8,7 @@
 EASTERN_TIME = ZoneInfo("US/Eastern")
 
 # The most recent date for which we have monthly data
-MAX_MONTH_DATA_DATE = "2024-03-31"
+MAX_MONTH_DATA_DATE = "2024-04-30"
```
rudiejd, line 11: this will have to change

server/bus/gen_bus_data.sh
devinmatte: This whole process needs to be improved yeah.

I also need to move this whole process to mbta-performance repo

server/chalicelib/date_utils.py
devinmatte: Once we have monthly data up to June, we can update the PR and merge

devinmatte: Need to merge https://github.com/transitmatters/t-performance-dash/pull/992 first

nathan-weinberg (approved):

PR #998, opened 2024-06-25
deps: Use `@transitmatters/stripmap` package for line rendering

idreyn:
## Motivation

We have published our SVG line rendering code at `@transitmatters/stripmap` for easy reuse across our projects.

## Changes

Shouldn't induce any functional changes — strip maps should render exactly as before in slow zones view.

## Testing Instructions

I tested `/slowzones` and `/{line}/slowzones` and observed that they all worked. The tooltips appeared in the right place and the maps switched orientation at the expected breakpoints.


devinmatte (approved):
Amazing!

PR #575, opened 2023-05-15
Making a green line map

devinmatte:
## Motivation

If we ever want to show the map for slowzone or non-slowzone reasons, I figured we should have the green line

## Changes

Screenshot 2023-05-15 at 10 46 40 AM

PR #1017, opened 2024-11-26
feat: Add Green Line Slow Zones

rudiejd:
Picking up @devinmatte's work from https://github.com/transitmatters/t-performance-dash/pull/575/files#diff-3fa86ae9efe99c295f635c0b10722cdf83997ec7115bf45c93733fe9ec6038f1, but using the new stripmap component


rudiejd:
I will continue to toy with this a bit, but it's looking better :sweat_smile: 
image

package.json
```diff
@@ -13,8 +13,10 @@
     "@tanstack/react-query": "^5.54.1",
     "@tanstack/react-query-devtools": "^5.54.1",
     "@tippyjs/react": "^4.2.6",
-    "@transitmatters/stripmap": "^0.1.9",
+    "@transitmatters/stripmap": "^0.1.12",
     "@types/react-flatpickr": "^3.8.11",
+    "ajv": "^8.17.1",
+    "ajv-keywords": "^5.1.0",
```
devinmatte, line 19: What are these?

modules/slowzones/map/SlowZonesMap.tsx
```diff
@@ -149,7 +149,7 @@ export const SlowZonesMap: React.FC<SlowZonesMapProps> = ({
       diagram={diagram}
       strokeOptions={{ stroke: line.color }}
       direction={direction}
-      getSegments={getSegmentsForSlowZones}
+      getSegments={(isHorizontal) => []}
```
devinmatte, line 152: Unsure what the purpose of this is? Could this be why the highlights aren't showing on the map?

devinmatte:
Screenshot 2024-11-27 at 4 10 08 PM
Looks like the sidebar is still disabled

Screenshot 2024-11-27 at 4 10 42 PM

And the all slow zones page doesn't yet support the line

package.json
rudiejd: These were needed in order to get storybook to work

devinmatte:
Do we want to label all Green line tabs/pages with "Beta" for a week or two? In case there are any issues?

rudiejd:
That seems fine to me. I can add that label

rudiejd:
Added beta notice 
image

nathan-weinberg (approved):

skaplan-dev (approved):

common/components/notices/BetaSlowZoneDataNotice.tsx
```diff
@@ -13,7 +13,9 @@ export const BetaSlowZoneDataNotice: React.FC = () => {
           <div className="mt-2 text-sm text-yellow-700">
             <p>
               Due to the variable nature of service on the Green Line, we aren't able to detect slow
-              zones as easily as on the heavy rail lines
+              zones as easily as on the heavy rail lines. Additionaly, we are only monitoring slow
+              zones for stops serviced by the D line is that is the only line which is entirely
```
devinmatte, line 17: On the language here, perhaps an adjustment
```suggestion
              zones for stops serviced by the D line since that is the only line which is entirely
```

devinmatte: Also maybe we add "We're working on adding the other branches, but they will require additional logic to account for traffic patterns"

devinmatte (approved):

PR #1030, opened 2025-02-05
Fix landing chart data labels

PatrickCleary:
## Motivation

I have been noticing missing data labels on the landing pages. See below:
Screenshot 2025-02-05 at 2 03 23 PM

**Expected: **
Screenshot 2025-02-05 at 2 03 41 PM

## Changes

The issue is that the labels for the chart were naively generated under the assumption that all weeks would have data for every line (my fault, I wrote this code). So it only looks at red line data to generate the data labels. 

So it looks at the return trip_metrics json file and uses that to generate the labels. see here: https://dashboard.transitmatters.org/static/landing/trip_metrics.json

Currently, there was no data for the week 2024-11-18 in the dataset for the red line (another possible fix would be to generate empty data for weekly aggregates rather than exclude the data).

This caused the labels for the data to be misaligned and causes issues not only with the missing labels, but also all of the datasets were aligned on the dates present on the red line, and gaps weren't appearing in the proper locations.

**Upon further review I am pretty sure this issue stems from weekly red line data being missing. I see that for the green line, weeks with no data generate entries with values of `0` but not for the red line.** 

I tried to take a look at this but couldn't find my AWS credentials. Maybe another day.

Regardless, this change is probably good to have. 

## Testing Instructions

Verify that the results on each of the landing page charts aligns with the expected data from this endpoint: https://dashboard.transitmatters.org/static/landing/trip_metrics.json

You can also compare the dataset to the existing landing page charts and see there are errors. For example, there _is_ data for 11/18/2024 for the green line in the dataset, but the production site does not show it.

### DIFF
--- a/common/constants/charts.ts
+++ b/common/constants/charts.ts
@@ -31,6 +31,12 @@ export const DATA_LABELS_LANDING: LabelOptions = {
   font: {
     size: 12,
   },
-  display: (context) => context.dataIndex === context.dataset.data.length - 1,
+  display: (context) => {
+    const lastNonNullIndex = context.dataset.data.reduce(
+      (lastIndex, item, index) => (item !== null ? index : lastIndex),
+      -1
+    );
+    return context.dataIndex === lastNonNullIndex;
+  },
   formatter: (value) => `${value}%`, // Format the label content
 };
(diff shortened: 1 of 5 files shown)


devinmatte (approved):
🫡

PatrickCleary:
Added follow up issue
https://github.com/transitmatters/t-performance-dash/issues/1031

PR #1032, opened 2025-02-06
Fix Slow Zone map effective date

devinmatte:
## Motivation

We fixed the map when it was 0s of slow zones, but now it only works on day 2 of the slow zone. This fixes that

## Changes

- Checks dates and sets them to the start of day
- That way we're comparing apples to apples


idreyn (approved):
classic!

Issue #996, opened 2024-06-24
Graph bunching over time


mathcolo:
**Is your feature request related to a problem? Please describe.**
It would be nice to be able to know how bunching is changing over time. At the moment we only have a single statistic in single-day view:
<img width="252" alt="image" src="https://github.com/transitmatters/t-performance-dash/assets/193453/e9160a57-f46e-4bf1-8114-499e8afb8a1a">

**Describe the solution you'd like**
I can see a reduction in variance on the [current Route 1 multi-day trips page](https://dashboard.transitmatters.org/bus/trips/multi/?from=1-1-67&to=1-1-62&busRoute=1&startDate=2024-02-12&endDate=2025-02-12), but another chart showing the bunching metric specifically would be extra handy.

**Additional context**

PR #1039, opened 2025-02-16
Add some basic aggregate bunched stats

devinmatte:
## Motivation

Starts to work on #996 

## Changes

- Gives a number of bunched and on-time trips per day
- Shows in the stats under a headways over time chart

Screenshot 2025-02-16 at 2 56 34 PM


common/utils/widgets.ts
```diff
@@ -81,6 +93,20 @@ export const getAggDataWidgets = (aggData: AggregateDataPoint[], type: 'times' |
   ];
 };
 
+export const getAggHeadwayDataWidgets = (aggData: AggregateDataPoint[], type: 'times') => {
```
mathcolo, line 96: I’m unfamiliar with V4, what does this file do exactly?

server/chalicelib/aggregation.py
```diff
@@ -149,6 +150,21 @@ def headways_over_time(start_date: datetime.date, end_date: datetime.date, stops
     # combine summary stats
     summary_stats_final = pd.concat([summary_stats, summary_stats_peak])
 
+    grouped = df.groupby("service_date")
+    # Calculate the ratio of headway_time_sec to benchmark_headway_time_sec
+    df["benchmark_headway_time_sec"] = df["benchmark_headway_time_sec"].astype(float)
+    df["headway_ratio"] = df["headway_time_sec"] / df["benchmark_headway_time_sec"]
+
+    # Calculate the count of trips under 0.5 (bunched) per service_date
+    bunched = grouped.apply(lambda x: (x["headway_ratio"] <= 0.5).sum())
```
mathcolo, line 159: I know this definition of bunched trips isn’t new in this PR, but I want to double check that it’s correct. Are there any edge cases where headway_ratio would be <= 0.5 but the trip wouldn’t be bunched? I guess not?

Multiple variants of the 89 come to mind, but if those were scheduled close together then I suppose the headway_ratio would still be 1 if they were on time.

server/chalicelib/aggregation.py
```diff
@@ -149,6 +150,21 @@ def headways_over_time(start_date: datetime.date, end_date: datetime.date, stops
     # combine summary stats
     summary_stats_final = pd.concat([summary_stats, summary_stats_peak])
 
+    grouped = df.groupby("service_date")
+    # Calculate the ratio of headway_time_sec to benchmark_headway_time_sec
+    df["benchmark_headway_time_sec"] = df["benchmark_headway_time_sec"].astype(float)
```
mathcolo, line 155: This float cast isn’t giving us sig figs we don’t actually have, is it?

mathcolo (approved):

common/utils/widgets.ts
devinmatte: This is what generates the widgets below the graph, so the summary values
Screenshot 2025-02-20 at 8 46 33 PM

server/chalicelib/aggregation.py
devinmatte: Yeah we discussed this when initially introducing this logic of a bunched trip. It scales with scheduled headways. If buses are meant to be 5 minutes apart, then a bus arriving 2 minutes later is bunched. Even though these are totally fine to human amounts of time to wait, it is bunched, if the next bus comes at its scheduled time.

server/chalicelib/aggregation.py
devinmatte: It shouldn't no. This cast was just needed because the value was a string

PR #1047, opened 2025-03-02
feat: add trendlines

devinmatte:
## Motivation
- Trendlines baby

## Changes
Screenshot 2025-04-14 at 8 59 27 PM


devinmatte:
I really think if we add this it needs to be toggleable, even if we can support clicking the legend to turn it on/off

skaplan-dev:
> I really think if we add this it needs to be toggleable, even if we can support clicking the legend to turn it on/off



> I really think if we add this it needs to be toggleable, even if we can support clicking the legend to turn it on/off
Screenshot 2025-04-14 at 9 42 56 PM


it is! :P

devinmatte (commented):
Approved, but weirdly it's my PR

PR #1052, opened 2025-03-29
Dashboard openapi spec

devinmatte:
## Motivation

I wanted OpenAPI specs, we use this API in a lot of places, and it would be cool to have

## Changes

- Adds a new script to generate openapi specs
- Adds the tools to generate them, and some models


skaplan-dev:
it would be neet to make a simple gitbook or something that shows this. There are a ton of openapi -> md converters.

skaplan-dev (approved):

PR #1055, opened 2025-04-22
Added Mattapan slow zone support

ankoure:
## Motivation
Added support for Mattapan line slow zones per April, 3rd 2025 TM Labs meeting. 

## Changes

Added "Mattapan" to a bunch of Arrays, so that Mattapan line will have visualization for slow zone data.

### DIFF
--- a/common/constants/pages.ts
+++ b/common/constants/pages.ts
@@ -143,7 +143,7 @@ export const ALL_PAGES: PageMap = {
     key: 'slowzones',
     path: '/slowzones',
     name: 'Slow zones',
-    lines: ['line-red', 'line-blue', 'line-orange', 'line-green'],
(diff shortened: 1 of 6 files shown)


devinmatte:
Should be good to merge after https://github.com/transitmatters/mbta-slow-zone-bot/pull/55

skaplan-dev (approved):

devinmatte (changes requested):
We need the segment colors for the Mattapan line

Screenshot 2025-05-04 at 6 24 18 PM

We also need to wait for the Mattapan speed restrictions PR https://github.com/transitmatters/data-ingestion/pull/130

Screenshot 2025-05-04 at 6 25 09 PM

Looks like a few other things are failing to load, may be other checks we need to fix

devinmatte (approved):
Let's go!

Issue #1071, opened 2025-08-13
Include Vehicle label/consist on Trip tooltip


devinmatte:
<img width="1226" height="503" alt="Image" src="https://github.com/user-attachments/assets/0922b41e-1c3c-444f-ac8b-b3bc2d606730" />

Include vehicle info in tooltip when we have it, so we link problem trips to specific cars

PR #1072, opened 2025-08-14
Added vehicle_consist to backend API, and Frontend Tooltip on Travel …

ankoure:
…Times, Headways, and Dwells charts (#1071)

## Motivation

This PR tackles Issue 1071 to add the new field Vehicle_Consist to the API to pull from LAMP/Gobble. It adds the relevant vehicle numbers to the Trip chart tooltips to see if we can correlate problematic trips with specific vehicles. 

## Changes
Added vehicle_consist to the API which consists of a pipe delimited string of the vehicle numbers. For testing purposes a default value of "placeholder|value|for|now" was assigned to test the splitting of the string.
Example of the pipe string split out into human readable format:
image
Example of a vehicle number for the 1 Bus
image

## Testing Instructions

Follow the directions to set up a test environment. Then for one of the lines navigate to trips/single and examine some data points in the charts.

### DIFF
--- a/common/components/charts/SingleDayLineChart.tsx
+++ b/common/components/charts/SingleDayLineChart.tsx
@@ -190,9 +190,27 @@ export const SingleDayLineChart: React.FC<SingleDayLineProps> = ({
                     const { dataIndex } = tooltipItems[0];
                     const dataPoint = data[dataIndex];
                     if (dataPoint?.vehicle_consist) {
-                      result.push(
-                        `Vehicle Number(s): ${dataPoint.vehicle_consist.replaceAll('|', ', ')}`
-                      );
+                      const arrNums = dataPoint.vehicle_consist.split('|').map(Number);
+                      if (arrNums.length > 1) {
+                        result.push(
+                          `Vehicle Number(s): ${dataPoint.vehicle_consist.replaceAll('|', ', ')}`
+                        );
+                      } else {
+                        result.push(`Vehicle Number: ${dataPoint.vehicle_consist}`);
+                      }
+                    }
+                    if (dataPoint?.vehicle_label) {
+                      if (dataPoint?.vehicle_consist) {
+                        const arrNums = dataPoint.vehicle_consist.split('|').map(Number);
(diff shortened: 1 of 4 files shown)


server/chalicelib/s3_historical.py
```diff
@@ -47,13 +47,17 @@ def dwells(stop_ids: list, start_date: date, end_date: date):
...
+            # not every vehicle will have vehicle_consist
+            vehicle_consist = maybe_a_departure.get("vehicle_consist", "placeholder|value|for|now")
             dwells.append(
                 {
                     "route_id": maybe_a_departure["route_id"],
                     "direction": int(maybe_a_departure["direction_id"]),
                     "arr_dt": date_utils.return_formatted_date(arr_dt),
                     "dep_dt": date_utils.return_formatted_date(dep_dt),
                     "dwell_time_sec": delta.total_seconds(),
+                    "vehicle_consist": vehicle_consist,
```
devinmatte, line 60: I feel like we can also include `vehicle_label`, since that shows the head vehicle. Although `vehicle_consist` may always show the head first?

common/components/charts/SingleDayLineChart.tsx
```diff
@@ -174,11 +174,28 @@ export const SingleDayLineChart: React.FC<SingleDayLineProps> = ({
...
+                    if (departureInfo) {
+                      result.push(departureInfo);
+                    }
+
+                    // Add vehicle consist information if available
+                    const { dataIndex } = tooltipItems[0];
+                    const dataPoint = data[dataIndex];
+                    if (dataPoint?.vehicle_consist) {
+                      result.push(
+                        `Vehicle Number(s): ${dataPoint.vehicle_consist.replaceAll('|', ', ')}`
```
devinmatte, line 194: This is a nitpick: We can count how many vehicles, and pick if the `s` is needed or not

ankoure:
Removed placeholder value for the .get('vehicle_consist'). Now if the API returns null it will fall back to vehicle_label on the tooltip.

PR #1075, opened 2025-08-19
927 automatically determine historic maximum for ridership

### LINKED ISSUE #927: Automatically determine historic maximum for Ridership
Right now we manually pin a historic maximum for the ridership charts, typically this is a pre-2020 peak ridership number.

We should instead determine this number automatically based on the full dataset.

My proposal is we pick the highest weekly ridership number between `1 year ago` and `beginning of data`. That way we can still show 115% of historic maximum without forever pinning to 2019

ankoure:
## Motivation

The goal of this PR is to streamline the process of calculating baseline values for ridership.  #927 

## Changes

Adds new function to the dynamo file to query the max ridership given an input start/end date and line_id. Adds a script in the server folder to call the new function (default is 1/1/2016 -> 1 year ago).  I have included the lines as they appear in baselines.ts

## Testing Instructions

I honestly need some guidance on completing this one. Since it's pulling data from Dynamo, does Dynamo follow the same formatting as the API for line-Red, line-Fairmount, etc? If so I will need to tweak the function to convert to the correct format as needed but save the .ts format. 

Also how should line-commuter-rail or DEFAULT be calculated/handled?


devinmatte (approved):

devinmatte:
Wanna look at using this soon. I want to update all numbers up to 12/31/2023 so we can push up the minimum quality expectation

I think 3 years look back is clean

Issue #968, opened 2024-04-05
Support Silver Line


devinmatte:
We should support Silver line routes in the Bus data

SL routes are not as simple as standard bus routes, so they may need a little work to display nicely

devinmatte:
We have ridership under two keys
```
line-SLWashington
line-SLWaterfront
```

PR #1076, opened 2025-08-19
Support Silver Line (#968)

ankoure:
## Motivation

Add support for Silver Line ridership #968 

## Changes

image

image

## Testing Instructions

I need help to properly diagnose the linter errors. If I change from line-SLWashington to just SLWashington, prettier converts 'SLWashington' to SLWashington as a variable(?) in baselines.ts. Which then breaks the ridership page. So i'm not sure what steps to take for this one.


devinmatte:
Does the silver line trip data work with this key?

devinmatte:
> I don't see the Silver Line Routes in Gobble and as far as I know LAMP is only subway data. Not sure if we have a comparable bus dataset.

We do have an equivalent dataset. It's managed through the T-performance-dash repository. All the new bus routes being added will need to add their historic data. I assume SL is in there as well, but not sure what the ids will be

ankoure:
@devinmatte I updated the mappings to present the ridership data correctly. We do have data for this from Gobble, however I'm pretty sure I tried to make the manifest files by hand so the station keys do not line up with what is in S3. 

When you have time it would be great to go over how to correctly create the manifests as I've been trying to use the shell scripts for the Silver Line routes and nothing is being output. I think for these manifests we will have to add the relevant branch information. I've been using the wikipedia page as a guide for how to group the different lines: [https://en.wikipedia.org/wiki/Silver_Line_(MBTA)](https://en.wikipedia.org/wiki/Silver_Line_(MBTA))

ankoure:
Ended up breaking out SL1, SL2, SL3, SL4, SL5, and SLW to their own drop down tab. Do we have some sort of data notice component that could be used to notify users that the ridership for SL1,2,3 and W are aggregated as "Silver Line Waterfront" and SL4 and SL5 are aggregated as "Silver Line Washington"?

Example Charts below: 
image

image


How the lines appear in the drop down: 

image

ankoure:
Added a notice about the aggregated data. The wording needs some word smithing as I am not totally sure if the T is providing the numbers aggregated or if we are aggregating the numbers. 

> Silver Line ridership is aggregated based on whether the line is a part of the Silver Line Waterfront (SL1, SL2, SL3, SLW) or apart of Silver Line Washington (SL4/SL5) grouping. Ridership graphs will be the same across each sub-line of the grouping.



image

devinmatte:
Okay this is looking good! I kind of want to make the SL routes have SL colors in the UI, but, this isn't worth blocking over. We should make an issue for it after we merge this

devinmatte:
Do we only have gobble data or did we also ingest historic?

devinmatte (approved):
LGTM (after fixing linting/typing issues)

Issue #1077, opened 2025-08-19
Add support for ferries


ankoure:
**Is your feature request related to a problem? Please describe.**

I would like there to be a section for Ferry data similar to the Red, Orange, Blue, Green, Bus, and Commuter Rail sections. 

**Describe the solution you'd like**

All the standard charts like Headways, Travel Times, Ridership. But also some interesting overlays could be weather (cloudy, rainy, sunny) or temperature. 

**Additional context**

OPMI has a dataset that is updated twice a year that contains ferry daily ridership by trip, route and stop. It also includes the number of passengers boarding, on board and getting off. 

[https://gis.data.mass.gov/datasets/MassDOT::mbta-ferry-daily-ridership-by-trip-route-and-stop/explore](https://gis.data.mass.gov/datasets/MassDOT::mbta-ferry-daily-ridership-by-trip-route-and-stop/explore)

ankoure:
## [Data Dictionary](https://www.arcgis.com/sharing/rest/content/items/ae21643bbe60488db8520cc694f882aa/info/metadata/metadata.xml?format=default&output=html)

| Name | Description | Data Type | Example 
| -------- | ------- | -------- | ------- |
| service_date |	Day for which service was provided. | Date | 12/31/2019
| route_id | GTFS-compatible route on which service was provided. | String |F2H 
| route_name | Description of route id. |	String | F2H-Hingham/Hull
| trip_freq | Days of the week for which the trip is available. | String | M-F
| trip_freq_adj | Indicates whether or not the trip operates under special service. “All”; if normal service. | String | Seasonal 
| sub_route | Specific route which was taken; includes stop order/origin and destination. | String | F2H-Hingham-Hull-Boston 
| trip_id | 	Time of origin departure - route id - origin code - destination code | String | 05:40-F2H-HNG-LWN 
| stop_id |	Time of stop departure - route id - stop code - destination code | String | 06:00-F2H-HUL-LWN 
| departure_terminal | Name of the stop for which boardings and loads are reported. | String | Hull 
| scheduled_departure | Time at which the vehicle is scheduled to depart the departure terminal. | Date Time | 6:00
| actual_departure | Time at which the vehicle departed the departure terminal. | Date Time | 6:00
| pax_on | The number of passengers that boarded at the departure terminal. | Integer | 16
| pax_load | The number of passengers that are on the vehicle when it leaves the departure terminal.  | Integer | 16
[...]

PR #1078, opened 2025-08-22
1077 add support for ferries

ankoure:
## Motivation

Setting up dashboard support to display Ferry Data. This also handles issue #1077 

Of particular interest is to look at patterns in ridership during Blue Line shutdowns for the East Boston Ferry. 

## Changes

- Added Ferries as their own tab with a dropdown. 
- Ridership graph
- GTFS-RT alerts 
- BUG: TripExplorer does not render for routes with only two stops (i.e East Boston). Adding another stop, causes the page to render properly. 

image

image

image

image

image

## Testing Instructions

Checkout the branch, spin up dev environment and just mess around with the data.


devinmatte:
This is great! Because it's so large and there are some issues, could you split this PR up a bit.

Perhaps add all the required style and consts to a new PR that makes them available in the code without showing anything to the user yet to shrink it a bit

devinmatte:
The whole page going blank for 2 stop routes, is because we use `end of line -1` for the default selections. So we need to update the logic for when there's only 2-3 possible stops

devinmatte:
Page errors on "Speed" chart when there is data https://dashboard-beta.labs.transitmatters.org/ferry/trips/multi/?from=Boat-Blossom&to=Boat-Long&ferryRoute=Boat-Lynn&startDate=2024-01-01&endDate=2024-12-31

Let's remove the travel time/speed toggle for ferry for now, unless we have the route distances

ankoure:
Fixed the 500 errors by updating pandas logic to handle None values

devinmatte:
@ankoure We manage ridership in data-ingestor

ankoure:
@devinmatte I fixed most of the big issues on this one. Let me know if you think it needs anything else.

devinmatte:
Weird behavior when selecting a month on Ferry

Screenshot 2025-09-28 at 8 07 37 PM

Also we should give the same behavior to single day selections, so that we can give actually useful defaults that will produce data. We used to do this for bus, but unsure if code is still around to copy

Screenshot 2025-09-28 at 8 07 51 PM

ankoure:
image

image

image

devinmatte (approved):
🚢 🇮🇹

PR #1094, opened 2025-10-01
Migrate core Covid Recovery Dashboard UI

idreyn:
## Motivation

We want to sunset the Covid Recovery Dash and centralize its functionality in the DD. We already have a daily ingestion job that will keep this up to date (as long as we manually pull files from MassDOT).

## Changes

image

image

Non-default time range
image

## Testing Instructions

Visit `/system/ridership` from the sidebar. Observe that:
- You can select Weekday/Sat/Sun for each card
- Cards load as you scroll
- The service/ridership charts respond to the global date picker


modules/serviceAndRidership/LineCard/LineCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const tabs = (
+    <TabPicker
+      className={styles.tabs}
+      value={serviceDay}
+      items={serviceDayItems}
+      onSelectValue={(d) => setServiceDay(d as any)}
+      baseId={`line-day-selector-${id}`}
+      aria-label="Select day of service"
+    />
+  );
```
devinmatte, line 96: I think we can swap this for the same selector we have throughout the dashboard, and it would look good

idreyn: image

You're right it does look pretty good 👍

modules/serviceAndRidership/LineCard/LineCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const title = shortName || longName;
+  const { startDate, endDate } = useServiceAndRidershipContext();
+
+  const ridershipAndFrequencyLabel = ridershipHistory
+    ? 'Weekday ridership and service levels'
+    : 'Weekday service levels (ridership not available)';
+
+  const renderSectionLabel = (title: string, rightElement: React.ReactNode = null) => {
+    return (
+      <h3 className={styles.sectionLabel}>
```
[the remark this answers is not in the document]
idreyn: I want to keep a consistent header color for each card rather than having a red one, green one, blue one etc. all on the same page. On the individual line pages, the colored headers look good, but I think they look messy when mixed.

I like the icon idea but can we take that as a follow-up?

modules/serviceAndRidership/ServiceAndRidershipProvider/index.ts
devinmatte, line 1: Nitpick: We should avoid barrel files https://tkdodo.eu/blog/please-stop-using-barrel-files

package-lock.json
devinmatte, line 1: Looks like this may have been generated with an older npm version (I can fix it for you)

devinmatte (approved):
Small comments, I can address myself or we can fix later

🚢 🇮🇹

PR #1101, opened 2025-10-16
1100 add toggle for weekend data

ankoure:
## Motivation

It was requested to add a filter so that only weekday/weekend values could be compared. This change adds a filter to show only Weekday values to more accurately compare headway values. 

## Changes
image

image

image

## Testing Instructions

Spin up a dev server instance and test the functionality.

### DIFF
--- a/server/chalicelib/aggregation.py
+++ b/server/chalicelib/aggregation.py
@@ -13,10 +13,20 @@ def add_holidays(
     df: pd.DataFrame,
     holiday_col_name: str = "holiday",
     service_date_col_name: str = "service_date",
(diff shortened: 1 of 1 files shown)


server/chalicelib/aggregation.py
```diff
@@ -97,6 +97,17 @@ def calc_travel_times_by_date(df: pd.DataFrame):
     # combine summary stats
     summary_stats_final = pd.concat([summary_stats, summary_stats_peak])
 
+    cal = USFederalHolidayCalendar()
```
skaplan-dev: i didnt know this was a built in pandas thing

PR #1126, opened 2026-02-15
Added docstrings to speed

ankoure:


server/chalicelib/speed_restrictions.py
```diff
@@ -25,6 +36,15 @@ def get_boundary_date(line_id: str, first: bool):
 
 
 def query_speed_restrictions(line_id: str, on_date: str):
+    """
+
+    Args:
+      line_id: str:
```
ankoure: Needs to be filled out

PR #1128, opened 2026-02-15
Add docstrings util functions

ankoure:


server/chalicelib/speed_restrictions.py
```diff
@@ -15,6 +17,15 @@ def set_speed_restrictions_table():
 
 
 def get_boundary_date(line_id: str, first: bool):
+    """
+
+    Args:
+      line_id: str:
```
ankoure: Needs to be filled out

PR #1132, opened 2026-02-15
Added docstings to another chunk 3

ankoure:


server/chalicelib/ridership.py
```diff
@@ -14,6 +18,16 @@ def get_ridership(
     start_date: date,
     end_date: date,
 ) -> List[RidershipEntry]:
+    """
+
+    Args:
+      line_id: str:
+      start_date: date:
```
ankoure: Needs to be filled out

devinmatte (approved):

PR #1157, opened 2026-04-17
Use TransitMatters benchmark as replacement for MBTA on travel-time chart

devinmatte:
## Summary
Where we have enough historical data, the travel-time chart now uses the TransitMatters benchmark in place of the MBTA scheduled benchmark — one band, not two. Dot colors (on-time / 25%+ off / etc.) are computed against whichever benchmark applies to that point. When TM data is missing (bus, CR, a stop pair with <1 year of history), the chart transparently falls back to the MBTA value.

The legend header swaps "MBTA benchmark:" for "TransitMatters benchmark:" when the chart has any TM-derived points. Tooltip labels each point's benchmark honestly — a TM-mode chart can still show "MBTA Benchmark" on individual points that fell back. Departure-from-normal text switches from "25%+ over schedule" to "25%+ over benchmark" on TM-mode charts.

Depends on mbta-performance [#85](https://github.com/transitmatters/mbta-performance/pull/85) producing the per-line JSON files this reads.

## Changes
- **Backend** — `server/chalicelib/tm_benchmarks.py` lazily loads `Benchmarks-tm/traveltimes/{Color}.json` per line (module-level cache, refreshed when the Lambda container recycles). `s3_historical.travel_times()` attaches `tm_benchmark_travel_time_sec` to each record. Existing policy already covers the new S3 prefix.
- **Chart** — effective benchmark per point is `tm[i] ?? mbta[i]`. That value drives both the gray benchmark band and the dot-color thresholds. Dataset label swaps between "TransitMatters Benchmark" and "MBTA Benchmark" so native Chart.js elements name the right source.
- **Tooltip** — per-point label names the actual benchmark source used for that point; "over schedule" → "over benchmark" when the chart is TM-mode.
- **Legend** — "Compare to … MBTA benchmark:" → "Compare to … TransitMatters benchmark:" when TM is in use. Otherwise unchanged.

## Scope
Travel times, rapid transit only. Other single-day charts (headways, dwells) and non-rapid-transit routes keep the MBTA benchmark unchanged, including the "over schedule" tooltip wording.

## Test plan
- [x] Blocked on [#85](https://github.com/transitmatters/mbta-performance/pull/85) merging and the benchmark JSONs landing in S3.
- [x] Davis → Kendall / Alewife → Ashmont: confirm the gray band sits visibly below the prior MBTA value on lines with current speed restrictions.
- [x] Bus / CR: confirm chart looks identical to main (no regression; legend still says "MBTA benchmark").
- [x] Confirm dot colors follow the new (stricter) benchmark for TM points and the MBTA benchmark for fallback points.
- [x] Hover a TM-mode chart: tooltip should say "TransitMatters Benchmark: 9m 30s" with "25%+ over benchmark" text.
- [x] Hover a non-TM chart: unchanged wording ("MBTA Benchmark", "over schedule").


ankoure (approved):
Ran through testing plan and things are looking good! Also compared against the current prod dashboard using todays data 8/12/2026

